PR #29, opened 2011-04-21
Have fred execute in background mode (reduced I/O priority) on windows NT

CAFxX:
Have fred execute in background mode (reduced I/O priority) on windows NT6+
This patch uses JNA (3.2.7) to call the native OS libraries.
jna.jar is not included in the patch but can be fetched from http://java.net/projects/jna/downloads


Thynix:
Hah. That weekend. I expect this coming build (1471) will be updating freenet-ext with the split version in `contrib` `master`. Perhaps we can add JNA to these?

bertm:
Looks like this was closed accidentally. @Thynix can you reopen this?

Thynix:
1471 isn't going to split freenet-ext either. Hrm.

nextgens:
I've merged this in next in 4b91834c5b37b8b1c7e8930f8c6e45e9d98f5157

PR #157, opened 2013-02-27
Caching store tracker

toad:


nextgens:
bump

xor-freenet:
```
<toad_> how much testing have we had of pre-releases with the caching store tracker stuff?
<toad_> IIRC there was a risk of deadlock if there were bugs, so it needed extensive testing?
<toad_> serious, update-breaking deadlock???
<toad_> has this been tested by testers?
```

Thynix:
People have been running the prereleases, yes. For finding deadlocks review
and static analysis seem a more reliable than casual testing.

PR #200, opened 2013-10-25
[INCOMPLETE] Plugin talker sync send

xor-freenet:
This pull request adds synchronous FCP send function for plugins running in the same node.

The codepath for plugins which are NOT running in the same node is NOT implemented. Therefore this pull-request is incomplete. 
I am sending it already nevertheless because WOT master branch needs this to compile.


xor-freenet:
Yes, closing this was right, the code was completely ditched in favor of plugin-fcp-rewrite to do it the proper way & follow the suggested improvements.

Issue #210, opened 2014-03-24
Make Data Store Seek Less


waldheinz:
There was some discussion on IRC yesterday about some potential improvements for the (salted hash) data store I suggested. I'd like to have this issue here attract some discussion and track the progress (if any).

The initial idea was to reduce the number of seeks (on rotating media) needed to perform the store's duties. There are two things I brought up:

  * First, I'd like to store the headers in a single file together with the payload data. Currently, this data is kept in two [separate files][1]. This already cuts the number of seeks half for the usual "get some data from the store" case. I assume this is the usual case, because the bloom filters prevent false positives with great success.
  * Next, I'd like to replace the [double-hashing scheme][2] used to probe for candidate slots when reading/writing with linear probing. This would eliminate the need for seeks here as well. There are also concerns about the function used for the double-hashing: `h + 141 i^2 + 13 i` -- the quality of this function is unknown, but voice has been raised that `141` is not prime, but maybe there should be a prime used in this place. My maths is too weak to figure that out.

To keep things simple, above would only be used for new nodes. Existing data stores are not converted. 

[1]: https://github.com/freenet/fred-staging/blob/next/src/freenet/store/saltedhash/SaltedHashFreenetStore.java#L562
[2]: https://github.com/freenet/fred-staging/blob/next/src/freenet/store/saltedhash/SaltedHashFreenetStore.java#L2090

xor-freenet:
We do not use this bugtracker.
The proper bugtracker is at: https://bugs.freenetproject.org

Can you please migrate this issue to that one on your own? If not, feel free to tell me, then I can take care of it.

PR #211, opened 2014-03-24
Unified store

waldheinz:
Because issue #210 already proposes using a new store layout for new nodes, I'd like to go one step further with the stores. The basic idea I'm following here goes like this:

  *  we want to maximize the success rate of our store
  * because the store size is bounded, the best thing we can do is to make the key distribution in the store match the incoming request distribution

So I propose having a only a single store, which

  * when not full, simply stores every block coming across the node
  * when full, stores some block `B` with probability `P_store(B)`

So what should `P_store` look like? What I'm currently experimenting with is a normal distribution around the node's location, like this:

~~~java
final double p = Math.random() * Math.random();
final double twoDist = 2 * Location.distance(block.location(), node.location());
return twoDist < p;
~~~

This superimposes a distribution which prefers nearby keys atop the incoming request distribution. I'm not sure if this is really needed, as the routing already prefers giving us requests which are close to our location. But it might help in some degenerate situations.

An even simpler scheme, closer to the original idea would be to just store blocks with an fixed probability `P`. This would be a tuneable which determines how much the store prefers keeping old data over storing new data.

This unified store would serve the purposes of the current store and cache, because:

  * there is a good chance that it contains data close to our node (because (a) data close to our location is more likely to be handled by our node and possibly (b) preferring locations close to out location by other means, like in the code above)
  * there is a good chance that it contains data requested often (because the more often some data is requested, the more often it get's a chance to make it into the store)

I currently have two nodes running in a "shootout" for a little over a week:

  * the "classic" node has a success rate of __1,1949%__
    * 0,4978% for the CHK store
    * 0,6971% for the CHK cache
  * the "unified store" node has a success rate of __1,9228%__

That's an improvement of 60%, which seems promising.

PS: I'm fully aware that this pull request cannot be merged in it's current form, but I'd like to get some input on this, and maybe someone wants to play with it, too.

### DIFF
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -4221,10 +4221,18 @@ public class Node implements TimeSkewDetectorCallback {
 	}
         
         private boolean shouldStore(double loc) {
-            if (this.chkDatastore.keyCount() < this.chkDatastore.getMaxKeys()) return true;
-            final double WIDTH = 0.05;
+            if (this.chkDatastore.keyCount() < this.chkDatastore.getMaxKeys()){
+                System.out.println("store not full");
+                return true;
+            }
+            
             final double p = Math.random() * Math.random();
-            return (2 * Location.distance(loc, this.lm.getLocation())) < Math.random();
+            final double twoDist = 2 * Location.distance(loc, this.lm.getLocation());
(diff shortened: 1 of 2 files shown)


Thynix:
Closing until there is activity.

PR #243, opened 2014-06-25
Let it burn

TheSeeker:
Since toad gave tacit approval (or at least apathy) here are the two main patches from Frost. Apply and watch the network wither and die, or eat crow and spend months arguing about why it works.  I used the less insane seeming values from mmaalfdks for requeststarter delay reduction... if the network burns to the ground with those values, oo's original values should be far worse.


xor-freenet:
Bug filed for this in case we want to reconsider it: https://bugs.freenetproject.org/view.php?id=6998

PR #247, opened 2014-06-26
Strict fairness for load management

toad:
Each peer is allowed exactly 1/N'th of the total. This eliminates the previous grossly nonlinear nonsense where we would accept all requests up to 50% overall load.

This should have some effect on maliciously patched greedy nodes. More importantly it should improve overall performance by making the AIMDs better behaved, though it may take a little while for them to settle.

NOT TESTED, POSTED FOR REVIEW. However, if people are doing crazy things to the network to boost performance, we should get in on the act too! :)


toad:
I think this is probably a bad idea.

PR #254, opened 2014-07-05
Add bandwidth configuration tests over FCP

Thynix:
This seems like an easier way to test these things instead of twisting Java unit tests into doing it.

Does the design of the test harness make sense? Should the config file in the repo be a `config.ini_sample` instead?

Could this make its way into the ant test suite somehow?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 *~
 \#*\#
 /override.properties
+*.pyc
--- a/integration-tests/config.ini
+++ b/integration-tests/config.ini
@@ -1,0 +1,4 @@
+[config]
+fcpport=9481
+# Path to a Freenet installation. Supports shell expansions.
+path=~/Freenet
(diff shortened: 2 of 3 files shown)


xor-freenet:
@bertm @Thynix Running a full node in a unit test and connecting to it by FCP is definitely possible, and not difficult. See https://github.com/freenet/plugin-WebOfTrust/blob/master/test/plugins/WebOfTrust/AbstractFullNodeTest.java
It would have the advantage of being automatically run in the JUnit test suite which the release scripts probably already run, so I would say it should definitely be rewritten in Java. Sorry :)
I would suggest waiting until plugin-fcp-rewrite is merged, as some of the simplicity of the above AbstractFullNodeTest is only possible due to (minor) improvements in fred with regards to node startup.
I am currently working on resolving the remaining merge blockers as of the current review results, and I would expect to finish it this week.

PR #258, opened 2014-07-08
Hashing API

unixninja92:
Here is the Hashing API I have written as part of my GSoC project. Also added a  check of policy files for JCE in JCELoader.


src/freenet/crypt/Hash.java
```diff
@@ -0,0 +1,205 @@
...
+	public final byte[] genHash(byte[]... input) {
+		digest.reset();
+		addBytes(input);
+		return genHash();
+	}
+	
+	/**
+	 * Generates the HashResult of all the bytes in the buffer added with the
+	 * addBytes methods. The buffer is then cleared after the hash has been
+	 * generated.
```
unixninja92: With my experience with crypto APIs, it seems standard to have the coder keep track of internal state.

unixninja92:
IIRC all the issues @toad had with my other APIs was centered around how I was working with ByteBuffers. This API only has one method that uses ByteBuffers and it is only passing it on to a different method. 

I've squashed everything down to one commit, and am working on writing up a nice commit message right now.

unixninja92:
@Thynix Does this commit message meet the commit message standard, or should I add a newline between the hashing api and jceloader lines?

Thynix:
It's close. The standards (won't be merged until at least someone else says they're happy with them - do they seem reasonable to you?) are [here](https://github.com/freenet/fred/pull/338/files#diff-6a3371457528722a734f3c51d9238c13R45).

Looking at the standards, that message needs two changes for formatting:

* The summary needs to be imperative: "added" should be "add".
* The second line must be blank.

Other than that, this looks like it should be two commits (after all, it was trying to include two summaries) - one for adding the hashing API and another for the JCELoader changes. My initial reaction is that some parts of the commit body seem unnecessary to mention - this is intended to give relevant background information to someone reviewing the change when it is made, seeing the summary in the changelog, or looking back at it to get details on why the change was made and what it was intended to do.

Therefore I'd suggest something more like:

    Add hashing API and tests
    
    This makes it easier to use and change hashing algorithms.

and

    JceLoader: log warning on limited policy file
    
    Users with limited policy files should have an indication that
    performance is known to be degraded and why.

Thoughts? Is this a reasonable response for me to make? I'd do something like this at my job but I'm having a lot of trouble figuring out what kind of feedback to give in open-source projects.

src/freenet/crypt/Hash.java
Thynix: Okay, sounds good then.

src/freenet/crypt/Hash.java
```diff
@@ -0,0 +1,205 @@
...
+	 * @param data The data to be hashed
+	 * @return Returns true if the generated hash matches the passed in hash. 
+	 * Otherwise returns false.
+	 */
+	public final boolean verify(byte[] hash, byte[]... data){
+		return MessageDigest.isEqual(hash, genHash(data));
+	}
+	
+	/**
+	 * Checks to see if the HashResults passed in are equivalent. Does a simple byte
```
Thynix: Oh! Okay.

src/freenet/crypt/JceLoader.java
```diff
@@ -55,8 +68,20 @@ static private boolean checkUse(String prop, String def)
 		}
 		BouncyCastle = p;
 		// optional
+		if (checkUse("use.SunJCE")) {
+		    try{
+		        KeyGenerator kgen = KeyGenerator.getInstance("AES", "SunJCE");
+		        kgen.init(256);
+		    }
+		    catch(Throwable e) {
+		        final String msg = "Error with SunJCE. Unlimited policy file not installed.";
```
Thynix: It looks like this string was intended to also be printed to stdout / stderr? Otherwise I'd expect it to be within the logger call.

Longer-term I'm interested in Freenet moving to an existing logging framework and doing stdout/stderr with log level configuration instead of manually.

unixninja92:
@Thynix The policy seems reasonable to me. 

Working on splitting the commit right now.

src/freenet/crypt/JceLoader.java
unixninja92: That makes sense. I think at the time I was just copying convention used at the file, but I'll change it.

unixninja92:
Should be good now. @Thynix should I merge it with master or should I leave that to you?

Thynix:
Thanks! I'll have to make a 1468 release branch in order to merge this to next, but I plan to do so ... soon. Maybe tomorrow - my schedule is still fuzzy currently.

Thynix:
Can you remind me why this is a single Hash class that supports multiple hash types instead of an interface and multiple implementing classes? I might have asked this before but I haven't been able to find it.

Thynix:
I get failing tests:

    [junit] null
    [junit] junit.framework.AssertionFailedError
    [junit] 	at freenet.crypt.HashResult.<init>(HashResult.java:30)
    [junit] 	at freenet.crypt.HashTest.testVerifyHashResultByteArrayWrongSizeMac(HashTest.java:333)

Thynix:
I've merged this into a `HashingAPI` branch for inspection.

bertm:
That assertion is expected to fail there. Consider the constructor code, and the name of the test where it fails.

```java
public HashResult(HashType hashType, byte[] bs) {
	[...]
	assert(bs.length == type.hashLength);
}
```
We'd need to get rid of that assertion (at least for unit testing). Adding another (package-private) constructor without the assertion to accomplish this may be the easiest way out, without having to drop this assertion where it is useful.

@unixninja92 please update your configuration to enable assertions on unit testing.

unixninja92:
@Thynix It's a single hash class to make it as easy as possible to upgrade to different hash types in the future. 

I'm gonna try setting up my comp for freenet dev again and resolving the test issues. I think I did a fresh install since I last worked on this.

unixninja92:
So I've fixed the assertion issue as @bertm suggested. I can't seem to figure out how to enable assertions for unit testing, so please let me know if there are any other places this needs to be fixed. 
@Thynix I'll fix up commit messages after that has been tested.

unixninja92:
@Thynix Fixed up the commits

Thynix:
Thanks! 1471 is settling down for release, so we can look at merging this
after its release. 1471-pre3 - hopefully its last prerelease - is scheduled
for this weekend.

xor-freenet:
@unixninja92 
> I can't seem to figure out how to enable assertions for unit testing, so please let me know if there are any other places this needs to be fixed.

They need to be enabled by passing ```-ea``` / ```-enableassertions``` to the JVM, and I would consider assertions being disabled in unit tests as a bug which should please be fixed.

With the Ant builder it works like [this](https://github.com/freenet/plugin-WebOfTrust/blob/14a568350112e5795c3dd48f1aef5095ea6acc3a/build.xml#L225).

Now that we've replaced Ant with Gradle on next it would be done like that:
```Groovy
tasks.withType(Test) {
    enableAssertions = true
}
```

Notice: It may be the default to enable them in Gradle as I obtained this code snipped from a [stackexchange question ](https://stackoverflow.com/questions/26399404/how-to-disable-assert-in-gradle-test)where someone asked how to disable them.

If you want to increase the probability of this getting merged then please merge next backwards into this and apply the fix for Gradle if necessary instead of just fixing it for the Ant code you had based this upon.

PR #273, opened 2014-07-31
PluginReplySender: Add getConnectionIdentifier()

xor-freenet:
- This allow identifying to which client connection a PluginReplySender belongs.
- It is need for WOT SubscriptionManager (= event-notifications) so it can determine whether a certain client is already subscribed to certain events. Also, it will have to store the connection ID in the database so when events happen in the future it can obtain the PluginReplySender with the matching ID to send the events back to.
- For pure FCP network connections, the data needed for this was already stored: A random per-network-connection identifier at FCPConnectionHandler.connectionIdentifier. It was only a matter of exposing it through a getter.
- For intra-node connections between plugins, a new random ID is stored for each PluginTalker
- The result of getConnectionIdentifier() is not only the connection ID but also contains the client-specified "Identifier" from FCP which was already existing in the code. This is to not break clients which try to maintain multiple connections by using a different "Identifier" field.

### DIFF
--- a/src/freenet/node/fcp/FCPConnectionHandler.java
+++ b/src/freenet/node/fcp/FCPConnectionHandler.java
@@ -73,7 +73,7 @@ public class FCPConnectionHandler implements Closeable {
 	private FCPClient foreverClient;
 	final BucketFactory bf;
 	final HashMap<String, ClientRequest> requestsByIdentifier;
-	protected final String connectionIdentifier;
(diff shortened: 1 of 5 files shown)


xor-freenet:
(For completeness: The review discussion of this doesn't need to be finished as PR #319 replaces the whole API which this attempted to fix.)

PR #275, opened 2014-08-10
Deprecate PluginManager.getPluginInfo: It doesn't do what its JavaDoc says

xor-freenet:
- It will not only return plugins with a matching class name but also with a matching filename
- I should have reviewed the callers for whether they need the filename matching and removed the functionality completely if they don't. But the tight WOT release schedule forbids me from spending too much time on this. So insteadd this commit just deprecates the function.
- Add replacement functions for querying by class name and by filename. Do not add one for querying by both - I doubt that this is needed. If someone needs it, it will be easy to add.

### DIFF
--- a/src/freenet/pluginmanager/PluginManager.java
+++ b/src/freenet/pluginmanager/PluginManager.java
@@ -863,13 +863,44 @@ public class PluginManager {
 	 * look for PluginInfo for a Plugin with given classname
 	 * @param plugname
 	 * @return the PluginInfo or null if not found
+	 * @deprecated As opposed to its JavaDoc, this function will also return plugins with a matching filename, no only class name.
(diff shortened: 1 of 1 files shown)


src/freenet/pluginmanager/PluginManager.java
```diff
@@ -876,6 +879,34 @@ public PluginInfoWrapper getPluginInfo(String plugname) {
...
+				if(piw.getPluginClassName().equals(pluginClassName))
+					return piw;
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @param pluginFileName The filename of the JAR from which the plugin was loaded.
+	 * @return Null if no matching plugin was found.
```
xor-freenet: Fixed by https://github.com/freenet/fred/pull/339 which had to be a separate pull request since this one is merged already.
Thanks!

PR #313, opened 2014-11-21
Load management send slow down when close to limit

ArneBab:
This is a copy of #248 with the two requested fixes added on top (255 as max byte and 65535 as max word).

The commits were actually made before I was aware of the new commit message rules, so the messages are still too long. Sorry about that…


toad:
We should separate "strict fairness" from "send slow-down when close to overload". I don't think we want strict fairness unless we're sure that load management can't work without it. Oscillation around some threshold isn't necessarily a bad thing, look at TCP.

toad:
I will submit a new pull request without the 1/N stuff, and with some cleanup.

PR #317, opened 2014-11-23
Some experimental code trying to get rid of legacy

nextgens:
Don't consider running it without backups.

NOT READY FOR MERGING


Thynix:
In addition to the bug fixes, does this need anything to go mandatory before it can be deployed?

nextgens:
nope; it's all about removing what should be dead-code

Thynix:
We talked in IRC about this. To make sure I understand - is opennet/announcement broken for all nodes or only between nodes with these changes and nodes 3 years old or older?

nextgens:
I'll split this up in several safer PRs instead

PR #319, opened 2014-12-09
plugin-fcp-rewrite (with purge-db4o merged already)

xor-freenet:
(The following text contains some words with strikethrough ("~~blah~~"). This is to indicate things which have been changed already as a consequence of the review results)

### Merge situation

By "with purge-db4o merged already" I mean that I recently merged branch next into this, so you should be able to merge it without conflicts.

### Using plugin-fcp-rewrite

A good entry-point for understanding the code is to read the JavaDoc of:
- Function PluginRespirator.connectToOtherPlugin() which is the entry point which allows plugins to use plugin-fcp-rerwite.
- Interface FredPluginFCPMessageHandler, which plugins implement to handle received messages. It notably specifies the format of the messages.
- ~~Class FCPPluginClient~~ Interface FCPPluginConnection, which plugins use to send messages to each other. In opposite to the user-focused JavaDoc at the interface, the Implementation FCPPluginConnectionImpl contains a developer/maintainer-focused overview of internal code paths which messages flow through, suitable for people who want to dig into the plugin-fcp-rewrite code.

### Changelog of what plugin-fcp-rewrite improves

- As opposed to the old PluginTalker and PluginReplySender interfaces, this new API is as symmetric as possible:
  - With the old interface, the server could only <i>reply</i> to messages of the client. It could not keep a reference to the client to send messages to it in the future to deploy events, which is absolutely critical for efficient implementation of things such as WOT.
    With this implementation, server and client are free to send messages to each others whenever they like to.
    The only restriction upon this is that the opening and closing of connections is dictated by the client. The server cannot connect to a client on its own.
  - With the new interface, it is now possible for client and server to send messages to *themselves*, i.e. server-to-server and vice versa, while pretending that they came from the remote side. While I must admit that I currently cannot think of a specific use case, it seems probable that one will arise some day. A theoretical example might be that it can be necessary to do something in a new thread - sending a message to yourself will cause it to be processed in a fresh one.

- It is more similar to standard remote procedure call:
  - Each message dialog typically involves an initial message (= a procedure call), and receives a reply message (= return value, indication of success, error code, translated error message).
  - Previously, reply messages were optional, and their format was not standardized. Nothing did even indicate whether a message was a reply. Previously, clients were free to stick error messages anywhere in the SimpleFieldSet. Now we have a standardization of when reply messages happen, and their result format (success/failure, error code, translated error message).
  - The relationship between message and reply is now is now clearly identified by the FCP "Identifier" field. Previously, it would be difficult to impossible to clearly track reply messages because their identification depended upon whether the client properly used the already existing "Identifier" field: This would usually not be the case as the identifier's location in the class hierarchy made it look like a connection identifier, not a message identifier. So clients would use the same identifier for multiple unrelated messages, which made it impossible to determine to what the client is replying. But the documentation of FCP and common sense clearly shows that it should be a message identifier, because only that allows proper tracking of message results (and because connection identifiers should be chosen by the server, not client, to prevent collisions with other clients). plugin-fcp-rewrite makes it very clear that its a per-message identifier, and provides means to use it as such (= implicit automatic random generation per message).

- Because we *can* track reply messages now, we can have an additonal function sendSynchronous(), that is a *blocking* send function which waits for a reply from the remote side and returns it. Previously, there was only send send() which would return immediately, and the reply would arrive as a callback in another thread. My development of two WOT client applications (Freetalk, and the WOT FCPClientReferenceImplementation) and review of several other WOT client appliations has shown that a blocking send is needed because WOT clients already contain at least 4 multiple, possibly bugged wheel-re-inventions of synchronous sending (Freemail, Freetalk, Content Curation, FlogHelper - dunno whether they all sucked but at least the one I wrote for Freetalk did). So its better to do it once and the right way. The advantages are:
  - ability to wait for a reply in stuff such as a GUI, where it makes sense to wait for it: If you use the web interface of a WOT client application to create an identity, you want to immediately get feedback whether the operation succeeded, or maybe failed due to invalid characters in the nickname, etc.
  - ability to not only wait for an "ACK" but actually receive the reply message in the same thread instead of in a message handler thread
  - ability to enforce a proper order of messages. Ordering is important for some algorithms, such as event-notifications of WOT. Events are a record of the past, and thus they wouldn't make sense if their order was changed.
  - The disadvantage of this possibly causing thread congestion and bad code design shall be alleviated by:
    - enforcing the caller to specify a timeout, and having an assert(timeout is smaller than sane maximal value) .
    - designing the thread in sendSynchronous() which waits for the reply to properly handle InterruptedException so the user can interrupt() all of the threads during shutdown.
    - having huge, bold warnings in the JavaDoc which provide examples when to use and when to NOT use synchronous sends.

- Interface and class names are easier to understand:
  - FredPluginFCP (old) vs FredPluginFCPMessageHandler.ServerSideMessageHandler (new)
  - FredPluginTalker (old) vs FredPluginFCPMessageHandler.ClientSideMessageHandler (new) [For symmetry, this is actually the same interface as ServerSideMessageHandler - due to the parent interface FredPluginFCPMessageHandler. The child interface merely exists to javadoc the different rules of behavior allowed for servers / clients.]
  - PluginTalker (old) vs ~~FCPPluginClient~~ FCPPluginConnection (new)
  - PluginReplySender (old) vs ~~FCPPluginClient~~ FCPPluginConnection (new). Notice that the old version duplicates the above PluginTalker as a representation of the client. This is the aforementioned non-beautiful asymmetry of the old code: To the client application, the client was a PluginTalker, to the server application, it was a PluginReplySender. Now there is just one class - ~~FCPPluginClient~~ FCPPluginConnection - which is used by both server and client.

- The public API, that is anything which shall be used by server / client plugin implementations, has very extensive JavaDoc which specifies how it can and must be used. This will prevent misuse as it has happened with the previous API as explained above. The previous API was undocumented. Thats another reason why it was chosen to rewrite the stuff instead of extending it: Backwards compatibility would have been very difficult to do without documentation which specifies what exactly we have to be compatible to.

- Also, even the internal part of the new code is documented very extensively to ensure that it will be easy to extend. This was indicated by the fact that I had to re-read the code of the old API every few months to understand how it works even though I had worked rather often with it. It was a strange architecture. Now there is a large overview of the internal paths through which messages flow at class ~~FCPPluginClient~~ FCPPluginConnectionImpl.

- It should be mostly backwards compatible with regards to external libraries: What was changed about the on-network format of FCP is only addition of new fields. The existing fields still exist, albeit their semantics has changed a bit (which shouldn't break stuff, its mostly the "Identifier" field which has been explained above to have been wrongly used in the past anyway). Backwards compatibility should be as follows:
  - Server plugin which only implements the old API + Client which only implements the old API = Old, unmodified plugins = Should keep working.
  - Server plugin which implements both the old and the new API + Client which only implements the old API = Should work. (WOT does this already.)
  - Server plugin which only implements the new API + Client which only implements the old API = Should keep working as long as the client is connected by network, not running as a plugin. "By network" = External libraries will keep working. The incompatibility between new API servers and old API clients can be worked-around by implementing *both* the old and the new API in servers. WOT does this already. Implementing both the old and new API in server plugins is less work than implementing a compatibility layer in fred, so this approach was chosen.

- It is possible to specify the thread priority of each individual message! For example this could be used to specify a higher priority for FCP functions which are typically used by the UI.

Unrelated, bonus code in the pull request - had to write this for being able to adapt WOT to plugin-fcp-rewrite nicely:
- General code quality improvements were made to ease the creation of full nodes within unit tests. This was done so I can test WOT FCP within a unit test. An example of how to create a node in unit tests is at https://github.com/freenet/plugin-WebOfTrust/blob/e62fb5824381dcb27abd38a0c65888a13b73e2e3/test/plugins/WebOfTrust/AbstractFullNodeTest.java ... Some of the aforementioned changes are:
  - It's now possible to specify the random number generator of those fake unit test nodes to be a DummyRandomSource. This is a lot faster than the standard Yarrow because it does not wait for entropy, and it allows specification of a seed for determinism and being able to log the seed of unit tests so their results can be reproduced.
  - It's now easily possible to specify the directory where the node will store its files, so you can tell JUnit to delete them afterwards. JUnit 4 is needed for this AFAIK.

### Testing situation

The following aspects of plugin-fcp-rewrite have been tested *before* merging purge-db4o into it:
- Whether communication between a plugin-fcp-rewrite server plugin and a plugin-fcp-rewrite client plugin works. This has been tested very throroughly with the WOT adaption to plugin-fcp-rewrite and its reference FCP client.
- Whether communication between a plugin-fcp-rewrite server and a networked pre-plugin-fcp-rewrite client works. This has been tested briefly using jSite.
- Whether communication between a pre-plugin-fcp-rewrite server plugin and a pre-plugin-fcp-rewrite client plugin works. This has been tested briefly using old WOT and Freetalk.

Also, TheSeeker did a test run of the most recent code *including* purge-db4o, result:
> <TheSeeker> [21:29:32] xor-freenet: WoT is loaded, and Freemail FlogHelper and Sone seem to be working.

In other words: All combinations of new server / old server / new client / old client which have above been promised to be compatible have been tested.

**Notice:** plugin-fcp-rewrite was NOT tested by myself again *after* merging purge-db4o into it, this was only tested by TheSeeker: 
There are no official adaptions of the WOT client plugins to purge-db4o available yet, only the unofficial unreviewed code of TheSeeker. I cannot execute unreviewed code right now, but might deal with reviewing them soon. Nevertheless, it would be nice if you could merge the pull request already: The merging of purge-db4o into it was done carefully, post-merge compiler issues have been fixed, the code compiles fine now, and as cited TheSeeker says it *does* work with his purge-db4o-adapted plugins.


Thynix:
Please let me know when this is ready to merge.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+            + "We should have only decided so if the handler is not null.";
+        
+        // Since the message handler is determined to be local at this point, we now must check
+        // whether it is a blocking sendSynchronous() thread instead of a regular
+        // FredPluginFCPMessageHandler.
+        // sendSynchronous() does the following: It sends a message and then blocks its thread
+        // waiting for a message replying to it to arrive so it can return it to the caller.
+        // If the message we are processing here is a reply, it might be the one which a
+        // sendSynchronous() is waiting for.
+        if(maybeDispatchMessageLocallyToSendSynchronousThread(direction, message))
```
xor-freenet: Commit df8e981e8e359d2fec7ad3af0e1b18c1c4e767e8 improves the preceding comment to better explain the "when?".
It also renames the function to be more self-explanatory.

The new name is even longer though :D This is because I have two policies for function name length, depending on the type of function:
1. Functions for "daily use", i.e. which will be used frequently in new code, shall have short names. The perfect example is the send() function which we are discussing here. send()ing network messages is something which developers will do frequently in new code, and thus the name is very short.
2. One-shot functions / internal code. This is the case with the maybe...() function whose name you wanted to shorten. It only has *one* caller; it is private; and it only exists to split up send(), not to provide any API for future use. So as its unlikely that anyone will ever have to type its name again, the name being self-explanatory is more important than saving keystrokes.

I hope you can agree with the reasoning with regards to name length.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+            + "So for networked messages, we should always be sending to the client.";
+        
+        assert (clientConnection != null)
+            : "Trying to send a message over the network to the client. "
+            + "So the network connection to it should not be null.";
+        
+        if (clientConnection.isClosed())
+            throw new IOException("Connection to client closed for " + this);
+        
+        clientConnection.outputHandler.queue(new FCPPluginServerMessage(serverPluginName, message));
```
xor-freenet: Fixed by commit c57abe82dae05e8a3a131ef5c6352d004426fda7 and 8d303650b29a1670e1cea0dd03466c2856058431.
Notice that there are probably hundreds of places in other fred code where this is also done wrong, which I did not fix. I added deprecation to encourage people to fix it in the future.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+     * {@link FredPluginFCPMessageHandler}.<br><br>
+     * 
+     * This shall only be called for messages for which it was determined that the message handler
+     * is a plugin running in the local VM.<br><br>
+     * 
+     * The message will be dispatched in a separate thread so this function can return quickly.
+     */
+    private void dispatchMessageLocallyToMessageHandler(
+            final FredPluginFCPMessageHandler messageHandler, final SendDirection direction,
+            final FCPPluginMessage message) {
```
xor-freenet: >Most of this method is message handling logic. Why is this implemented here instead of making the FredPluginFCPMessageHandler responsible for this?

FredPluginFCPMessageHandler is an interface, not an abstract class, so it cannot contain any code. And I would like to keep it that way, i.e. not make it abstract and add code, to ensure that client/server authors have a simple, short thing which they can read to understand message handling. It shall only contain stuff interesting to the server/client, not internal implementation details. Arguable its member class FCPPluginMessage violates that, but it shall be moved outside before the pull request is merged I suppose. Dunno whether someone complained about that already, but I guess I shall do it nevertheless.
(Another reason for it not being an abstract class but an interface is that it is necessary due to the way plugins are loaded: Their main class shall implement all relevant FredPlugin... interfaces. Having multiple would not be possible if they were abstract, as Java does not allow multiple inheritance, only multiple interface implementations. It would also be a severe restriction upon plugin authors to force them to extend this specific class.)
ACK?

> The getPriority() implementation in the Runnable ìn this method could then also be far prettier (i.e. without instanceof magic).

Hmm. What do you suggest instead?
Also, why are you against instanceof? IMHO optionally implementing an interface is a beautiful way, if not the de-facto way, of allowing code to add  optional functions. getPriority() being optional is a good example. If you're against it for performance reasons: Google "instanceof performance" yields several stackexchange.com threads which you might check for proof whether it is actually slow enough to justify more ugly code. The class hierarchy might be cached by the JVM very nicely... Have not checked myself though.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+            //   ignores the JavaDoc or changes the constructor.
+            // - If the assert is not true, then the following put() will replace the old
+            //   SynchronousSend, so its Condition will never get signaled, and its
+            //   thread waiting in sendSynchronous() will timeout safely. It IS possible that this
+            //   thread will then get a reply which does not belong to it. But the wrong reply will
+            //   only affect the caller, the FCPPluginClient will keep working fine, especially
+            //   no threads will become stalled for ever. As the caller is at fault for the issue,
+            //   it is fine if he breaks his own stuff :) The JavaDoc also documents this.
+            
+            as
```
xor-freenet: Nice catch!
Fixed by cdce839369b4d1ede9d8430790910230780d3b7e by moving the function to the message dispatcher class.
(It is still public so it can override toString(), which it was renamed to since there was no toString() there yet.)

src/freenet/clients/fcp/FCPPluginClientTracker.java
```diff
@@ -0,0 +1,212 @@
...
+            ref = clientsByID.get(clientID);
+        } finally {
+            clientsByIDLock.readLock().unlock();
+        }
+        
+        FCPPluginClient client = ref != null ? ref.get() : null;
+        
+        if(client == null) {
+            throw new IOException("FCPPluginClient not found, maybe it has disconnected already."
+                + " ClientID: " + clientID);
```
xor-freenet: IOException is used wherever disconnection is the most likely explanation. This is done to follow the convention of Java Sockets, for example see: https://docs.oracle.com/javase/6/docs/api/java/net/Socket.html
The non-Socket replacement would be PluginNotFoundException as thats what PluginManager typically uses when trying to query a plugin, or one of its interfaces, such as notably the FCP interface which the pull request is about.

It was difficult to decide when to use PluginNotFoundException and when to use IOException.
I decided to transition to use IOException whenever disconnection CAN be the reason, and to use PluginNotFoundException when disconnection CANNOT be the reason.
Here disconnection can be the reason: The official way for a client plugin to close a FCPPluginClient connection is to discard all hard references to it. The sole purpose of this class is to detect that by monitoring the WeakReference tree. So it pretty much is about detecting disconnection, and thats why it throws IOException.
Consequently, the callers of this function are very simple, they all just throw it outward, and it makes more and more sense the higher you go in the hierarchy. In the end it will be thrown outside a send() function.

(An example of when disconnection is not relevant and thus PluginNotFoundException is thrown is PluginRespirator.connectToOtherPlugin())

ACK to keep as is?

xor-freenet: IOException is used wherever disconnection is the most likely explanation. This is done to follow the convention of Java Sockets, for example see: https://docs.oracle.com/javase/6/docs/api/java/net/Socket.html
The non-Socket replacement would be PluginNotFoundException as thats what PluginManager typically uses when trying to query a plugin, or one of its interfaces, such as notably the FCP interface which the pull request is about.

It was difficult to decide when to use PluginNotFoundException and when to use IOException.
I decided to transition to use IOException whenever disconnection CAN be the reason, and to use PluginNotFoundException when disconnection CANNOT be the reason.
Here disconnection can be the reason: The official way for a client plugin to close a FCPPluginClient connection is to discard all hard references to it. The sole purpose of this class we are discussing here at that comment is to detect that by monitoring the WeakReference tree. So it pretty much is about detecting disconnection, and thats why it throws IOException.
Consequently, the callers of this function are very simple, they all just throw it outward, and it makes more and more sense the higher you go in the hierarchy. In the end it will be thrown outside a send() function.

(An example of when disconnection is not relevant and thus PluginNotFoundException is thrown is PluginRespirator.connectToOtherPlugin())

ACK to keep as is?

src/freenet/clients/fcp/FCPPluginClientTracker.java
```diff
@@ -0,0 +1,212 @@
...
+            } catch(InterruptedException e) {
+                // We did setDaemon(true), which causes the JVM to exit even if the thread is still
+                // running: Daemon threads are force terminated during shutdown.
+                // Thus, this thread does not need an exit mechanism, it can be an infinite loop. So
+                // nothing should try to terminate it by InterruptedException. If it does happen
+                // nevertheless, we honor it by exiting the thread, because interrupt requests
+                // should never be ignored, but log it as an error.
+                Logger.error(this,
+                    "Thread interruption requested even though this is a daemon thread!", e);
+                throw new RuntimeException(e);
```
xor-freenet: I recently told you that ThreadGroup can be used to add a handler for uncaught exceptions in the case of standard Java class Thread. I would suspect that this is also possible with our NativeThread, and it is possible that something will listen.  Thus I would like to keep it that way. It is an error after all: Nobody is supposed to shutdown the garbage-collector, this will cause memory leaks. So we do what you do upon error - throw.
ACK to keep as is?

src/freenet/clients/fcp/FCPPluginServerMessage.java
```diff
@@ -0,0 +1,183 @@
...
+	private final String plugname;
+
+    /** @see FCPPluginMessage#identifier */
+	private final String identifier;
+
+    /** @see FCPPluginMessage#params */
+	private final SimpleFieldSet plugparams;
+
+    /** @see FCPPluginMessage#success */
+    private final Boolean success;
```
xor-freenet: Look more closely: It is explained at FCPPluginMessage.success, which is also linked by the JavaDoc.

src/freenet/clients/fcp/FCPServer.java
```diff
@@ -454,6 +471,81 @@ private static String l10n(String key, String pattern, String value) {
...
+     * queried from this server by ID via the frontend {@link #getPluginClient(UUID)} as long as
+     * something else keeps a strong reference to it.<br/>
+     * Once it becomes weakly reachable, it will be garbage-collected from the backend
+     * {@link FCPPluginClientTracker} and {@link #getPluginClient(UUID)} will not work anymore.<br/>
+     * In other words, you don't have to take care of registering or unregistering clients. You only
+     * have to take care of keeping a strong reference to them while they are in use.</p>
+     * 
+     * <p>You should probably instead use the frontend function:
+     * {@link PluginRespirator#connectToOtherPlugin(String,
+     * FredPluginFCPMessageHandler.ClientSideFCPMessageHandler)}.</p>
```
xor-freenet: Nice catch. Plugins should only use PluginRespirator, they shouldn't mess with the FCPServer.
Fixed by commit 8a9a952bab8b74fcd64f3aafc1362d3c01049edc

src/freenet/crypt/PersistentRandomSource.java
```diff
@@ -0,0 +1,31 @@
...
+     * Explanation of the purpose of this mechanism is at its interface
+     * {@link PersistentRandomSource}.
+     * 
+     * @param force
+     *            If false, the implementation might decide to ignore this function call if the
+     *            seed file was written to disk a short time ago already.
+     */
+    void write_seed(boolean force);
+
+}
```
xor-freenet: Github for me specifically shows an icon which says that there is NOT a trailing newline here.
Do you want me to add one? Why?

src/freenet/pluginmanager/FredPluginFCPMessageHandler.java
```diff
@@ -0,0 +1,580 @@
...
+         * 
+         * This function is typically to construct incoming messages for passing them to the message
+         * handling function
+         * {@link FredPluginFCPMessageHandler#handlePluginFCPMessage(FCPPluginClient,
+         * FCPPluginMessage)}.<br><br>
+         * 
+         * See the JavaDoc of the member variables with the same name as the parameters for an
+         * explanation of the parameters.<br>
+         */
+        public static FCPPluginMessage constructRawMessage(ClientPermissions permissions,
```
xor-freenet: TODO for myself: This cannot be package private at the moment since the code which uses it is in package freenet.clients.fcp. However, one we move this member class FCPPluginMessage to its own file in package freenet.clients.fcp, it can be. It has grown large enough for this to make sense anyway.

src/freenet/pluginmanager/FredPluginFCPMessageHandler.java
```diff
@@ -0,0 +1,580 @@
...
+     * Plugins which provide FCP services to clients must implement this interface.<br/>
+     * The purpose of this interface is to provide a message handling function for processing
+     * messages received from the clients.
+     * 
+     * @see FredPluginFCPMessageHandler
+     *          The parent interface FredPluginFCPMessageHandler provides an overview.
+     * @see ClientSideFCPMessageHandler
+     *          The opposite version of this interface for client plugins
+     */
+    public interface ServerSideFCPMessageHandler extends FredPluginFCPMessageHandler {
```
xor-freenet: ServerSideFCPMessageHandler, and its counterpart ClientSideFCPMessageHandler, are nested in this class for the primary purpose of allowing the developer to have an overview of the "flow" of messages, and the different restrictions for servers/clients, in one place.
I would like to keep it that way.
I would suggest that I alleviate the issue of the size of this file by instead moving the member class FCPPluginMessage to its own file.
ACK?

src/freenet/pluginmanager/FredPluginFCPMessageHandler.java
```diff
@@ -0,0 +1,580 @@
...
+     * Client plugins which connect to a FCP server plugin must implement this interface.<br/>
+     * The purpose of this interface is to provide a message handling function for processing
+     * messages received from the server.
+     * 
+     * @see FredPluginFCPMessageHandler
+     *          The parent interface FredPluginFCPMessageHandler provides an overview.
+     * @see ServerSideFCPMessageHandler
+     *          The opposite version of this interface for server plugins
+     */
+    public interface ClientSideFCPMessageHandler extends FredPluginFCPMessageHandler {
```
xor-freenet: ACK, see my reply there.

src/freenet/pluginmanager/PluginInfoWrapper.java
```diff
@@ -58,7 +61,9 @@ public PluginInfoWrapper(Node node, FredPlugin plug, String filename, boolean is
 		isIPDetectorPlugin = (plug instanceof FredPluginIPDetector);
 		isPortForwardPlugin = (plug instanceof FredPluginPortForward);
 		isMultiplePlugin = (plug instanceof FredPluginMultiple);
-		isFCPPlugin = (plug instanceof FredPluginFCP);
+        isFCPPlugin = (plug instanceof FredPluginFCP);
+        isFCPServerPlugin
+            = (plug instanceof FredPluginFCPMessageHandler.ServerSideFCPMessageHandler);
```
xor-freenet: I've added what you said as "TODO: Code quality: " in commit 2ebdd5610c36db91f95215aaed8771f81bab2526 (while specifying you as commit author).
I would suggest that I do keep the code which I added the way it is, so it is coherent with the existing code while the existing code is the way it is / so that I don't have to start converting the whole of it to keep things coherent.

ACK?

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -883,10 +883,13 @@ public String dumpPlugins() {
 	 * look for PluginInfo for a Plugin with given classname
 	 * @param plugname
 	 * @return the PluginInfo or null if not found
-	 * @deprecated As opposed to its JavaDoc, this function will also return plugins with a matching filename, no only class name.
-	 *             To fix this ambiguity, instead use either {@link #getPluginInfoByClassName(String)} or {@link #getPluginInfoByFileName(String)}.
+     * @deprecated As opposed to its JavaDoc, this function will also return plugins with a matching
```
xor-freenet: That would require me to review all callers for whether they are correct under the assumption of the new JavaDoc - which is what I was precisely trying to avoid by doing it this way :)
Was trying to speed up the release of WOT by avoiding this tedious job.
ACK to leaving it as is?

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -898,39 +901,46 @@ public PluginInfoWrapper getPluginInfo(String plugname) {
...
+            }
+        }
+        return null;
+    }
 
 	/**
 	 * look for a FCPPlugin with given classname
 	 * @param plugname
 	 * @return the plugin or null if not found
+     * @deprecated Use {@link #getPluginFCPServer(String)} instead.
```
xor-freenet: The commits up to commit 84aae9936ac4f1001e74d131d8b37526dbd8dbba add JavaDoc which explains this.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -942,6 +952,27 @@ public FredPluginFCP getFCPPlugin(String plugname) {
 		return null;
 	}
 
+    /**
+     * Get the {@link FredPluginFCPMessageHandler.ServerSideFCPMessageHandler} of the plugin with
+     * the given class name.
```
xor-freenet: IMHO the fact that the plugin and the message handler are the same object is irrelevant to the outside calling code: The FCPPluginClient for example only cares about the fact that the ServerSideFCPMessageHandler IS a ServerSideFCPMessageHandler and that it thus can deliver messages to it, it doesn't need any other functionality of it. Thus it doesn't need to know that it is in fact more than just that.

Not promising more than the one interface is a good idea for future code quality IMHO: It will allow us to change the architecture of the way plugins are loaded to allow the plugin to specify DIFFERENT objects to implement the dozens of interfaces which there are. This will cut down the size of their main plugin class, which is indicated as a good idea in the main class of WOT already because it is HUGE: https://github.com/freenet/plugin-WebOfTrust/blob/50de01d423694abd1e8f71fa5563b304385c6d6a/src/plugins/WebOfTrust/WebOfTrust.java
All its handlers of plugin API, such as the FCP message handlers, are already redirected by wrapper functions to other classes anyway, which could be avoided by allowing the interfaces to be implemented by something else. Cut the problem at its root.
The uglyness is best visible for example at this function:
https://github.com/freenet/plugin-WebOfTrust/blob/50de01d423694abd1e8f71fa5563b304385c6d6a/src/plugins/WebOfTrust/WebOfTrust.java#L3770
Nobody would expect a class WebOfTrust to implement a "getString()".

ACK to keep as is?

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -942,6 +952,27 @@ public FredPluginFCP getFCPPlugin(String plugname) {
...
+     */
+    public FredPluginFCPMessageHandler.ServerSideFCPMessageHandler
+            getPluginFCPServer(String pluginClassName)
+                throws PluginNotFoundException{
+        
+        PluginInfoWrapper piw = getPluginInfoByClassName(pluginClassName);
+        if(piw != null && piw.isFCPServerPlugin()) {
+            return (FredPluginFCPMessageHandler.ServerSideFCPMessageHandler)piw.plug;
+        } else {
+            throw new PluginNotFoundException(pluginClassName);
```
xor-freenet: This function is used to create a FCP connection to a plugin. You aren't supposed to try to reconnect immediately if it fails, as the user is not expected to load/unload plugins frequently. Thus, this code can be assumed to be called infrequently, and thus the more object-oriented approach was chosen. 

ACK to keep as is?

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -942,6 +952,27 @@ public FredPluginFCP getFCPPlugin(String plugname) {
...
+     * @throws PluginNotFoundException
+     *             If the specified plugin is not loaded or does not provide an FCP server.
+     */
+    public FredPluginFCPMessageHandler.ServerSideFCPMessageHandler
+            getPluginFCPServer(String pluginClassName)
+                throws PluginNotFoundException{
+        
+        PluginInfoWrapper piw = getPluginInfoByClassName(pluginClassName);
+        if(piw != null && piw.isFCPServerPlugin()) {
+            return (FredPluginFCPMessageHandler.ServerSideFCPMessageHandler)piw.plug;
```
xor-freenet: Fixed by commit 6e00f172fdebc4e11fbeab9a0f5dad01351e6c44 - which also is a good preparation of what I suggested at https://github.com/freenet/fred/pull/319/files#r22537901 (= to allow having the plugin interfaces be implemented by different classes than the plugin main class)

src/freenet/pluginmanager/PluginReplySenderFCP.java
```diff
@@ -30,8 +33,7 @@ public void send(SimpleFieldSet params, Bucket bucket) throws PluginNotFoundExce
 		// like in linux everthing is a file, in Plugintalker everything is a plugin. So it throws PluginNotFoundException
 		// instead fcp connection errors 
 		if (handler.isClosed()) throw new PluginNotFoundException("FCP connection closed");
-		FCPPluginReply reply = new FCPPluginReply(pluginname, clientSideIdentifier, params, bucket);
+        FCPPluginServerMessage reply = new FCPPluginServerMessage(pluginname, clientSideIdentifier, params, bucket);
```
xor-freenet: The whole class which the above changes is deprecated - it is the legacy plugin-fcp API which exists so old plugins keep working. Thus it uses the deprecated constructor which was part of it.
-> No change required.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+        // We first have to compute the message.permissions field ourselves - we shall ignore what
+        // caller said for security.
+        ClientPermissions currentPermissions = (direction == SendDirection.ToClient) ?
+             null // Server-to-client messages do not have permissions.
+             : computePermissions();
+
+        // We set the permissions by creating a fresh FCPPluginMessage object so the caller cannot
+        // overwrite what we compute.
+        message = FCPPluginMessage.constructRawMessage(currentPermissions, message.identifier,
+            message.params, message.data, message.success, message.errorCode, message.errorMessage);
```
xor-freenet: I would like to keep the amount of internal functions in this class as low as possible:
- It aims at being read by all plugin authors before writing a plugin, as an overview for what fields a FCP plugin network message provides
- It is already rather large.
- Using package private visibility is not an option since it will be in the freenet.pluginmanager package.

ACK to keep as is?

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1190 @@
...
+    /**
+     * @return The permission level of this client, depending on things such as its IP address.<br>
+     *         For intra-node connections, it is {@link ClientPermissions#ACCESS_DIRECT}.<br><br>
+     * 
+     *         <b>ATTENTION:</b> The return value can change at any point in time, so you should
+     *         check this before deploying each FCP message.<br>
+     *         This is because the user is free to reconfigure IP-address restrictions on the node's
+     *         web interface whenever he wants to.
+     */
+    private ClientPermissions computePermissions() {
```
xor-freenet: Good idea!
Fixed by commit 968c86f95832b8df09ab90e210bb404ecfd1b166

src/freenet/pluginmanager/FredPluginFCPMessageHandler.java
xor-freenet: Fixed by commit c5ba5350a1893a5636619a9e585403235d3127ed

src/freenet/pluginmanager/FredPluginFCPMessageHandler.java
xor-freenet: Commit 8b1e652d8451b5817ee7e1d95893e5dad3f60420 moves FCPPluginMessage to its own file, which reduces the file size as suggested by the previous comment.

Also, addendum to my original reply:
You said:
> It needlessly complicates the interface name in external code.

Java allows import of member classes, so the external code can reference it by "ServerSideFCPMessageHandler" instead of "FredPluginFCPMessageHandler.ServerSideFCPMessageHandler".
The more lengthy version is used in some places of this pull request for making the code more understandable in its specific context. Plugin authors are IMHO fine to use the short version.

src/freenet/clients/fcp/FCPPluginClient.java
```diff
@@ -0,0 +1,1189 @@
...
+     *             in your {@link FredPluginFCPMessageHandler}.<br>
+     *             If you need to know whether the send succeeded on the same thread which shall
+     *             call the send function, you can also use {@link #sendSynchronous(SendDirection,
+     *             FCPPluginMessage, long)} which will return the reply right away.</p>
+     * @see #sendSynchronous(SendDirection, FCPPluginMessage, long)
+     *          You may instead use the blocking sendSynchronous() if your thread needs to know
+     *          whether messages arrived, to ensure a certain order of arrival, or to know
+     *          the reply to a message.
+     */
+    public void send(final SendDirection direction, FCPPluginMessage message) throws IOException {
```
xor-freenet: The requested improvements have been implemented as follows:

> Both ends are called the client? That there's some confusing naming!

FCPPluginClient has been renamed to FCPPluginConnection

> Should we really specify here in which direction we are about to send? It's rather obvious that the server only sends to the plugin, and the plugin only sends to the server, in all reasonable use cases. Maybe make this some kind of instance variable?

FCPPluginConnection now contains additional send() and sendSynchronous() which do not require a SendDirection.
They are implemented by wrapping the FCPPluginConnectionImpl in a "class DefaultSendDirectionAdapter implements FCPPluginConnection" which specifies the default direction. The wrapping is done automatically by the internal code, so plugin authors won't have to deal with it: The FCPPluginConnection objects they will be given will automatically be wrapped with the correct SendDirection depending on whether the FCPPluginConnection was handed out by API for servers or for clients.

xor-freenet: Thhe requested improvements have been implemented as follows:

    Both ends are called the client? That there's some confusing naming!

FCPPluginClient has been renamed to FCPPluginConnection.
Also, it has been converted to from a class to an interface, which leaves room for changing the implementation in the future. This is also used for adding adapters as explained below....

    Should we really specify here in which direction we are about to send? It's rather obvious that the server only sends to the plugin, and the plugin only sends to the server, in all reasonable use cases. Maybe make this some kind of instance variable?

FCPPluginConnection now contains additional send() and sendSynchronous() which do not require a SendDirection.
They are implemented by wrapping the FCPPluginConnectionImpl in a "class DefaultSendDirectionAdapter implements FCPPluginConnection" which specifies the default direction. The wrapping is done automatically by the internal code, so plugin authors won't have to deal with it: The FCPPluginConnection objects they will be given will automatically be wrapped with the correct SendDirection depending on whether the FCPPluginConnection was handed out by API for servers or for clients. They won't even notice that they're using the adapter since the type which fred returns is the interface FCPPluginConnection, not the class of the adapter.

src/freenet/pluginmanager/PluginRespirator.java
```diff
@@ -94,11 +100,115 @@ public HTMLNode addFormChild(HTMLNode parentNode, String target, String name) {
...
+     *          provides an explanation of when to use this.
+     * @param clientID
+     *            The ID as obtained by {@link FCPPluginClient#getID()}
+     * @return The client if it is still connected.
+     * @throws IOException
+     *             If there has been no client with the given ID or if it has disconnected already.
+     *             <br>If this happens, you should consider the client UUID as invalid forever and
+     *             discard it.
+     */
+    public FCPPluginClient getFCPPluginClientByID(UUID clientID) throws IOException {
```
xor-freenet: Fixed in a recent commit. It is now called getPluginConnectionByID()

xor-freenet:
I have processed all current review results. If you're OK with my resolutions, feel free to merge it now. It is finished from my perspective :)

src/freenet/clients/fcp/FCPPluginClient.java
xor-freenet: Also notice that the DefaultSendDirectionAdapter contains an additional improvement which has only been demanded on IRC:
API functions had previously required server authors to not keep a strong reference to the FCPPluginConnection. This was because garbage collection is being monitored by the FCPPluginConnectionTracker to implement client disconnection by merely requiring the client to drop all strong references to the FCPPluginConnection. 

The adapter removes the requirement of servers not keeping strong references: The version of it which is handed to servers keeps a WeakReference to the real FCPPluginConnection. 
As a consequence, there are now two ways with which servers can keep track of clients:
1) By keeping a reference to the FCPPluginConnection (which actually is a DefaultSendDirectionAdapter)
2) By merely storing the UUID of the FCPPluginConnection, and using PluginRespirator.getPluginConnectionByID() whenever they need the FCPPluginConnection object. This is useful for plugins which want to serialize everything to disk and not keep anything in memory - UUIDs can be serialized.

ArneBab:
Did anyone already get to review the changes in the new version?

Thynix:
I haven't yet because I didn't realize this was ready for merge-review.

src/freenet/pluginmanager/PluginManager.java
Thynix: I'm with @bertm. Please make the Javadoc document what it actually does in addition to the deprecation notice. This is not a breaking change because it keeps the same behavior. In the release announcement we can mention the newly deprecated methods.

Thynix:
Requiring JUnit 4 is fine. It's part of the [coding standards](https://wiki.freenetproject.org/Coding_standards#Tools_and_libraries) already.

ArneBab:
xor wrote on IRC that he can implement the old API in WOT, so old WOT plugins would not get broken.

If this would make these changes not break any plugin, I’d be for merging this (if there are no huge blockers). We’ve been shipping WoT and Freetalk changes by xor mostly unreviewed for years, and I don’t think it’s worth blocking a release with the new WoT to get this perfect.

    “Perfect is the enemy of good” — Voltaire

ArneBab:
xor wrote on IRC that he can implement the old API in WOT, so old WOT plugins would not get broken.

If this would make these changes not break any plugin, I’d be for merging this (if there are no huge blockers). We’ve been shipping WoT and Freetalk changes by xor mostly unreviewed for years, and I don’t think it’s worth blocking a release with the new WoT to get this perfect.

> “Perfect is the enemy of good” — Voltaire

xor-freenet:
To help with the justified complaints that WOT was changed to require client plugins to implement the new API, i.e. to be incompatible with old client plugins, I have implemented backwards compatibility to the old API in WOT: https://github.com/freenet/plugin-WebOfTrust/commit/5f52b4a471fdff344e2ba49d21da08ab11ea95bc
(Not merged to WOT master yet since I don't have any WOT-using plugin at hand which was adapted to compile against purge-db4o already.)

This can be done in 78 lines, which is an acceptable tradeoff for implementers of the new API IMHO. I.e. I don't think we need to add this code to fred itself since it will only affect servers which implement the new API. It is also a lot easier to implement this in server plugins than in fred because server plugins *know* whether they need to use functions of the new API, and in that case can tell the client to connect using the new API.

xor-freenet:
I've updated the pull request's main message to the recent changes due to the review results, so you can use it as a message for the merge commit.
NOTICE: Please be aware that the strikethrough-text should be removed before copy-pasting it into a text-only format such as the merge commit, because text-only cannot contain strikethrough.

xor-freenet:
TheSeeker has done a test run to check whether the backwards-compatibility fixes I implemented work, and according to him they do:
> <TheSeeker> [21:29:32] xor-freenet: WoT is loaded, and Freemail FlogHelper and Sone seem to be working.

So I'd say this code is really good to go :)

Thynix:
I don't want to merge code under time pressure without review, especially if such code is an API that we will have to maintain. Has anyone other than its author reviewed this code and is willing to vouch for it? Good intentions and testing are not a substitute for another person's review. That code functions correctly is the absolute bare minimum. One would hope also for clean design, well-considered concepts, and careful foundation. We did merge `purge-db4o` without review and only with testing because of how overwhelmingly large a changeset it was, and that attempting to do so would have paralyzed the project even more than it has currently, but that is **absolutely not** something to be expected. I agree that "perfect is the enemy of good," but that applies to known-good quantities, and this changeset is an unknown.

I do have some concerns already:

1. This changeset is very difficult to review because it is several thousand lines spread over hundreds of commits. Reviewing the diff is impractical both because of its sheer size and because it mixes separate changes. Reviewing the commits is impractical because there are so many of them and because of the need to track which concerns one has about a commit are fixed later. As far as not wanting to throw away the documentation of changes in the commit messages, designs attempted and/or decided against are only worth documenting if they seem as though they would be preferable or a viable alternative to the current design but are actually not. Documenting each change made during an incremental building toward this final functionality is valuable during initial development, but when considering the changeset as a whole it is unhelpful noise. It would be excellent to see (not to mention comparatively easy to review) this changeset as a relatively small number of commits with commit messages reflecting the present design, each changing a semantically distinct subset (often) of the overall diff.
2. At the Java API level there is a distinction between the synchronous and asynchronous communication. At the network level I would prefer this distinction be maintained instead of introducing a different mode of operation for FCP. This could be a new protocol (preferably even provided by a plugin) hosted on a different port.

What I would like to do currently is release 1468 separate from continuing to review and refine this change until people are happy with it. This means 1468 will not contain this change, so 1468's version of WoT does not depend on it. I am basing this on the understanding that WoT has not made any significant performance improvements apart from that offered by its new event-based API, nor has it fixed any correctness errors.

This is always the case, but I will state it explicitly because it seems a significant enough occasion: we can still discuss this, so I might be able to be convinced to merge this unreviewed, but I have reservations about doing so. One problem is that it has demonstrated to not be possible for volunteers to keep up with reviewing commits as they are written by even a part-time paid developer.

src/freenet/pluginmanager/PluginManager.java
Thynix: To be clear it is not necessary to review callers when making this change to documentation.

Thynix:
To relay the conclusions from discussion on IRC: this is a synchronous message API, which is an improvement in functionality beyond the current async-only plugin communication. What the users of this API actually want (and use the sync messages to build) is publish/subscribe. We want to release 1468 in less than the time it would take to add this API, so this is something to extend to incrementally later. It is @xor-freenet's evaluation that it would take longer to adapt a version of WoT which does not depend on these changes for purge-db4o than it would to finish reviewing these changes. Therefore we plan to finish reviewing these patches.

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
```diff
@@ -0,0 +1,1255 @@
...
+    /**
+     * The FCP server plugin to which this connection is connected.
+     * 
+     * <p>TODO: Optimization / Memory leak fix: Monitor this with a {@link ReferenceQueue} and if it
+     * becomes nulled, remove this FCPPluginConnectionImpl from the map
+     * {@link FCPConnectionHandler#pluginConnectionsByServerName}.<br/>
+     * Currently, it seems not necessary:<br/>
+     * - It can only become null if the server plugin is unloaded / reloaded. Plugin unloading /
+     *   reloading requires user interaction or auto update and shouldn't happen frequently.<br/>
+     * - It would only leak one WeakReference per plugin per client network connection. That won't
```
Thynix: Assuming a known leak will be small is not responsible behavior.

src/freenet/pluginmanager/PluginManager.java
xor-freenet: OK, sorry. Fixed by commit 700206ddee8a142be8a36e1ecb888533ba8264c7

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
xor-freenet: Actually, nowadays I would not even call this a real leak anymore, i.e. judge this to be even less important than the comment says.
The following comment actually does list the reasons why: It can only leak one WeakReference for each network connected client to the plugin, and this would only happen if a plugin was unloaded or updated. Remember reloading plugins is not part of normal operation, it typically only happens if we update them or if the user decides to stop using one.
It would also opportunistically get freed once the connecting client tries to send another message to the plugin - and if you do connect to a plugin by FCP, you do so because you frequently send messages.
The weight of having a thread running to monitor a ReferenceQueue for each network connection is too much for this IMHO.

ACK?

xor-freenet: Actually, nowadays I would not even call this a real leak anymore, i.e. judge this to be even less important than the comment says.
(The comment in the JavaDoc also explains this:) It can only leak one WeakReference for each network connected client to the plugin, and this would only happen if a plugin was unloaded or updated. Remember reloading plugins is not part of normal operation, it typically only happens if we update them or if the user decides to stop using one.
It would also opportunistically get freed once the connecting client tries to send another message to the plugin - and if you do connect to a plugin by FCP, you do so because you frequently send messages.
The weight of having a thread running to monitor a ReferenceQueue for each network connection is too much for this IMHO.

ACK?

xor-freenet: (Also, this is not remotely exploitable since loading/unloading plugins is not something which is exposed to the network AFAIK.)

Thynix: A memory leak is a bug. Do I understand correctly that the null entry staying in the map is the leak? This uses weak references in case the server disconnects? Does that not cause an event (when detected) that cleans up?

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
```diff
@@ -0,0 +1,1254 @@
...
+     *   a FCPPluginConnectionImpl whose server WeakReference is null, it is purged from the said
+     *   table at FCPConnectionHandler. So memory will not leak as long as the clients keep trying
+     *   to send messages to the nulled server plugin - which they probably will do because they did
+     *   already in the past.<br/>
+     * NOTICE: If you do implement this, make sure to not rewrite the ReferenceQueue polling thread
+     *         but instead base it upon {@link FCPPluginConnectionTracker}. You should probably
+     *         extract a generic class WeakValueMap from that one and use to to power both the
+     *         existing class and the one which deals with this variable here.
+     * </p>
+     * @see #isServerDead() Use isServerDead() to check whether this WeakReference is nulled.
```
Thynix: Why is there a function to check whether a reference is null? What abstraction is it intended to gain? Unless I'm missing something if one uses isServerDead(), they call it first, then get the reference, then have to check if it's null anyway because it's not impossible for it to have been GC'd in the meantime.

xor-freenet: This is used by FCPConnectionHandler to detect the situation that the WeakReference to the server has been nulled before it returns the FCPPluginConnection to the user.
This allows it to try creating a fresh FCPPluginConnection in case the plugin has been reloaded.

The WeakReference to the server is never actually accessed by the FCPConnectionHandler, so the optimization you suggested is not possible.

xor-freenet: (This is related to the semi-memory-leak you commented about earlier: This function allows the FCPConnectionHandler to detect that situation and opportunistically free it. It is important that it does so because otherwise sending messages to a plugin would not be possible anymore after it has been reloaded because the FCPConnectionHandler would always hand out the old FCPPluginConnection whose WeakRefernece is null)

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
xor-freenet: The leak is that the
```java
TreeMap<String, FCPPluginConnectionImpl> pluginConnectionsByServerName
```

in FCPConnectionHandler - of which there is one for each TCP FCP connection - will contain a single FCPPluginConnectionImpl after a given plugin has been unloaded. The leak will exist until the TCP client tries to send a message to the plugin again. Then FCPConnectionHandler will remove the stale entry from the table.

I would fix this by abstracting a generic WeakReference table tracker from FCPPluginConnectionTracker (which is used for non-networked FCP connections), but have decided to not do so yet because:
- to do it right, one would implement a generic WeakReference tracker thing, i.e. a library class, so it can be used both here and as replacement for FCPPluginConnectionTracker. Doing library classes should be done careful, and thus will take some time. Also, this is a really primitive concept, and I would expect it to be solved by common Java class libraries. So I would rather wait until we add something to freenet-ext such as Apache Java commons.
- the memory weight of a whole thread which *always* runs is probably more than the weight of a FCPPluginConnectionImpl which only exists under certain rare conditions.

I would say a middle path would be to have me document this TODO as a bugtracker entry so it is more prominently documented than a in-code FIXME. OK?

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
Thynix: This was making me concerned because I thought that the intended use case was that one would call that function to check whether `server` was null, then grab the reference and use it assuming it wasn't null if it returned false. We talked about how this is only used for the opportunistic cleanup of disconnected entries.

src/freenet/clients/fcp/FCPPluginConnectionImpl.java
Thynix: Yes, that sounds acceptable.

Thynix:
Any word on the merge conflicts? Please mark this as [unstable](https://emu.freenetproject.org/pipermail/devl/2015-March/038003.html) and fix the merge conflicts and I will merge it.

PR #320, opened 2014-12-09
Fix Eclipse not finding bcprov.jar

xor-freenet:
Commit 08da52a45633704c1486a6a92d536713c80e4bce changed
the Ant builder to consume "lib/bcprov-jdk15on-151.jar"
instead of "lib/bcprov.jar". Unfortunately, it did not update the
Eclipse project configuration to reflect that change.

This commit fixes that.


nextgens:
Superseded by #384

PR #327, opened 2015-01-11
set addItem to public

saces:
at least one plugin uses it: SchwachkopfEinsteck (gitplugin)

### DIFF
--- a/src/freenet/client/async/BaseManifestPutter.java
+++ b/src/freenet/client/async/BaseManifestPutter.java
@@ -1490,7 +1490,7 @@ public abstract class BaseManifestPutter extends ManifestPutter {
 			addItem(name, me, isDefaultDoc);
 		}
 
-		private void addItem(String name, ManifestElement element, boolean isDefaultDoc) {
+		public void addItem(String name, ManifestElement element, boolean isDefaultDoc) {
 			currentDir.put(name, element);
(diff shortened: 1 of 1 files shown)


Thynix:
Maybe this is just what happens without a sane / clearly marked plugin API, but I assume it was private for a reason - do you know if there's another way to do this? It it something that should have been public all along? (Why was the plugin written to use something private? Did it used to be public?)

ArneBab:
I guess it might be because addItem allows setting the index-file (defaultDoc). I’m not sure whether that’s the best way to do it, though.

Thynix:
Any word on this?

ArneBab:
I’d say make it public. It’s also used extensively in the DefaultManifestPutter, so I don’t see it changing anytime soon, whatever the original intention of making it private.

saces:
Sorry for the delay.. so life and stuff..

From what I remember:
The site update code gets the manifestelement tree from the node (previuos edition), modifies the items in questions and feed the stuff back to the node for insert. Yes, that simple. :)

And yes, it was public in the past: https://github.com/freenet/fred/commit/ac31ec0d36c7f014aea7f0b9325ddefcf5653d11
Praise 'git blame' ;)

Thynix:
Okay. Thanks!

ArneBab:
thank you for merging!

PR #329, opened 2015-01-14
Remove redundant function

TheSeeker:


Bombe:
While the changed code is not functionally equivalent (there’s a `core.node.config.store()` in `setAdvancedMode` that’s not in `enableAdvancedMode`) I think this is okay.

Thynix:
Merged. Thanks to TheSeeker for writing it and to Bombe for reviewing.

PR #330, opened 2015-01-20
we already know it exists

huenigre:


src/freenet/clients/fcp/ClientGet.java
```diff
@@ -183,7 +183,7 @@ public ClientGet(PersistentRequestClient globalClient, FreenetURI uri, boolean d
 			        targetFile.delete();
                     Logger.error(this, "Target file already exists but is zero length, deleting...");
 			    }
-			    if(targetFile.exists())
+			    else
 			        throw new IOException("Target filename exists already: "+targetFile);
```
bertm: Your proposed change appears valid. Unfortunately this does not comply with the coding standards, in particular this lacks the necessary curly brackets.

Suggested alternative:
```java
    ...
    Logger.error(this, "Target file already exists but is zero length, deleting...");
} else {
    throw new IOException("Target filename exists already: " + targetFile);
}
...
```

bertm: I'm sorry, this is not a valid change. In the existing code, an `IOException` is thrown when the file existed, was empty, but could not be deleted. This commit changes that behaviour, leaving a (probably unwritable) existing file in place without exception.

ArneBab:
@huenigre I think a clean fix would be to change the text of the IOException to "Target filename exists already and could not be deleted: "+targetFile

That should clear up the meaning of the code.

huenigre:
ok that does clear it up. i figured that it meant that the file was undeletable after pushing the change, but was unsure how to delete it/ got busy at work

ArneBab:
could you adjust the text of the IOException and update this pull request?

Thynix:
@huenigre Ping. Do you plan to update this pull request?

Thynix:
Closed for inactivity. Please comment if you'd like it reopened.

PR #331, opened 2015-01-25
Updated contrib/ to latest commit

RdrOfTheSt0rm:


Thynix:
What's there already is what's currently in use. Do you have a reason for updating it beyond that?

RdrOfTheSt0rm:
The contrib already there doesnt build, only the newest commons-compress source is hosted on that mirror and is version 1.1. The new contrib uses the correct (archive) server and has been updated to vversion 1.4.1.
So my main reason is that it builds out of the box.
Why arnt we using the latest commit? Is it untested?

Thynix:
I think it's because we weren't aware of a reason to and changing things
takes effort. Would it make sense to start a Debian package branch that
makes these sorts of changes? I don't know what the differences are between
v29 and master, and master is not tested.

RdrOfTheSt0rm:
Your probably right about the branch i wasent just looking at this for debian but i want to be able to build everything from source which i always seem to have trouble doing..
I guess this should be looked at when purge-db40 comes out as well? to pull it from contrib too?

Whats your recomendation here? should this just be droped? or is there someone who could maybe review contrib? Would anyone want to? is it a waste of time?
ps sorry if im wasting your time, i do not mean to im just a curoius person.

Thynix:
Hm. I recall it being inconvenient but possible when I did it last. Build audibility is definitely important.  Is it really impossible to build contrib anymore?

I think for the time being I'd rather drop this and come back to it with separating freenet-ext into its components in mind. I'm not concerned so much with reviewing an existing (but untested) version as I am about testing it. Making it monolithic like it is has been awful for keeping it up to date as you've seen. Also - no worries - I consider time answering honest questions to be time well spent. This is a good topic to consider, too.

RdrOfTheSt0rm:
It is possible but does not work "out of the box" This makes it work for me: https://github.com/RdrOfTheSt0rm/contrib/tree/new_branch 
I cant get it as a pull request against v29 tag ce3b7d55c64126f8fcdc573be3fc37f70feb7ab8 
But im hoping you can work out what im trying to do
Which freenet-ext are we currently using? Im trying to checksum it against the one i just built.
Thanks

Bombe:
…and all this hassle is exactly why nobody is using ant as a build system anymore.

Thynix:
@Bombe I am definitely open to changing to a better build system, but I don't have the experience to complete such a move quickly or well myself. You're saying that ant's lack of a stable artifact repository led to this bootstrapping failure? Whatever we move to, I'd like it to be able to check artifacts against checksums. Do you have any books to recommend? [This one](http://www.sonatype.com/resources/books/maven-the-complete-reference) maybe? I don't plan to spend time myself with the build system until after 1468 is out, but after that I'm up for it.

@RdrOfTheSt0rm Oh good, so it does still build, it's just (very) inconvenient. If memory serves the currently deployed version of freenet-ext is v29. Checksumming a jar won't do what you might hope without fancy deterministic timestamp stuff we don't have at the moment, because jars include timestamps. (The mempo folks might have some suggestions here?) However, provided you're packaging exactly the same dependencies and using the same compiler you can use [verify-formatting-only](https://github.com/freenet/scripts/blob/master/verify-formatting-only) or something based on [verify-build](https://github.com/freenet/scripts/blob/master/verify-build) to check. `verify-formatting-only` has an overly-specialized name. That's what it was written for, but the core of it is more general: it verifies that the contents of `pre-merge.jar` and `post-merge.jar` are identical.

PR #333, opened 2015-02-14
freenet.support.TransferThread: Deprecate

xor-freenet:
Reasons are explained in the added JavaDoc.

(For completeness, it shall be noted that the fixed version in WOT which the JavaDoc talks about is not committed yet. However there is a bugtracker entry at WOT which documents that it needs to be done, and it is the immediate next entry on my TODO list, so IMHO we can deprecate this already to make sure that people don't run into the bugs)

### DIFF
--- a/src/freenet/support/TransferThread.java
+++ b/src/freenet/support/TransferThread.java
@@ -29,8 +29,15 @@ import freenet.support.io.TempBucketFactory;
  * When calling <code>start()</code>, the thread will iterate the first time after <code>getStartupDelay()</code> milliseconds.
  * After each iteration, it will sleep for <code>getSleepTime()</code> milliseconds.
  * 
+ * @deprecated
+ *     This class uses {@link TrivialTicker} in a bogus way.
+ *     See <a href="https://bugs.freenetproject.org/view.php?id=6423">issue 6423</a>.<br>
+ *     The Web Of Trust plugin contains a bugfixed version, please use that one instead.<br>
+ *     (The fixed version was not moved to fred because TransferThread is only used by WOT and
+ *     Freetalk, not by fred itself. Thus it should probably be moved to a common library.)
  * @author xor
  */
+@Deprecated
 public abstract class TransferThread implements PrioRunnable, ClientGetCallback, ClientPutCallback {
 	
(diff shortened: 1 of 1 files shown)


Thynix:
This is suggesting that people who need some kind of threading thing from Fred depend on WoT? That doesn't make sense to me.

xor-freenet:
The class TransferThread, which this deprecates, was not used by *any* fred code, only Web Of Trust and Freetalk used it.
I would have to move more classes from WOT to fred to fix this class instead of deprecating it.
So effectively, if i repaired this instead of deprecating it, I would be adding more classes to fred which are not used by fred.
When I discussed this on IRC previously, people became angry because they don't want me to add code to fred which fred does not use.

I don't care, I'm both OK with repairing it in fred, or with deprecating & removing it from fred as this PR suggests.
Please make up your mind on whether it is OK to use fred as a shared library for code which is needed by multiple plugins but not fred itself :)

Thynix:
The project cannot make API changes under the assumption it can modify or even knows of all the codebases that use it. 1468 is already breaking API, so we have two options:

1. Remove it.
2. Fix it.

Given that we have the opportunity the first seems preferable.

xor-freenet:
Well this pull request *is* about deprecating it so it can be removed some time, and you said you vote for removing it, so I suppose you can just merge this now? :)

Thynix:
I don't see a reason to deprecate it before removing it within the same
build.

xor-freenet:
To give people a chance to notice that they shouldn't use it anymore and migrate to something else?

Thynix:
I'm trying to avoid two things here:

1. Fred continuing to contain a library of utilities.
2. Other plugins depending on utility classes from WoT. (Why would that even be an option?)

Therefore I would like this to either be deprecated (without the suggestion of using classes from other plugins) or removed outright.

Thynix:
I am silly. I was not looking at the test site.

ArneBab:
I’m in favor of deprecating over removing: Quite a few plugins have already been adapted to the -pre3, so I think the window of opportunity for breaking API is closed.

Thynix:
Good point; I'll merge this then.

PR #334, opened 2015-02-16
Reduce startup page refresh period to 1 second

Thynix:
Node startup can be much faster than 20 seconds, so to retain
responsiveness, especially when waiting on the first run setup, a faster
response rate is preferable.


bertm:
On a non-local node this may be a bit too fast (request fulfilment may take longer than the refresh timeout), but given the absence of any interactivity on the page that is refreshed, and modern browsers not defaulting to a blank page while refreshing any more, this should not be a problem.

Looks good to me.

Thynix:
This leaves the restart page refresh time unchanged, but I don't know if that's something to address.

Thynix:
Thanks for reviewing and good to have you back, Bert!

bertm:
> This leaves the restart page refresh time unchanged, but I don't know if that's something to address.

That's way less important, people will have become familiar with Freenet by the time they find a need to restart their nodes. Out-of-box experience for first-time users is what matters here.

PR #335, opened 2015-02-19
Checkbox: select all friends

ArneBab:
This adds a feature I wanted to have for months: Just select all my friends with a single click.

A nice improvement would be to have a central place for the Javascript, but I’m not sure where to put that.

I’m working on adding the select-all logic for the queue, too. But I think it should ideally be in all tables which have check boxes on each line: click the box in the header to check or uncheck all boxes.

### DIFF
--- a/src/freenet/clients/http/BookmarkEditorToadlet.java
+++ b/src/freenet/clients/http/BookmarkEditorToadlet.java
@@ -272,8 +272,27 @@ public class BookmarkEditorToadlet extends Toadlet {
 						if(core.node.getDarknetConnections().length > 0 && ("addItem".equals(action) || "share".equals(action))) {
 							form.addChild("br");
 							form.addChild("br");
+							if (core.node.isFProxyJavascriptEnabled()) {
+								StringBuilder jsBuf = new StringBuilder();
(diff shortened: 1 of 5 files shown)


Thynix:
This looks like nice functionality! I think it's okay to not have a non-JS alternative. It still works - just less convenient.

How about a function (a place to put this doesn't occur to me yet) that writes the Javascript to check all the boxes in a given class?

ArneBab:
This (as independent feature) is ready for merging now. I’ll leave the generic function for another time (we’d first need to define a place for generic javascript and I think that’s something to discuss in devl - which shouldn’t  block this feature).

Thynix:
I remain reluctant to introduce something with four copy-pasted blocks of duplicate JS.

Thynix:
What do you think about adding a `checkall.js` to `src/freenet/clients/http/staticfiles/js`?

Thynix:
Is there a reason to conditionally add the `<script src`?

ArneBab:
I copied what the code in ConnectionsToadlet did: only adding Javascript when the node is set to support it.

Thynix:
Thanks!

PR #338, opened 2015-03-10
Add contributing guidelines

Thynix:
GitHub will display this on the pull request page. See https://github.com/blog/1184-contributing-guidelines

Merging this will allow most of the wiki [coding standards](https://wiki.freenetproject.org/Coding_standards) page to link to this document's "Standards" section instead.

Should this document cover different things? Cover anything in a different way?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,7 @@ start with:
 * The [bug tracker](https://bugs.freenetproject.org/my_view_page.php) has tasks.
   List only bugs filed against this repository by selecting "Freenet" in the
   "Project" drop-down in the upper right. Do any of these look interesting?
+* Check the [projects](https://wiki.freenetproject.org/Projects) page.
 * Ask the [development mailing list](https://emu.freenetproject.org/cgi-bin/mailman/listinfo/devl)
   or join us in [IRC](https://freenetproject.org/irc.html) - `#freenet` on
   `chat.freenode.net`.
@@ -29,15 +30,16 @@ From the git commit man page:
 > description. The text up to the first blank line in a commit message is
 > treated as the commit title...
 
-Except for proper nouns, only the first character of the first line may be
-capitalized. It can start with the class / module the change applies to,
+Except for proper nouns, only the first character of the title may be
+capitalized. It may start with the class / module the change applies to,
 followed by a colon, especially if it helps to explain the change and give it
-context. The main part of it is a short imperative description that is at most
-72 characters and does not end with a full stop. The second line must be blank
-for, among other things, git shortlog to display only the title instead of also
-including in the description. If the message contains additional prose
-description on the third and subsequent lines, it is wrapped to 72 characters.
-Long lines that do not split well, such as URLs, are an exception to this.
+context. The main part of it is a short imperative description - a command given
+to the codebase. The title is at most 72 characters and does not end with a
+period. The second line must be blank for, among other things, git shortlog to
+display only the title instead of also including the description. If the message
+contains additional prose description on the third and subsequent lines, it is
+wrapped to 72 characters. Long lines that do not split well, such as URLs, are
+an exception to this.
 
 50 characters on the first line is a soft limit. For more discussion see
 https://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches#n87
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,10 +20,12 @@ start with:
 Once you have a change ready for review, please submit it as a [pull
 request](https://help.github.com/articles/using-pull-requests/#initiating-the-pull-request)
 against the `next` branch. It's usually a good idea to work on a topic / feature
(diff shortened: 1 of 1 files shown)


Thynix:
For the future - it's helpful to comment on the diff instead of individual commits so that outdated comments are hidden.

Thynix:
Anything else?

Thynix:
Thoughts on linking [the ten commandments of egoless programming](http://blog.codinghorror.com/the-ten-commandments-of-egoless-programming/)? Would that be too preemptive? Maybe just "Leave your ego at the door." ?

Thynix:
I've added a mention about opening pull requests. Is anyone happy with this? I'm not one to take silence as agreement.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+    Additional description of what the change does, why it is a good idea,
+    and alternate solutions decided against, if any, goes here. For some
+    changes this might not be necessary.
+
+For more discussion see the [git patch submission documentation](https://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches#n87).
+
+Text editors can be configured to assist in formatting messages this way, and
+git packages sometimes ship with such configuration.
+
+## Commit separation
```
sanity: I don't agree.  Pull requests should be reviewed, it should not be necessary to review commits.

For example, if one commit undoes the effect of another, must both be reviewed?  That would be ridiculous, yet reviewing on a commit-by-commit basis would require it.  Reviewing a pull request would not.

Thynix: Agreed, it would be ridiculous, and this is what squashing commits for
review avoids. It leaves the history easier to follow as well.

sanity: No, that's what the pull request mechanism avoids.  Squashing commits for review would only seem like a good idea if you didn't understand what a pull request was for.

Bombe: …or what commits are for. Commits and pull requests are both separate units of work; they are not merely different representations of the same thing. Squashing destroys information, a pull request does not.

Thynix: I understand that they are separate units of work, and the squashing I propose does not treat them as the same thing. I am willing to go with whatever we as a community agree to do; I have clarified and summarized things [here](https://emu.freenetproject.org/pipermail/devl/2015-March/038035.html).

ArneBab: How about saying “we avoid destroying history except if that is necessary to avoid making review a bottleneck”?

Thynix: In addition to or instead of the current content?

sanity: If you are reviewing using pull requests rather than commits, which is how
the vast majority of people use github for code review, then there is NEVER
a reason to destroy history.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+
+Except for proper nouns, only the first character of the title may be
+capitalized. It may start with the class / module the change applies to
+followed by a colon if it helps to explain the change and give it context. The
+main part of it is a short imperative description - a command given to the
+codebase. The title is at most 72 characters and does not end with a period. The
+second line is blank. If the message contains additional prose description on
+the third and subsequent lines, it is wrapped to 72 characters. Long lines that
+do not split well, such as URLs or stack traces, are an exception to this.
+
```
ArneBab: maybe we can simplify this:

> The first line is the title:
> 
> - Write the title as command to the codebase
> - Optional context: separate it with a colon
> - Use at most 72 characters
> - Do not put a period at the end

> The second line is blank. If the message contains additional prose description on
> the third and subsequent lines, it is wrapped to 72 characters. Long lines that
> do not split well, such as URLs or stack traces, are an exception to this.

Thynix: Ooh, this is a nice rephrase. It'd be nice to have consistency, and the capitalization isn't as easy to automate as one would hope as it allows capitalizing proper nouns.

Thoughts on adding

* Except for proper nouns do not capitalize beyond the first character

?

ArneBab: I’d prefer to just say “Capitalize the first word and proper nouns”

For the script we could use something like this:

    echo "foobar: limit the length to 72 characters" | sed 's/\(^.*: \)\(.\)\|\(.\)/\1\U\2\U\3/'

If there is a category (separated with ": "), capitalize the word after it. If there isn’t (`or` == `\|`), capitalize the first letter.

(thanks for the solution as often to Stack Overflow: http://stackoverflow.com/a/12422397/7666 )

ArneBab: > The first line is the title:
> 
> - Write it as a command to the codebase
> - optional context: separate it with a colon
> - Limit the length to 72 characters
> - Do not put a period at the end
> - Capitalize the first word and proper nounds

> The second line is blank. If the message contains additional prose description on
> the third and subsequent lines, it is wrapped to 72 characters. Long lines that
> do not split well, such as URLs or stack traces, are an exception to this.

Thynix:
Merged in 53b0a2f.

PR #265, opened 2014-07-14
 Toadlet: Add function showAsToadlet(ToadletContext) to replace the one without a ToadletContext parameter

xor-freenet:
This is useful for plugins which need to decide which Toadlet to return using the session cookies which can only be obtained through the ToadletContext.

PR #340, opened 2015-03-15
Toadlet.showAsToadlet(): Documentation tweaks

xor-freenet:
This fixes the review results of https://github.com/freenet/fred/pull/265 which was merged already, which is why the results have to be fixed in this fresh pull request.

### DIFF
--- a/src/freenet/clients/http/Toadlet.java
+++ b/src/freenet/clients/http/Toadlet.java
@@ -69,8 +69,15 @@ public abstract class Toadlet {
 	 * this function is necessary.
 	 * 
 	 * @param context Can be used to decide the return value, for example to check session cookies using {@link SessionManager}.
-	 * @return The result of {@link #showAsToadlet()}, which is <code>this</code> by default. Override this function to return something else for invisible
-	 *         Toadlets as explained above.
+     * @return
+     *     The result of {@link #showAsToadlet()}, which is <code>this</code> by default.<br>
+     *     The result of that deprecated function is returned instead of directly returning
+     *     <code>this</code> to keep legacy Toadlet implementations working which aim to override
+     *     the showAsToadlet() mechanism but only override the old function instead of this
+     *     one here.<br><br>
+     *     
+     *     Override this function to return something else for invisible Toadlets as explained
+     *     above.
 	 */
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/Toadlet.java
```diff
@@ -69,8 +69,15 @@
...
 	 * 
 	 * @param context Can be used to decide the return value, for example to check session cookies using {@link SessionManager}.
-	 * @return The result of {@link #showAsToadlet()}, which is <code>this</code> by default. Override this function to return something else for invisible
-	 *         Toadlets as explained above.
+     * @return
+     *     The result of {@link #showAsToadlet()}, which is <code>this</code> by default.<br>
+     *     The result of that deprecated function is returned instead of directly returning
+     *     <code>this</code> to keep legacy Toadlet implementations working which aim to override
+     *     the showAsToadlet() mechanism but only override the old function instead of this
+     *     one here.<br><br>
```
Thynix: I had trouble following this. May I suggest a rerphase?

    * The result of {@link #showAsToadlet()}, which is <code>this</code> by default.<br>
    * This is for backwards compatibility with code that overrides that function.

xor-freenet: Thanks. Hmm... your suggested rephrase is not 100%-well-defined I think.
It is not clear what the "This" in "This is.." means. The meaning of "the fact that we return the value of a deprecated function instead of returning the value itself directly" should be somehow suggested.

Here's another try:
```
* The result of {@link #showAsToadlet()}. It returns <code>this</code> by default.<br>
* The result of that deprecated function is returned instead of directly returning
* <code>this</code> to keep legacy Toadlets working as they might override the old
* function.<br><br>
```

Would that one be OK to you?

Thynix: How about

    * The result of {@link #showAsToadlet()}, which is <code>this</code> by default.<br>
    * This behavior is for backwards compatibility with existing code which overrides that function.

xor-freenet: Good idea, thanks!
Committed: https://github.com/xor-freenet/fred-staging/commit/0a02eb570d5f7f4df0902d4138fb5794bea20fa1

PR #343, opened 2015-03-26
hide the no-library warning.

ArneBab:
I kept the switch code inside to allow for easily re-adding this in case library should work well again (since we want to have a working search box again).

I used spaces for indentation - should I switch to tabs for consistency or keep spaces because that’s what we want to have in future?

If we have no other pull-requests against the WelcomeToadlet, I could also reindent the whole file.

Caveat: Library does not load for me at all, so I cannot test whether the search box gets shown correctly. For that we’d need a Library plugin which works with purge-db4o.

### DIFF
--- a/src/freenet/clients/http/WelcomeToadlet.java
+++ b/src/freenet/clients/http/WelcomeToadlet.java
@@ -112,12 +112,15 @@ public class WelcomeToadlet extends Toadlet {
 
     public boolean showSearchBox()  {
         // Only show it if Library is loaded.
-        if (node.pluginManager != null &&
-            node.pluginManager.isPluginLoaded("plugins.Library.Main")) {
-            return true;
-        } else {
-            return false;
-        }
+        return (node.pluginManager != null &&
+                node.pluginManager.isPluginLoaded("plugins.Library.Main"));
+    }
+    
+    public boolean showSearchBoxLoading()  {
+        // Only show it if Library is loaded.
+        return (node.pluginManager == null ||
+                (!node.pluginManager.isPluginLoaded("plugins.Library.Main") &&
+                 node.pluginManager.isPluginLoadedOrLoadingOrWantLoad("Library")));
     }
 
     public void addSearchBox(HTMLNode contentNode)  {
@@ -128,8 +131,7 @@ public class WelcomeToadlet extends Toadlet {
         searchBox.addChild("div", "class", "infobox-header").addChild("span", "class", "search-title-label", NodeL10n.getBase().getString("WelcomeToadlet.searchBoxLabel"));
         HTMLNode searchBoxContent = searchBox.addChild("div", "class", "infobox-content");
         // Search form
-        if(node.pluginManager != null &&
-           node.pluginManager.isPluginLoaded("plugins.Library.Main")) {
+        if(showSearchBox()) {
             // FIXME: Remove this once we have a non-broken index.
             searchBoxContent.addChild("span", "class", "search-warning-text", l10n("searchBoxWarningSlow"));
             HTMLNode searchForm = container.addFormChild(searchBoxContent, "/library/", "searchform");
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -414,33 +414,37 @@ public void handleMethodGET(URI uri, HTTPRequest request, ToadletContext ctx) th
...
-			NodeL10n.getBase().addL10nSubstitution(textSpan, "WelcomeToadlet.searchPluginLoading", new String[] { "link" }, new HTMLNode[] { HTMLNode.link("/plugins/") });
-        } else {
-			// Warn that search plugin is not loaded.
-			HTMLNode textSpan = searchBoxContent.addChild("span", "class", "search-not-availible-warning");
-			NodeL10n.getBase().addL10nSubstitution(textSpan, "WelcomeToadlet.searchPluginNotLoaded", new String[] { "link" }, new HTMLNode[] { HTMLNode.link("/plugins/") });
-		}
-		
+        // Search Box: Only show it if Library is loaded.
+        if(node.pluginManager != null &&
+           node.pluginManager.isPluginLoaded("plugins.Library.Main")) {
```
Bombe: Personally, I would add a method for the condition: showSearchBoxForLibrary() or something like that.

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -414,33 +414,37 @@ public void handleMethodGET(URI uri, HTTPRequest request, ToadletContext ctx) th
...
+            // search box is BELOW bookmarks for now, until we get search fixed properly.
+            HTMLNode searchBox = contentNode.addChild("div", "class", "infobox infobox-normal");
```
Bombe: …and I would equally add a method for the actual rendering of the search box.

ArneBab: Maybe addSearchBoxForLibrary(), since it adds the box to the contentNode?

Do I understand it correctly that  you mean replacing the whole affected code by something like

    if (showSearchBox()) {
        addSearchBox();
    }

Bombe: Plus a parameter or two, yes. It doesn’t get much clearer than this.

ArneBab:
I refactored the code into functions. Ready to merge?

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -110,6 +110,45 @@ public boolean allowPOSTWithoutPassword() {
     	return true;
     }
 
+    public boolean showSearchBox()  {
+        // Only show it if Library is loaded.
+        if (node.pluginManager != null &&
```
Bombe: `if (a) then return true else return false˙? Are you kidding me? :)

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -110,6 +110,45 @@ public boolean allowPOSTWithoutPassword() {
...
+
+    public void addSearchBox(HTMLNode contentNode)  {
+        // This function still contains legacy cruft because we might
+        // need that again when Library becomes usable again.
+        HTMLNode searchBox = contentNode.addChild("div", "class", "infobox infobox-normal");
+        searchBox.addAttribute("id", "search-freenet");
+        searchBox.addChild("div", "class", "infobox-header").addChild("span", "class", "search-title-label", NodeL10n.getBase().getString("WelcomeToadlet.searchBoxLabel"));
+        HTMLNode searchBoxContent = searchBox.addChild("div", "class", "infobox-content");
+        // Search form
+        if(node.pluginManager != null &&
```
Bombe: Hey, looks like the perfect opportunity to use the `showSearchBox` function, doesn’t it? :)

src/freenet/clients/http/WelcomeToadlet.java
ArneBab: I wasn’t sure whether Java balks out on the simpler “return (foo && bar)”.

src/freenet/clients/http/WelcomeToadlet.java
ArneBab: if I go for that, I should additionally create a “showSearchBoxLoading” function.

ArneBab:
looks much cleaner - thanks!

Thynix:
These commit messages do not meet [the standards](https://github.com/freenet/fred/pull/338/files#diff-6a3371457528722a734f3c51d9238c13R47). I don't know what to do in that situation if we cannot rewrite them, which it seems like getting consensus on contributing guidelines will require. Suggestions?

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -110,6 +110,46 @@ public boolean allowPOSTWithoutPassword() {
     	return true;
     }
 
+    public boolean showSearchBox()  {
```
Bombe: Wow, also, what’s up with the additional space here before the `{`? :)

ArneBab:
@Thynix damn, yes. That’s a point I forgot in the commit message discussion: We want to use the commit messages to reduce the work to create changelogs, and for that we need clean commit messages. “Sorry Ian, you’re not in our shoes”…

What is wrong with the messages? (I tried to follow the guidelines)

Oh, dang, the period at the end…

Thynix:
Thank you for intending to keep them in mind. The shortlog is a primary
motivation here yes.

In addition to the ending period they are not capitalized consistently
(though the standard leaves it an option), and "simpler boolean functions"
and "formatting (space after if)" are not phrased as commands to the
codebase.

ArneBab:
Thank you for merging!

PR #345, opened 2015-04-13
Fix leak warnings and issues with temporary files/buckets/buffers

toad:
Fixes bogus leak warnings by ensuring that the Buffer calls the Bucket when it is closed. This exposes some wider issues with temp buckets. Also adds a few comments explaining as this can be a bit tangled.


toad:
Sorry, mixing up stuff...

PR #347, opened 2015-04-13
Fix bucket leak on non-persistent inserts.

toad:
Fixes a leak.

### DIFF
--- a/src/freenet/clients/fcp/ClientPutBase.java
+++ b/src/freenet/clients/fcp/ClientPutBase.java
@@ -176,8 +176,8 @@ public abstract class ClientPutBase extends ClientRequest implements ClientPutCa
 			if(generatedURI == null)
 				Logger.error(this, "No generated URI in onSuccess() for "+this+" from "+state);
 		}
-		// Could restart, and is on the putter, don't free data until we remove the putter
-		//freeData(container);
+		if(persistence == Persistence.CONNECTION)
+		    freeData();
 		finish();
 		trySendFinalMessage(null);
 		if(client != null)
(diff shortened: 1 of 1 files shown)


bertm:
What makes that failed or succeeded ClientPuts with "connection" persistence cannot be restarted, and their longer-lived counterparts can? I have a feeling that we cannot just free the data here.

toad:
That's a basic property of FCP. A side-effect of Persistence=connection is once we find the data we transmit it across the connection and then it's gone, we can free it. That allows a simple request/response where there is only state while it is in progress, which is what some apps want.

bertm:
Does that imply that for Persistence=[reboot,forever], there *is* some mechanism to restart after completion or failure? I can't seem to find an FCPv2 message type providing that operation.

Nor can I find any documentation that mentions any differences in side-effects between Persistence=[connection,reboot,forever], except for when their persistence is void (e.g. on disconnecting the client, on restarting the node, and never, respectively).

So why must we treat connection-persistent requests differently here? It looks as if either we *always* need to free the data here, or should free it at the end of the persistence period (e.g. when the client disconnects, for Persistence=connection).
Am I seeing problems that are just not there, or may I be missing something entirely obvious?

toad:
IIRC there is indeed a restart mechanism, at least for inserts, but that's not the point. Reporting the data to the client is deferred for a persistent request because they may not be connected. The data is deleted only when the client explicitly removes the request with RemovePersistentRequest (name may be wrong). When the client connects we send a notice saying we have the data, but we don't send the data until they ask for it with GetRequestStatus. Possibly the FCP docs need updating.

bertm:
> Reporting the data to the client is deferred for a persistent request because they may not be connected.

Thanks, that was the part of the design logic behind this that I missed.

FCP docs do need updating to state that for Persistence=connection, the `RemoveRequest` message is implicitly received immediately after `PutFailed` or `PutSuccessful`, and subsequent `RemoveRequest` messages for the same request are invalid. Is this indeed the correct interpretation?

PR #348, opened 2015-04-13
Fix GetCHKOnly

toad:
Should fix bug #6457 and a related bogus error message.

### DIFF
--- a/src/freenet/client/async/ClientPutter.java
+++ b/src/freenet/client/async/ClientPutter.java
@@ -287,7 +287,7 @@ public class ClientPutter extends BaseClientPutter implements PutCompletionCallb
 		}
 		if(super.failedBlocks > 0 || super.fatallyFailedBlocks > 0 || super.successfulBlocks < super.totalBlocks) {
 			// USK auxiliary inserts are allowed to fail.
-			if(!uri.isUSK())
+			if((!uri.isUSK()) && !ctx.getCHKOnly)
 				Logger.error(this, "Failed blocks: "+failedBlocks+", Fatally failed blocks: "+fatallyFailedBlocks+
 						", Successful blocks: "+successfulBlocks+", Total blocks: "+totalBlocks+" but success?! on "+this+" from "+state,
 						new Exception("debug"));
--- a/src/freenet/client/async/ClientPutter.java
+++ b/src/freenet/client/async/ClientPutter.java
@@ -287,7 +287,8 @@ public class ClientPutter extends BaseClientPutter implements PutCompletionCallb
 		}
 		if(super.failedBlocks > 0 || super.fatallyFailedBlocks > 0 || super.successfulBlocks < super.totalBlocks) {
 			// USK auxiliary inserts are allowed to fail.
-			if((!uri.isUSK()) && !ctx.getCHKOnly)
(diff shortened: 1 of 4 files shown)


src/freenet/client/async/ClientPutter.java
```diff
@@ -287,7 +287,7 @@ public void onSuccess(ClientPutState state, ClientContext context) {
 		}
 		if(super.failedBlocks > 0 || super.fatallyFailedBlocks > 0 || super.successfulBlocks < super.totalBlocks) {
 			// USK auxiliary inserts are allowed to fail.
-			if(!uri.isUSK())
+			if((!uri.isUSK()) && !ctx.getCHKOnly)
```
Thynix: I don't understand - why is it okay for a CHK-only insert to have failed like this?

Thynix: Why the additional parenthesis around `!uri.isUSK()`? They don't seem necessary or helpful in understanding so my inclination is to leave them out.

While this line is changing anyway we might as well fix the whitespace to meet the [coding standards](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace): this needs a single space between `if` and the following `(`.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -175,6 +175,7 @@ public void onFinishedEncode() {
 
     @Override
     public void encodingProgress() {
+        if(ctx.getCHKOnly) return;
```
Thynix: Why does making a change to `encodingProgress()` change what's done on encoding finish? That isn't `onFinishedEncode()`?

It wasn't immediately clear to me that `getCHKOnly` meant to only generate the key, which made this confusing. Makes sense now, but how about a comment? Maybe something like

    // This request is only to generate the key. Do not encode.

This line needs a space between `if` and `(` to meet the [coding standards](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace).

ArneBab:
With this change, freesitemgr works again (tested locally).

src/freenet/client/async/ClientPutter.java
toad: The auxiliary blocks are not inserted at all, we just generate the keys for the blocks and then stop, so the block counts might be inconsistent. For a normal insert we'd generate the keys *then* insert them, and report them on the actual insert. Would it be better for SplitFileInserter to lie about the counts in this case?

toad: As regards (), in general code is unclear and thus potentially buggy if understanding it relies on remembering the operator precedence.

src/freenet/client/async/SplitFileInserter.java
toad: We have already generated the blocks. start() would check that we've generated the blocks and then attempt to actually insert. The check does disk I/O, so this is unnecessary cost, and IIRC we've already closed the file at this point in this case.

toad:
Pushed clarifications and code style fixups. Please merge.

ArneBab:
thank you for merging!

PR #349, opened 2015-04-13
Resume ClientPutDir.manifestElements on startup, fixes NPE

toad:


bertm:
Why should `ClientPutDir` know about `ContainerInserter` at all? `DefaultManifestPutter` in instance variable `putter` is supposed to handle these implementation details. Why not resume the putter instead?

toad:
True, the method needs moving somewhere more appropriate.

ArneBab:
is there a solution where to put the method?

Thynix:
Any word on this?

Thynix:
Poke poke.

toad:
I don't see why we can't just merge this? It's ugly but it's not really violating any meaningful layering?

PR #350, opened 2015-04-13
Updater insert main jar less often

toad:
This may make inserts work better for the first several hours after restarting... Please review carefully as with all updater code, and it will need to be tested too.

For 1469 or later!


toad:
This should be considered for a snapshot. It makes inserts very slow when first loading up the node and probably has an effect on the network too. Of course it needs to be looked at carefully, but in the worst case UOM can work directly without inserting the jar.

Thynix:
Would a longer-term solution be integrating some kind of KeepAlive?

toad:
I agree we do need to integrate keepalive, and there's at least one bug about what we need for it to work well. But no, in this case occasionally inserting it (from a blob on disk!) after restarting is probably the right answer.

PR #351, opened 2015-04-14
Fix broken splitfile wakeup times

toad:
Fix a serious bug in the new client layer causing completed inserts (and possibly requests too) to not get removed from the internal queue structures (RGAs) and thus prevent other inserts from running. This is the minimal version of the fix. However the other one may make more sense given that the locking is pretty tricky either way. Please review with care; this one is much shorter!

It is important that this goes in for 1468, i.e. before purge-db4o is public.

### DIFF
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1642,9 +1642,8 @@ public class SplitFileFetcherStorage {
 
     /** Returns -1 if the request is finished, otherwise the wakeup time. */
     public long getCooldownWakeupTime(long now) {
-        // It is OK to use the main lock here.
-        // Cooldown is a separate lock because updating it is tricky.
-        // FIXME maybe set cooldown when finish and avoid this.
+        // LOCKING: hasFinished() uses (this), separate from cooldownLock.
+        // It is safe to use both here (on the request selection thread), one after the other.
         if(hasFinished()) return -1;
         synchronized(cooldownLock) {
             if(overallCooldownWakeupTime < now) overallCooldownWakeupTime = 0;
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1474,6 +1474,9 @@ public class SplitFileFetcherStorage {
     /** Choose a random key which can be fetched at the moment. Must not update any persistent data;
      * it's okay to update caches and other stuff that isn't stored to disk. If we fail etc we 
      * should do it off-thread.
+     * 
+     * FIXME make SplitFileFetcherGet per-segment, eliminate all this unnecessary complexity!
+     * 
      * @return The block number to be fetched, as an integer.
      */
     public MyKey chooseRandomKey() {
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1647,9 +1647,9 @@ public class SplitFileFetcherStorage {
     public long getCooldownWakeupTime(long now) {
         // LOCKING: hasFinished() uses (this), separate from cooldownLock.
         // It is safe to use both here (on the request selection thread), one after the other.
-        if(hasFinished()) return -1;
+        if (hasFinished()) return -1;
         synchronized(cooldownLock) {
-            if(overallCooldownWakeupTime < now) overallCooldownWakeupTime = 0;
+            if (overallCooldownWakeupTime < now) overallCooldownWakeupTime = 0;
             return overallCooldownWakeupTime;
         }
     }
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1640,7 +1640,12 @@ public class SplitFileFetcherStorage {
         fetcher.clearCooldown();
     }
 
+    /** Returns -1 if the request is finished, otherwise the wakeup time. */
     public long getCooldownWakeupTime(long now) {
+        // It is OK to use the main lock here.
+        // Cooldown is a separate lock because updating it is tricky.
+        // FIXME maybe set cooldown when finish and avoid this.
+        if(hasFinished()) return -1;
         synchronized(cooldownLock) {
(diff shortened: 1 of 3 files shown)


toad:
This is an essential bugfix, please merge it. I've dropped #352 for now and will work on a new branch to make a single SendableGet/SendableInsert (e.g. SplitFileFetcherGet) per segment, which eliminates all the messy cooldown code, however IMHO the code as present is viable, so that cleanup can be added in a future version.

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1640,7 +1640,12 @@ public void maybeClearCooldown() {
         fetcher.clearCooldown();
     }
 
+    /** Returns -1 if the request is finished, otherwise the wakeup time. */
     public long getCooldownWakeupTime(long now) {
+        // It is OK to use the main lock here.
```
Thynix: In what way does this use a lock, and what is the main lock?

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1640,7 +1640,12 @@ public void maybeClearCooldown() {
...
+        // Cooldown is a separate lock because updating it is tricky.
+        // FIXME maybe set cooldown when finish and avoid this.
+        if(hasFinished()) return -1;
```
Thynix: Space between `if` and `(` to meet the [style guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace).

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1640,7 +1640,12 @@ public void maybeClearCooldown() {
         fetcher.clearCooldown();
     }
 
+    /** Returns -1 if the request is finished, otherwise the wakeup time. */
     public long getCooldownWakeupTime(long now) {
+        // It is OK to use the main lock here.
+        // Cooldown is a separate lock because updating it is tricky.
+        // FIXME maybe set cooldown when finish and avoid this.
```
Thynix: Avoid what?

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1760,4 +1760,17 @@ public void clearCooldown() {
         this.callback.clearCooldown();
     }
 
+    /** @return -1 if the insert has finished, 0 if has blocks to send, otherwise Long.MAX_VALUE. */
+    public long getWakeupTime(ClientContext context, long now) {
+        // As with the fetcher, it is safe to use the main lock here, cooldown is separate because
+        // it's messy calculating it.
+        // FIXME set the flag when calculating it.
```
Thynix: Same questions as above.

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1760,4 +1760,17 @@ public void clearCooldown() {
...
+        if(hasFinished())
```
Thynix: Needs a space.

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1760,4 +1760,17 @@ public void clearCooldown() {
...
     }
 
+    /** @return -1 if the insert has finished, 0 if has blocks to send, otherwise Long.MAX_VALUE. */
+    public long getWakeupTime(ClientContext context, long now) {
+        // As with the fetcher, it is safe to use the main lock here, cooldown is separate because
+        // it's messy calculating it.
+        // FIXME set the flag when calculating it.
+        if(hasFinished()) 
+            return -1;
+        if(noBlocksToSend())
```
Thynix: Needs a space.

toad:
Updated, please review. Thanks.

Thynix:
Looks good, thanks. I will merge this during my next chunk of Freenet time.

bertm:
As a general note on this part of the code: this could use some more documentation on the semantics of the values returned (i.e. do they represent times relative to `now`?). I'm not asking for that to be done in this pull request, though.

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1760,4 +1762,16 @@ public void clearCooldown() {
...
     }
 
+    /** @return -1 if the insert has finished, 0 if has blocks to send, otherwise Long.MAX_VALUE. */
+    public long getWakeupTime(ClientContext context, long now) {
+        // LOCKING: hasFinished() uses (this), separate from cooldownLock.
+        // It is safe to use both here (on the request selection thread), one after the other.
+        if (hasFinished()) 
+            return -1;
+        if (noBlocksToSend())
+            return Long.MAX_VALUE;
```
bertm: It looks like this introduces a race condition. Consider the following execution:

Thread1: hasFinishes() → false
Thread2: arbitrary sequence of operations that finish the insert (possible, since no locks are taken)
Thread1: noBlocksToSend() → true; returns `Long.MAX_VALUE` where `-1` is expected, since the request is now finished

A `synchronized(coolDownBlock) { synchronized(this) { ... }}` surrounding this entire function may avoid this. `chooseBlock()` also takes locks in this order.

Note: I'm not sure whether this race condition is actually harmful, I'm merely pointing out that it exists.

toad: The function is called inside the lock on the whole request selection tree (in ClientRequestSelector). So in theory thread 2 may set the finished flag, but then it will have to wait for the first thread to complete before removing it from the selection tree.

The bug that I saw was that inserts were finishing and somehow not getting removed from the request selection tree (indicating some other bug / race condition?), and had noBlocksToSend = false (because we were not selecting blocks), so it would return 0, so ClientRequestSelector chooses the request, and it can't create a block to send, so it blocked any other inserts from running.

The simplest solution to that bug is to make getWakeupTime() implement the contract, that is, return -1 when the request has finished already. This does not necessarily indicate an error, so there may be another bug somewhere...

This will be much, much simpler after I have made SplitFileFetcherGet/SplitFileInserterSender be per-segment, which would involve only the segment lock.

bertm: Thanks for the explanation on the expected usage of this code path. In that case, all should be fine.

bertm:
Looks good to me.

PR #352, opened 2015-04-14
Optimise wakeup times

toad:
Fuller version of pull request #351. Resolves some of the locking issues, and adds a bit of off-thread paranoia, but is basically an optimisation. Please consider carefully. Multiple locks and nested locking in both versions, this may actually be cleaner.

Either way, we need this extensively tested before deployment - but one or other must be included in 1468 IMHO.

### DIFF
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -177,7 +177,12 @@ public class SplitFileFetcherStorage {
     final int cooldownTries;
     /** Cooldown lasts this long for each key. */
     final long cooldownLength;
-    /** Only set if all segments are in cooldown. */
+    /** Separate lock for cooldown operations, which must be serialized. Must be taken *BEFORE*
+     * segment locks. DO NOT TAKE THIS LOCK WHILE HOLDING ANY OTHER LOCK! */
+    private final Object cooldownLock = new Object();
+    /** Only set if all segments are in cooldown. Will be either 0 if not in cooldown, -1 if 
+     * finished, or a time at which the request will become fetchable again. 
+     * LOCKING: Protected by cooldownLock, be careful! */
     private long overallCooldownWakeupTime;
     final CompatibilityMode finalMinCompatMode;
     
@@ -1594,10 +1599,6 @@ public class SplitFileFetcherStorage {
             
         });
     }
-
-    /** Separate lock for cooldown operations, which must be serialized. Must be taken *BEFORE*
-     * segment locks. */
-    private final Object cooldownLock = new Object();
 
     /** Called when a segment goes into overall cooldown. */
     void increaseCooldown(SplitFileFetcherSegmentStorage splitFileFetcherSegmentStorage,
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1640,7 +1640,12 @@ public class SplitFileFetcherStorage {
         fetcher.clearCooldown();
     }
 
+    /** Returns -1 if the request is finished, otherwise the wakeup time. */
     public long getCooldownWakeupTime(long now) {
+        // It is OK to use the main lock here.
+        // Cooldown is a separate lock because updating it is tricky.
+        // FIXME maybe set cooldown when finish and avoid this.
+        if(hasFinished()) return -1;
         synchronized(cooldownLock) {
             if(overallCooldownWakeupTime < now) overallCooldownWakeupTime = 0;
             return overallCooldownWakeupTime;
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1604,7 +1604,7 @@ public class SplitFileFetcherStorage {
 
             @Override
             public boolean run(ClientContext context) {
-                maybeClearCooldown();
+                maybeClearCooldownInner();
                 fetcher.restartedAfterDataCorruption();
                 return false;
             }
@@ -1650,6 +1650,18 @@ public class SplitFileFetcherStorage {
     /** Called when a segment exits cooldown e.g. due to a request completing and becoming 
      * retryable. Must NOT be called with segment locks held. */
     public void maybeClearCooldown() {
+        jobRunner.queueNormalOrDrop(new PersistentJob() {
+
+            @Override
+            public boolean run(ClientContext context) {
+                maybeClearCooldownInner();
+                return false;
+            }
+            
+        });
+    }
+
+    protected void maybeClearCooldownInner() {
         synchronized(cooldownLock) {
             if(overallCooldownWakeupTime == -1) return;
             if(overallCooldownWakeupTime == 0 || 
--- a/src/freenet/client/async/SplitFileFetcherStorage.java
+++ b/src/freenet/client/async/SplitFileFetcherStorage.java
@@ -1149,6 +1149,9 @@ public class SplitFileFetcherStorage {
                     if(succeeded) return false;
                     succeeded = true;
                 }
+                synchronized(cooldownLock) {
+                    overallCooldownWakeupTime = -1;
+                }
                 fetcher.onSuccess();
                 return true;
             }
@@ -1356,6 +1359,9 @@ public class SplitFileFetcherStorage {
             
             @Override
             public boolean run(ClientContext context) {
+                synchronized(cooldownLock) {
+                    overallCooldownWakeupTime = -1;
+                }
                 fetcher.fail(e);
                 return true;
             }
@@ -1377,6 +1383,9 @@ public class SplitFileFetcherStorage {
 
             @Override
             public boolean run(ClientContext context) {
+                synchronized(cooldownLock) {
+                    overallCooldownWakeupTime = -1;
+                }
                 fetcher.failOnDiskError(e);
                 return true;
             }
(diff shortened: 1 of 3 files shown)


src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -177,7 +177,12 @@
     final int cooldownTries;
     /** Cooldown lasts this long for each key. */
     final long cooldownLength;
-    /** Only set if all segments are in cooldown. */
+    /** Separate lock for cooldown operations, which must be serialized. Must be taken *BEFORE*
+     * segment locks. DO NOT TAKE THIS LOCK WHILE HOLDING ANY OTHER LOCK! */
```
Thynix: Why can't this lock be taken while holding any other lock? It'd also be nice for this comment to include less all-caps shouting.

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -177,7 +177,12 @@
...
+    private final Object cooldownLock = new Object();
+    /** Only set if all segments are in cooldown. Will be either 0 if not in cooldown, -1 if 
+     * finished, or a time at which the request will become fetchable again. 
+     * LOCKING: Protected by cooldownLock, be careful! */
```
Thynix: How about taking out "be careful!" and maybe even "LOCKING:"? I'd be fine with a blank line between the above and the locking documentation.

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1144,6 +1149,9 @@ public boolean run(ClientContext context) {
                     if(succeeded) return false;
                     succeeded = true;
                 }
+                synchronized(cooldownLock) {
+                    overallCooldownWakeupTime = -1;
```
Thynix: A restructuring to avoid using a magic value would probably be extensive, so I won't propose doing that right now. What do you think of a FETCH_FINISHED constant that's -1?

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1144,6 +1149,9 @@ public boolean run(ClientContext context) {
...
+                }
                 fetcher.onSuccess();
```
Thynix: Is it okay to set overallCooldownWakeupTime to completed before calling `onSuccess()` / `fail()` / `failOnDiskError()` like this? I'd be concerned about something seeing the request as completed before the success / failure is marked. Does that not happen, or is it not a problem, and why?

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1385,6 +1399,9 @@ public void failOnDiskError(final ChecksumFailedException e) {
 
             @Override
             public boolean run(ClientContext context) {
+                synchronized(cooldownLock) {
+                    overallCooldownWakeupTime = -1;
+                }
```
Thynix: This `PersistentJob` is duplicated. Can you break it out in a separate commit please?

src/freenet/client/async/SplitFileFetcherStorage.java
```diff
@@ -1632,7 +1650,20 @@ public boolean run(ClientContext context) {
     /** Called when a segment exits cooldown e.g. due to a request completing and becoming 
      * retryable. Must NOT be called with segment locks held. */
     public void maybeClearCooldown() {
+        jobRunner.queueNormalOrDrop(new PersistentJob() {
```
Thynix: I'm confused - why bother clearing the cooldown off the thread? Were there performance problems? Did benchmarks indicate acquiring the `cooldownLock` was a bottleneck or something?

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -196,7 +196,12 @@
         NOT_STARTED, STARTED, ENCODED_CROSS_SEGMENTS, ENCODED, GENERATING_METADATA, SUCCEEDED, FAILED
     }
 
+    /** The current status of the insert. You must set finished if you set this to SUCCEEDED or 
+     * FAILED. */
     private Status status;
+    /** Set when status is set to SUCCEEDED or FAILED. Never unset. Really status is definitive, 
+     * but this is convenient to avoid locking (this). */
```
Thynix: Is locking `Status` not viable?

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -218,7 +223,16 @@
 
     private final int overallStatusLength;
     
+    /** Protects cooldown calculations and noBlocksToSend. You MUST NOT take this lock while 
+     * holding any other locks!!!
```
Thynix: Why?

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -218,7 +223,16 @@
...
+     * Cooldown calculation is tricky! Two major issues:
+     * 1. Consistency is essential because if we return Long.MAX_VALUE from getWakeupTime(), the
+     * insert will stall until it is explicitly rescheduled. That's the whole point, in fact.
```
Thynix: How does deferring the request until explicitly rescheduled interact with consistency? I don't think I understand.

src/freenet/client/async/SplitFileInserterStorage.java
toad: Sorry, it's rather tangled... When we choose a block, if we can't find a block to send, we compute the time at which there will next be a block to send. This is the cooldown time. The request selection structures (RandomGrabArray etc) summarise this across groups of requests and thus avoid having to check the whole tree every time. Hence if something changes, e.g. a request completes with a DNF or we encode a segment, we need to clear the cache and then reschedule the request / tell the RGAs about it.

So we have concurrency issues: 
- Thread A fails to choose a block and sets cooldown to MAX_VALUE.
- Thread B completes a request and sets cooldown to 0
=> If they complete in the wrong order, the request will stall. So we need locking.

But we can't just use (this) because we need to check *all* the segments before we conclude that there are no blocks to send. So using (this) would probably cause deadlocks. Similarly the warnings about not taking cooldownLock while holding other locks - deadlocks could result, because when recomputing it we take cooldownLock and then the segment-level locks.

On top of all that, in getWakeupTime(), it'd be nice not to have to take (this) and the unlock it, and then take (cooldownLock) and unlock that. It *SHOULD* be safe to do this, which is what the first version of the patch did, because it's only run on the request selection thread. Overloading the cooldown time (just as it is overloaded in the getWakeupTime API) is an optimisation to avoid having to take two locks (one at a time). It's safe because we only go to "finished" once, and it's okay for there to be a minor delay between finishing and being removed from the RGA's. Also, it is supposed to be removed explicitly... possibly there's a concurrency issue there too...

The best solution would probably be to have a separate getter for each segment, and let the RGA's do the aggregation. That would require a larger refactoring and would use slightly more memory (and I understand people are already complaining about memory?).

Thoughts?

toad:
Okay, the correct solution to this is to break out a separate SendableInsert/SendableGet for each segment. However that would be a largish refactoring. So I suggest you merge the other pull request.

toad:
I have a branch which changes the code to use a single getter (SplitFileFetcherGet/InserterSender) per segment rather than per splitfile. This improves the code a lot, and it is worth the smallish cost in memory. It is backwards compatible; it does not block releasing, unless this stuff really worries you.

I will push this when it is finished (I haven't done inserts yet). In the meantime please merge the other version of this fix (#351).

src/freenet/client/async/SplitFileInserterStorage.java
Thynix: I'm not familiar enough with the code to have an opinion yet. How much more memory usage? I don't get the impression that Fred itself is particularly memory-hungry - it always seems to be plugins or a huge queue causing problems. What complaints are you referring to? Are there bugs filed? Can we change to separate getters for each segment while maintaining API backwards compatibility? If not then I'm not interested in the approach.

toad: Yes, the whole request queueing structure including the getters is strictly in-RAM, so we can change from per-splitfile getters to per-segment getters easily. It won't cost a lot of memory as the segment (and the bloom filter for the segment) is IIRC much bigger than the getter.

What I am referring to is all the complexity I tried to explain above. Selecting blocks at the level of a splitfile involves a lot of ugly unnecessary code (SplitFileFetcherStorage.chooseBlock()) that duplicates functionality elsewhere (the actual queue e.g. RandomGrabArray), is inefficient (read the above method!) and involves some very messy locking. Moving to per-segment getters would eliminate a lot of unnecessary code, reduce CPU usage, and avoid messy nested locking.

bertm: > I don't get the impression that Fred itself is particularly memory-hungry - it always seems to be plugins or a huge queue causing problems.

… or a large data store (remember the 1/3000th rule of thumb for the bloom filters), or a large number of threads on 64-bit JVMs (~1MiB per thread), a lot of running requests (multi-tab browsing and image-heavy sites), and some minor effects from plenty of other areas.

Either way, I'm all for per-segment getters. The slightly greater memory usage is well worth the reduced code complexity* and locking IMHO. I guess this should be a separate discussion on the bugtracker instead of something blocking this pull request.

* not to mention the non-deterministic worse-than-linear algorithmic complexity of `chooseBlock()`, I'll have a fix incoming for that.

toad: See my branch refactor-one-getter-per-segment. But it only does requests so far, not inserts, needs re-cherry-picking against official next and needs cleaning up (and is relatively large).

PR #353, opened 2015-04-15
Fix bug 6459 flaky unit test

toad:
Fixes bug #6459.

### DIFF
--- a/test/freenet/client/async/SplitFileInserterStorageTest.java
+++ b/test/freenet/client/async/SplitFileInserterStorageTest.java
@@ -1386,13 +1386,8 @@ public class SplitFileInserterStorageTest extends TestCase {
             assertTrue(allSegmentsEncoding(storage));
         SplitFileInserterSegmentStorage segment = storage.segments[0];
         assertTrue(memoryLimitedJobRunner.getRunningThreads() > 0);
+        data.waitForWaiting(); // Wait until the segments are actually in pread().
         segment.onFailure(0, new InsertException(InsertExceptionMode.INTERNAL_ERROR));
-        try {
-            Thread.sleep(100);
-        } catch (InterruptedException e1) {
-            // Ignore.
-        }
-        data.waitForWaiting();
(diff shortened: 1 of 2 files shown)


test/freenet/client/async/SplitFileInserterStorageTest.java
```diff
@@ -1365,8 +1365,11 @@ public void testCancelAltCrossSegment() throws IOException, InsertException, Mis
         Random r = new Random(0xb395f44d);
         testCancelAlt(r, CHKBlock.DATA_LENGTH*128*21);
     }
-    
+
     private void testCancelAlt(Random r, long size) throws IOException, InsertException {
+        // FIXME tricky to wait for "all threads are in pread()", when # threads != # segments.
+        // So just set max threads to 1 (only affects this test).
+        memoryLimitedJobRunner.setMaxThreads(1);
```
Thynix: Does doing this limit the thoroughness of the test?

Thynix:
Do I understand correctly that `BarrierRandomAccessBuffer` could benefit greatly from a semaphore?

toad:
It could use a Lock and two Condition's. But it's test code, I don't think it matters much.

test/freenet/client/async/SplitFileInserterStorageTest.java
toad: The current code only calls it with 1 segment and with 21 segments. In both cases we can only run one FEC job at once because the memory limit is 9MB. However if we were to call it with a size of 7MB or so, it might try to do 2 jobs at once. In which case the test would fail, because we are assuming that *all the running FEC threads* are in pread().

bertm:
IMHO semaphores could be used to achieve the same goal, but that would be about as messy as it is right now. I agree with @toad that this is not important for test code. In fact, I think the way it is coded right now allows for easier understanding of what the `BarrierRandomAccessBuffer` does for testing, than if it were coded using semaphores.

PR #355, opened 2015-04-21
Fix not unregistering inserts

toad:
Fixes the underlying bug that put me onto #351. However that is still valid, as we should implement the contract.

Testing procedure: (Appears to be working)
- Start a node *without* #351 but with this patch (cherry-pick it).
- Start a bunch of inserts at priority 3 and a bunch at priority 4.
- After the inserts at 3 finish, the inserts at 4 should proceed.
- Without this patch, this does not happen. At least sometimes it doesn't. That's what put me on to #351...

### DIFF
--- a/src/freenet/client/async/SplitFileInserter.java
+++ b/src/freenet/client/async/SplitFileInserter.java
@@ -46,10 +46,12 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
     private final boolean freeData;
     /** The RAF that stores check blocks and status info, used and created by storage. */
     private final LockableRandomAccessBuffer raf;
-    /** Stores the state of the insert and does most of the work. */
-    private transient SplitFileInserterStorage storage;
-    /** Actually does the insert */
-    private transient SplitFileInserterSender sender;
+    /** Stores the state of the insert and does most of the work. 
+     * Created in onResume() or in the constructor, so must be volatile. */
+    private transient volatile SplitFileInserterStorage storage;
+    /** Actually does the insert.
+     * Created in onResume() or in the constructor, so must be volatile. */
+    private transient volatile SplitFileInserterSender sender;
     /** Used any time a callback from storage needs us to do something higher level */
     private transient ClientContext context;
     /** Is the insert real-time? */
--- a/src/freenet/client/async/SplitFileInserter.java
+++ b/src/freenet/client/async/SplitFileInserter.java
@@ -238,8 +238,7 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
     }
 
     protected void unregisterSender() {
-        if (sender != null)
-            sender.unregister(context, parent.getPriorityClass());
+        sender.unregister(context, parent.getPriorityClass());
     }
 
     protected void reportMetadata(Metadata metadata) {
--- a/src/freenet/client/async/SplitFileInserter.java
+++ b/src/freenet/client/async/SplitFileInserter.java
@@ -219,6 +219,7 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
             @Override
             public boolean run(ClientContext context) {
                 if(logMINOR) Logger.minor(this, "Succeeding on "+SplitFileInserter.this);
+                unregisterSender();
                 if(!(ctx.earlyEncode || ctx.getCHKOnly)) {
                     reportMetadata(metadata);
                 }
@@ -234,6 +235,11 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
         });
     }
 
+    protected void unregisterSender() {
+        if(sender != null)
+            sender.unregister(context, parent.getPriorityClass());
+    }
+
     protected void reportMetadata(Metadata metadata) {
         // Splitfile insert is always reportMetadataOnly, i.e. it always passes the metadata back 
         // to the parent SingleFileInserter, which will write it to a Bucket and probably insert it.
@@ -246,6 +252,7 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
 
             @Override
             public boolean run(ClientContext context) {
+                unregisterSender();
                 raf.close();
                 raf.free();
                 originalData.close();
--- a/src/freenet/client/async/SplitFileInserter.java
+++ b/src/freenet/client/async/SplitFileInserter.java
@@ -238,7 +238,7 @@ public class SplitFileInserter implements ClientPutState, Serializable, SplitFil
     }
 
     protected void unregisterSender() {
-        if(sender != null)
+        if (sender != null)
             sender.unregister(context, parent.getPriorityClass());
     }
 
--- a/src/freenet/client/async/SplitFileFetcher.java
+++ b/src/freenet/client/async/SplitFileFetcher.java
@@ -74,6 +74,7 @@ public class SplitFileFetcher implements ClientGetState, SplitFileFetcherStorage
         Logger.registerClass(SplitFileFetcher.class);
     }
 
+    /** Stores the progress of the download, including the actual data, in a separate file. */
     private transient SplitFileFetcherStorage storage;
     /** Kept here so we can resume from storage */
     private LockableRandomAccessBuffer raf;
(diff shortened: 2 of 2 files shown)


toad:
I have started some testing, will report back.

toad:
Testing going well. Please consider merging this patch.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -112,9 +112,13 @@ public void cancel(ClientContext context) {
...
         storage.start();
+        SplitFileInserterSender s;
+        synchronized(this) {
+            s = sender;
+        }
         if(!ctx.getCHKOnly) {
-            sender.clearWakeupTime(context);
-            sender.schedule(context);
+            s.clearWakeupTime(context);
+            s.schedule(context);
```
Thynix: Are these safe functions thread-safe? It's not clear to me from reading them that they are. It'd be obviously correct if the lock was held for both these calls instead.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -139,7 +143,9 @@ public void onResume(ClientContext context) throws InsertException, ResumeFailed
                     context.getChkInsertScheduler(realTime).fetchingKeys(), context.persistentFG, 
                     context.persistentFileTracker, context.getPersistentMasterSecret());
             storage.onResume(context);
-            this.sender = new SplitFileInserterSender(this, storage);
+            synchronized(this) {
+                this.sender = new SplitFileInserterSender(this, storage);
```
Thynix: Why does assigning this reference need the lock? (Although I own a copy of _Java Concurrency In Practice_ I have yet to read it.)

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -273,7 +290,14 @@ public void onShutdown(ClientContext context) {
...
     public void clearCooldown() {
-        sender.clearWakeupTime(context);
+        SplitFileInserterSender s;
+        synchronized(this) {
+            s = sender;
+        }
+        if(s != null)
+            s.clearWakeupTime(context);
+        else
+            Logger.error(this, "Sender is null in clearCooldown()!", new Exception());
```
Thynix: Why shouldn't that happen?

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -273,7 +290,14 @@ public void onShutdown(ClientContext context) {
 
     @Override
     public void clearCooldown() {
-        sender.clearWakeupTime(context);
+        SplitFileInserterSender s;
+        synchronized(this) {
+            s = sender;
+        }
+        if(s != null)
```
Thynix: Space between `if` and `(`.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -273,7 +290,14 @@ public void onShutdown(ClientContext context) {
 
     @Override
     public void clearCooldown() {
-        sender.clearWakeupTime(context);
+        SplitFileInserterSender s;
+        synchronized(this) {
```
Thynix: Space between `synchronized` and `(`.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -234,6 +241,15 @@ public boolean run(ClientContext context) {
         });
     }
 
+    protected void unregisterSender() {
+        SplitFileInserterSender s;
+        synchronized(this) {
+            s = sender;
+        }
+        if(s != null)
```
Thynix: Space.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -234,6 +241,15 @@ public boolean run(ClientContext context) {
         });
     }
 
+    protected void unregisterSender() {
+        SplitFileInserterSender s;
+        synchronized(this) {
```
Thynix: Space.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -139,7 +143,9 @@ public void onResume(ClientContext context) throws InsertException, ResumeFailed
                     context.getChkInsertScheduler(realTime).fetchingKeys(), context.persistentFG, 
                     context.persistentFileTracker, context.getPersistentMasterSecret());
             storage.onResume(context);
-            this.sender = new SplitFileInserterSender(this, storage);
+            synchronized(this) {
```
Thynix: Space.

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -112,9 +112,13 @@ public void cancel(ClientContext context) {
     public void schedule(ClientContext context) throws InsertException {
         cb.onBlockSetFinished(this, context);
         storage.start();
+        SplitFileInserterSender s;
+        synchronized(this) {
```
Thynix: Space.

Thynix:
It's not clear to me that it is sufficient for correct concurrent behavior to grab / set the reference while holding a lock. It's not clear to me that the `SplitFileInsertSender` is thread-safe. Is it?

src/freenet/client/async/SplitFileInserter.java
toad: The lock here isn't really necessary at all, it's mainly to shut up code analysis tools. sender is only set when we create a splitfile fetcher or when we load one from disk. Maybe we could make it an AtomicObject or volatile?

Both of the methods called are indeed thread-safe, as is more or less everything in the client layer. In detail:

clearWakeupTime(): If we don't have an RGA then we're not in the queue, which means we are not on the queue i.e. we are not registered. If so, then the schedule() will fix this. If we are on the queue, then RGA.clearWakeupTime() takes the lock on the ClientRequestSelector (i.e. the entire request queue for the given type of requests e.g. bulk CHK inserts), marks the request as sendable, and if necessary clears the cooldown for parents too (i.e. if a branch of the tree contains a sendable request, then that branch is itself sendable). In the event of a race condition (the RGA is no longer in the tree), we propagate this upwards anyway, which means that branch of the tree will be tried even though it doesn't have any requests to send; this is fine, it will be corrected automatically by the next attempt to find a request to send.

schedule() is very similar to schedule() on other requests, and is thread safe. If there is an RGA then we are already scheduled, that is we are on the request queue, so we don't need to be scheduled again. If another thread is unregistering it there are 2 possibilities:
1. It is being explicitly unregistered, by the means given in this patch. If so, it's finished, we don't need it to be scheduled. (If we restart the insert the ClientPutter will create a new SplitFileInserter).
2. It is removed from the RGA because it is not sendable. IIRC this only happens if the request is actually cancelled i.e. it returns -1 from getWakeupTime(), in which case there is a race condition but removing it from the queue is correct. This can happen if the user cancels the insert while it is still encoding, for example. If getWakeupTime() returns a positive value then it is kept on the queue, but marked as not fetchable until some point in the future; clearWakeupTime() clears this e.g. when requests fail and become retryable.

src/freenet/client/async/SplitFileInserter.java
toad: Because it can be accessed from multiple threads. However, it is only set on starting a new request and in onResume() (called when restarting a request after serialising in from client.dat). It sadly can't be final, but it won't change once it's set.

src/freenet/client/async/SplitFileInserter.java
toad: Yes an assert would make more sense.

src/freenet/client/async/SplitFileInserter.java
Thynix: I don't follow why that makes locking necessary. The reads of it are not synchronized, so what does locking this write prevent? Under what circumstances does an insert not yet resumed have methods called from other threads?

src/freenet/client/async/SplitFileInserter.java
```diff
@@ -234,6 +237,11 @@ public boolean run(ClientContext context) {
         });
     }
 
+    protected void unregisterSender() {
+        if (sender != null)
```
Thynix: Why have this check here? I don't see a code path that can result in `sender` being null, which the check implies could happen. If this check is necessary it needs locking to avoid a race condition.

PR #356, opened 2015-04-23
wizard l10n: suggest store values between 100MiB and 20GiB

ArneBab:
This is to avoid the case where users configure 200GiB and then complain about hours of high IO (which can slow the computer to a crawl). This happened with a user who used NTFS.

For xor: Linux IO sucks. It sucks. It really sucks. it’s the one thing which can easily slow down my system to a crawl. Windows IO sucks, too. IO sucks. 

Doing lots of IO efficiently isn’t easy. So we should not incite users to configure Freenet in a way which will give them an awful user experience for hours.


bertm:
IMHO the English phrasing is too restrictive to the end user. The given range of 100MiB to 20GiB should merely be a suggestion of recommended data store sizes, and should be phrased as such, not as a command to the user. The German string sounds a lot more friendly and suggesting than the English counterpart.

Also, I'd like to propose to raise the lower bound on the suggested range to 1GiB to help the network a little. Hard disk space is cheap nowadays, pretty much everyone has a few GiB left unused, and pre-allocating 1GiB should still be finished in less than a minute.

Suggested alternative phrasing:
```properties
FirstTimeWizardToadlet.datastoreSizeLong=Please select a size for your datastore. Recommended values are between 1GiB and 20GiB. A smaller datastore can result in problems downloading files, a larger datastore is beneficial to the network but might slow down your computer due to high filesystem load.
```

toad:
1. A better solution may be #29. This will dramatically reduce the overall store and thus data retention since if there is any churn - which there is. We should do #29 first. However there are packaging issues with it as it needs JNA.
2. This is not a good implementation, you need to change the actual wizard page to not suggest a default over 20GB.

ArneBab:
The wizard should allow setting values >20GiB, because the user might want to use a bigger store despite the warning. I reduced the number of values >20GiB to 50, 200 and 500 to make them smaller “visually”.

Also I adjusted automatic the size suggestion: stays < 20GiB, but takes a larger part of smaller disks (up to 25% of free space, gradually reduced to 10%).

nextgens:
Why not compromise here? Start with what Arne suggests ... and then spawn a useralert asking the user whether he wants to increase the datastore size if it reaches a certain throttle / or the node has a decent uptime?

I'd even be in favour of getting rid of that step in the wizard and starting small.

nextgens:
in fact if it's the preallocation we should use fallocate() ... the JVM doesn't expose it but we could with JNA...

https://groups.google.com/forum/#!topic/mechanical-sympathy/UMrKt75yOmg
On windows it's more tricky... like usual:
https://stackoverflow.com/questions/7970333/how-do-you-pre-allocate-space-for-a-file-in-c-c-on-windows

ArneBab:
What is needed to get this merged?

nextgens:
What's the status on that one @Thynix ? I agree with Arne, this should be merged sooner rather than later.

src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
```diff
@@ -154,31 +151,38 @@ private long maxDatastoreSize() {
...
+            return -1;
+
+        long freeSpace = core.node.getStoreDir().getUsableSpace();
+
+        if(freeSpace <= 0) {
+            return -1;
+        } else {
+            long shortSize;
+            long oneGiB = 1024 * 1024 * 1024L;
+            // Maximum for Freenet: 256GB. That's a 128MiB bloom filter.
```
Thynix: Does the maximum come from the default memory limit or what?

src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
```diff
@@ -154,31 +151,38 @@ private long maxDatastoreSize() {
...
+            long bloomFilter128MiBMax = 256 * oneGiB;
+            // Maximum to suggest to keep Disk I/O managable. This
+            // value might need revisiting when hardware or
+            // filesystems change.
+            long diskIoMax = 20 * oneGiB;
+            if(freeSpace / 50 > oneGiB) { // 50GB+ => 10%, but at least 10GiB
```
Thynix: May I suggest formatting these more like

    if (freeSpace > 50 * oneGiB) { // use 10% free space, or at least 10 GiB

I can do that if you like.

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -541,7 +541,7 @@ FirstTimeWizardToadlet.continue=Continue
 FirstTimeWizardToadlet.currentPrefix=Current size:
 FirstTimeWizardToadlet.currentSpeed=Current limit:
 FirstTimeWizardToadlet.datastoreSize=Datastore size
-FirstTimeWizardToadlet.datastoreSizeLong=Please select a size for your datastore. The datastore acts like a cache; storing data for the network will help you to get better throughput when requesting popular files. The more space you can afford the better it is for the community and the faster your node and especially your downloads will go.
+FirstTimeWizardToadlet.datastoreSizeLong=Please select a size for your datastore. Recommended values lie between 1GiB and 20GiB. A smaller datastore can result in problems downloading files, a larger datastore is beneficial to the network but might slow down your computer due to high filesystem load, especially in the first few hours after installation. The datastore acts like a cache; storing data for the network will help you to get better throughput when requesting popular files. The more space you can afford the better it is for the community and the faster your node and especially your downloads will go.
```
Thynix: s/, a larger/, and a larger/

I'm also wondering about removing some of the additional elaboration as overly verbose.

src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
ArneBab: it would be nice if you could do it

PR #357, opened 2015-04-25
Fix recovery condition

toad:
Should improve recovery behaviour: We were recovering old requests from the .crypt file when it exists, but not using the .bak.crypt when we needed it.

Not tested, will look into this myself when I get around to it.

### DIFF
--- a/src/freenet/client/async/ClientLayerPersister.java
+++ b/src/freenet/client/async/ClientLayerPersister.java
@@ -233,10 +233,10 @@ public class ClientLayerPersister extends PersistentJobRunnerImpl {
         if(clientDatExists) {
             innerLoad(loaded, makeBucket(dir, baseName, false, null), noSerialize, context, requestStarters, random);
         }
-        if(clientDatCryptExists && loaded.needsMore()) {
+        if(clientDatCryptExists) {
             innerLoad(loaded, makeBucket(dir, baseName, false, encryptionKey), noSerialize, context, requestStarters, random);
         }
-        if(clientDatBakExists) {
+        if(clientDatBakExists && loaded.needsMore()) {
             innerLoad(loaded, makeBucket(dir, baseName, true, null), noSerialize, context, requestStarters, random);
         }
         if(clientDatBakCryptExists && loaded.needsMore()) {
--- a/src/freenet/client/async/ClientLayerPersister.java
+++ b/src/freenet/client/async/ClientLayerPersister.java
@@ -230,16 +230,16 @@ public class ClientLayerPersister extends PersistentJobRunnerImpl {
         }
         boolean failedSerialize = false;
         PartialLoad loaded = new PartialLoad();
-        if(clientDatExists) {
+        if (clientDatExists) {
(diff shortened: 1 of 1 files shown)


src/freenet/client/async/ClientLayerPersister.java
```diff
@@ -230,16 +230,16 @@ private boolean innerSetFilesAndLoad(boolean noSerialize, File dir, String baseN
         }
         boolean failedSerialize = false;
         PartialLoad loaded = new PartialLoad();
-        if(clientDatExists) {
+        if (clientDatExists) {
             innerLoad(loaded, makeBucket(dir, baseName, false, null), noSerialize, context, requestStarters, random);
         }
-        if(clientDatCryptExists && loaded.needsMore()) {
+        if (clientDatCryptExists) {
```
bertm: If a corrupt `client.dat.crypt` exists, even when encryption is disabled, it will now always mess up the loaded `client.dat`.

Thynix:
> We should load the encrypted version if it exists whether or not we have errors.

Meaning a non-backup encrypted version? When would there be both encrypted and non-encrypted versions?

toad:
Something can go wrong when changing from encrypted to non-encrypted and vice versa. We want to be able to recover from this case. Hence we need to try both. Also more generally our encryption setting doesn't necessarily match what's on disk, so we have to try both. In general both shouldn't exist at the same time but it can happen.

src/freenet/client/async/ClientLayerPersister.java
toad: See my comments to Steve.

bertm:
> Also more generally our encryption setting doesn't necessarily match what's on disk, so we have to try both.

We expect that our encryption setting matches what's on disk. Trying both is a recovery mechanism, just like the `.bak` files. My objection is that we now always try to load the non-matching file (if it exists), even if the matching file would load just fine. Instead, we should first try the matching file, and only proceed loading backups and alternatives (i.e. non-matching and its backup) if that fails.

(I do agree that this should not normally happen, and that existence of both matching and non-matching files can be considered an error.)

toad:
Okay so:
- If we are expecting it to be encrypted we should try the encrypted file first. (BUG, we always try the unencrypted one first at present)
- We should try the backup if the main file (for the base crypto setting) doesn't exist or is corrupted.

One difficulty: Should we try the backup or the opposite-encryption-setting file first? If an opposite-encryption-setting file exists, it means one of:
1) The settings were changed after shutdown, or
2) We were interrupted before we deleted the old file, or
3) Something wierd happened (e.g. permissions)

So do we want (proceeding only if needsMore() i.e. partial corruption or no file):
tryload(crypto,main)
tryload(!crypto,main)
tryload(crypto,backup)
tryload(!crypto,backup) ?
The second point would seem to suggest this?

bertm:
The loading sequence you suggest looks like the right approach to me.

Thynix:
Anyone care to implement it?

Thynix:
Any word on this?

PR #358, opened 2015-04-25
Linear complexity bounds on SplitFileInserterStorage.chooseBlock / SplitFileFetcherStorage.chooseRandomKey

bertm:
This rewrites `chooseBlock` to be functionally equivalent to the existing code, but linear in worst case complexity in terms of number of segments (as compared to non-terminating previously). In the most common case, where (almost) all segments yield blocks, this method maintains O(1) complexity.

The primary feature of this pull request is to replace the following line, which is obviously non-terminating in the worst case (although such an infinite run happens with zero probability):
```java
while(tried[segNo = random.nextInt(segments.length)]);
```

Additional costs in memory are insignificant: they are limited to one integer per segment to maintain a permutation of segment array indices to shuffle.

### DIFF
--- a/src/freenet/client/async/SplitFileInserterStorage.java
+++ b/src/freenet/client/async/SplitFileInserterStorage.java
@@ -1714,32 +1714,35 @@ public class SplitFileInserterStorage {
      * FIXME make SplitFileInserterSender per-segment, eliminate a lot of unnecessary complexity.
      */
     public BlockInsert chooseBlock() {
-        // FIXME this should probably use SimpleBlockChooser and hence use lowest-retry-count from each segment?
+        // FIXME this should probably use SimpleBlockChooser and hence use lowest-retry-count from
+        // each segment?
         // Less important for inserts than for requests though...
         synchronized(cooldownLock) {
-        synchronized(this) {
-            if(status == Status.FAILED || status == Status.SUCCEEDED || 
-                    status == Status.GENERATING_METADATA || failing != null) return null;
-        }
-        // Generally segments are fairly well balanced, so we can usually pick a random segment 
-        // then a random key from it.
-        for (int i = 0; i < segments.length; i++) {
-            // Randomly pick one of the remaining segments, using a partial Fisher–Yates shuffle of
-            // all segments indices in segmentIndexPermutation
-            int j = random.nextInt(segments.length - i) + i;
-            int chosen = segmentIndexPermutation[j];
-            segmentIndexPermutation[j] = segmentIndexPermutation[i];
-            segmentIndexPermutation[i] = chosen;
-
-            BlockInsert ret = segments[chosen].chooseBlock();
-            if(ret != null) {
-                noBlocksToSend = false;
-                return ret;
+            synchronized(this) {
+                if (status == Status.FAILED || status == Status.SUCCEEDED
+                        || status == Status.GENERATING_METADATA || failing != null) {
+                    return null;
+                }
             }
-        }
+            // Generally segments are fairly well balanced, so we can usually pick a random segment
+            // then a random key from it.
+            for (int i = 0; i < segments.length; i++) {
+                // Randomly pick one of the remaining segments, using a partial Fisher–Yates shuffle
+                // of all segments indices in segmentIndexPermutation
+                int j = random.nextInt(segments.length - i) + i;
+                int chosen = segmentIndexPermutation[j];
+                segmentIndexPermutation[j] = segmentIndexPermutation[i];
+                segmentIndexPermutation[i] = chosen;
 
-        noBlocksToSend = true;
-        return null;
+                BlockInsert ret = segments[chosen].chooseBlock();
+                if (ret != null) {
+                    noBlocksToSend = false;
+                    return ret;
+                }
+            }
+
+            noBlocksToSend = true;
+            return null;
         }
     }
     
--- a/src/freenet/client/async/SplitFileInserterStorage.java
+++ b/src/freenet/client/async/SplitFileInserterStorage.java
@@ -35,6 +35,7 @@ import freenet.node.KeysFetchingLocally;
 import freenet.support.HexUtil;
 import freenet.support.Logger;
 import freenet.support.MemoryLimitedJobRunner;
+import freenet.support.RandomArrayIterator;
 import freenet.support.Ticker;
 import freenet.support.api.Bucket;
 import freenet.support.api.BucketFactory;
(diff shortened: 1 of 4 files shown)


Thynix:
Oh my. Yeah - this is hugely preferable. I take it you've tested this as working?

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1699,34 +1717,29 @@ public BlockInsert chooseBlock() {
...
+            for (int i = 0; i < segments.length; i++) {
+                // Randomly pick one of the remaining segments, using a partial Fisher–Yates shuffle
+                // of all segments indices in segmentIndexPermutation
+                int j = random.nextInt(segments.length - i) + i;
+                int chosen = segmentIndexPermutation[j];
+                segmentIndexPermutation[j] = segmentIndexPermutation[i];
+                segmentIndexPermutation[i] = chosen;
+
+                BlockInsert ret = segments[chosen].chooseBlock();
+                if(ret != null) {
```
Thynix: The style guide [requires](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace) a space between `if` and `(`.

src/freenet/client/async/SplitFileInserterStorage.java
```diff
@@ -1699,34 +1717,29 @@ public BlockInsert chooseBlock() {
...
-        for(int count=1; count<segments.length; count++) {
-            while(tried[segNo = random.nextInt(segments.length)]);
-            tried[segNo] = true;
-            segment = segments[segNo];
-            ret = segment.chooseBlock();
-            if(ret != null) {
-                noBlocksToSend = false;
-                return ret;
+            synchronized(this) {
+                if(status == Status.FAILED || status == Status.SUCCEEDED ||
```
Thynix: Might as well fix the whitespace and braces here too.

bertm:
That assumption is correct. Apart from some standalone testing of the algorithm outside of Fred, I've tested a few inserts (up to 2 segments) and those still finished fine. I'll have some more inserts running overnight.

`SplitFileFetcherStorage` exhibits the same algorithmic characteristics (as do `RandomGrabArray` and related). @toad is implementing a proper solution to the former. Awaiting the completion thereof, in the meantime I could rather quickly adapt the it as well.  Any opinions?

(As for the latter, it is far more convoluted and therefore harder to patch, probably has some other fundamental problems (heck, `SectoredRandomGrabArray` is not even properly randomized AFAICS from scrolling through the code) but operates on the block level and is therefore more influential. That part of Fred needs to receive some love in the near future as well.)

toad:
The correct solution to this, as I've explained, is to get rid of SFFS.chooseBlock() and SFIS.chooseBlock(), by making the getter be per-segment. That solves a number of other serious problems at the same time.

The situation with RGA/SRGA is somewhat different. For a start, any segment getter that has completed won't be in the RGA at all. So getWakeupTime() returns a positive value only if the segment is in cooldown or it is currently fetching all of its blocks. The latter cannot be true of a large number of segments simultaneously because we have a small finite capacity for running requests. The former can be true, but is handled efficiently. Occasionally we call removeRandomExhaustiveSearch() on a large RGA, but this does useful work, removing cancelled requests (okay that indicates a race condition) and computing the overall cooldown time for the RGA, which then propagates up to the SRGA, so the RGA won't be called again for some time - either until clearCooldown (if all the requests are running) or until a specific time at which the first request exits cooldown.

So while it surely could be more efficient, without using more memory (and memory does matter here), IMHO it's not that bad, and we can live with the linear worst-case because it's normally followed by a cooldown. Also this is really critical code, which could break things very badly, so be careful.

As far as SRGA goes I don't know what you are talking about; it is intentional that we pick a random RGA and then ask that, we don't *want* to balance the requests.

bertm:
The overnight multi-segment insert tests also finished just fine.

> The correct solution to this, as I've explained […]
I understand that this is not *the* correct solution per se. This is merely a patch to provide a deterministic worst case complexity for this part of the code, with as little modifications as necessary. Reason I asked is because I know you are working on a proper solution, but this may be eligible for merging earlier. If you expect your changes to be ready for merging "soon", I won't adapt SFFS.

(S)RGA is, indeed, quite another piece of code. My main point was that RGA uses the same potentially non-terminating selection strategy (line 153+, however limited by `MAX_EXCLUDED` failures, which makes the selection strategy a bit weird), and that it can be rewritten to use a linear worst case selection mechanism.

As for SRGA `removeRandomExhaustive()`, that comment was directed to the selection of an sector from `grabArrays`: a random sector is chosen, and on failure (`getWakeupTime() > 0`) , the *next consecutive* sector is chosen instead of a next random one. This introduces a bias towards sectors situated directly after one or more sectors with non-zero wakeup times, i.e. the probability of a sector being selected is directly proportional to the number of non-zero wakeup time sectors preceding it, plus one. That's something for another patch set, though.

In-place shuffling is preferable wherever possible, since it does not require additional data structures to be maintained. I did not choose to implement it that way here, for I didn't dare to assume that no parts of SFIS that rely on the ordering of segments (or at least the consistency thereof).

toad:
RGA tries a few slots (MAX_EXCLUDED) and then if it can't find one it does an exhaustive search. So the worst case is linear; it does terminate. And we do need the exhaustive search if we are to enter cooldown.

I agree that SRGA.removeRandomExhaustive isn't entirely random. So shuffling might help, although in general if we get to removeRandomExhaustive there's a fairly good chance that we won't find anything and will go into cooldown; the case where we lose is if we have a lot of requests all on long cooldowns and a few that aren't. In this case we'd probably like to keep them sorted by wake-up time, possibly sorting lazily (with an integer for how many requests from the beginning are valid so we can do a random selection). Or reinvent the old cooldown queue by unregistering and then adding back - requiring a global structure with the ability to remove stuff prematurely. But in general I doubt that either class of optimisation is worthwhile.

The good news is none of these structures has any persistence issues... so you can optimise them without breaking anything.

My change to get rid of chooseBlock() isn't imminent, and when it is ready it'll be a largish patch. So in principle I have no objection to this pull request. In practice ... you don't complete the shuffle, so the first few slots get swapped with random segments, while the last part of the array gets moved a lot less; is this a problem?

bertm:
Ah, thanks, you comments clarify the purpose of `MAX_EXCLUDED` a lot. In that case there is no real problem with the random selection in RGA itself, except that it is not very readable.
I don't expect real world problems in SRGA either: in the worst case, it's just not as random as it's supposed to be. Worth fixing at some point, but not right now.

Not completing the shuffle is not a problem. The way this is implemented, each invocation of `chooseBlock()` we use a new random permutation, we just don't always complete the actual shuffle. The next run, we again choose a new permutation. Note that randomly permuting *any* list, no matter the existing ordering, always results in a fully randomly permuted list. That includes a partially permuted list, i.e. the result of a previous partial shuffle. 

Your observation that we swap the first slots far more often is correct. However, due to this swapping there will always be different elements at the first positions, hence all elements stay equally likely to be in the first slots. This Fisher–Yates shuffle is known to generate an unbiased random permutation.

I'll write up a quick fix for SFFS as well, to be included in this pull request.

bertm:
Updated SFFS as well, using a now separated `RandomArrayIterator` implementation for both SFIS and SFFS.

bertm:
Unit tests pass, testing on my node suggests this works just fine as well.

PR #359, opened 2015-04-25
Fix another unit test race condition.

toad:
Assert is bogus. The callback is called from a thread on the executor, which
might or might not have actually exited at this point. Does not affect the
other 2 assertions, which should both definitely be true; the first one is
the crucial bit, we don't get the actual failure until after the encode has
completed, but then we get a failure and not an encoded notice.


toad:
This is from bug #6459. Thanks operhiem1.

PR #360, opened 2015-04-26
Fix RGA.wakeupTime never getting set.

toad:
Please consider for the next-but-one build. And test!!!
(Thanks for bertm for finding this problem, and for contributing some additional cleanup)

RGA.wakeupTime was always 0. Functionally this should not have any great
impact, although clearly the code was wrong.

SRGA.removeRandomExhaustive calls getWakeupTime and then if that returns 0
(indicating not in cooldown), it calls RGA.removeRandom. Which will then
return a positive cooldown time if it's actually in cooldown - so SRGA will
set itself into cooldown on the basis of that value. We were always returning
0 from RGA.getWakeupTime(); this fixes that, which should speed up SRGA
slightly.

There could be a correctness issue if reduceWakeupTime is being called a lot.
Currently IIRC we call clearWakeupTime instead in the common case where a
request finishes and becomes sendable (because it failed).

NEEDS TESTING!!! Bugs in this code could cause SEVERE problems!

### DIFF
--- a/src/freenet/support/RandomGrabArray.java
+++ b/src/freenet/support/RandomGrabArray.java
@@ -246,28 +246,28 @@ public class RandomGrabArray implements RemoveRandom, RequestSelectionTreeNode {
 					continue;
 				}
 				boolean excludeItem = false;
-				long excludeTime = item.getWakeupTime(context, now);
-				if(excludeTime > 0) {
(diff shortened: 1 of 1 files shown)


Thynix:
Is this ready to be merged?

PR #361, opened 2015-05-03
Add gantros index and resort default bookmarks

ArneBab:
order: 

- nerdageddon (clean)
- gantros (up to date, ordered)
- linkageddon (up to date)
- enzo (because it is referenced in translations)


Thynix:
Oh yay a new index!

ArneBab:
In my opinion this also resolves the wish from FMS to focus on nice index-sites: the first index is one which hides the ugly parts.

ArneBab:
PS: With the new index, it feels like the community is slowly pulling itself together: We’re slower than a year ago, but more resilient. When users ask for features, it feels like we can provide them more easily than before because we’re less focussed on large reorganization and redoing tasks and more on doing what’s useful.

Bombe:
Shouldn’t we wait until the index has been up and running for at least a week before we include it in the default bookmarks? :)

ArneBab:
I think it will have been active for a few weeks when we get the release out. There’s still some plugin-work to do before we can release. So sure, no problem waiting a week before clicking merge ☺

ArneBab:
it’s been up for a week now. Can we merge this? ☺

Bombe:
No, it’s still far from any kind of maturity. The week I mentioned earlier was not meant literally and I had assumed that was clear. My bad.

Also I have found entries without title and any kind of clickable link.

Bombe:
And it can not handle UTF-8 characters correctly. And it’s missing pages such as the Sone homepage (which I would have expected in the “Software” category).

ArneBab:
damn, you’re right. That’s something different then. I disentangled this from my other bookmarks pull-requests.

ArneBab:
Gantros:
> Sone page is in "Freenet-Messaging", like in Enzo's index. I want to enable more than one category for site, but I can't now.
> I thought I fix the UTF-8 problem two weeks ago, in 
http://127.0.0.1:8888/Sone/viewPost.html?post=d5cb3c7e-6368-4b1d-ab0a-827c79827550

Thynix:
As of edition 21 I haven't noticed anything out of place. I'm up for adding it.

Bombe:
In the flogs category, there are still a couple of entries with broken encodings.

Thynix:
Edition 31 has dropped the bar of categories on the right? This may not be stable enough yet.

Thynix:
Even if it's having problems is it better than not having it?

ArneBab:
I think it is better to have it than not, yes. There will always be problems from time to time. I think it’s the activity which counts: users need to see that Freenet is alive.

Thynix:
Yup - also TheSeeker found the unescaped quote which caused the filter to remove the side bar. I've only seen a few entries with encoding problems, and in other cases it works correctly, so I'm okay with merging it now. Hopefully we can help Gantros get Infocalypse working so we can contribute code to the index.

PR #364, opened 2015-05-13
default bookmarks: add mempo

ArneBab:
Adds the mempo repo: It is a really cool usage of Freenet, and having it in the default bookmarks should solve their “is it up to date” issue. Yes, I know that we cannot do this for every project out there. But lets start by making one project work well and do the general perfect fix later.


src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -129,6 +130,7 @@ Bookmarks.Defaults.ShortDescription.Publish=How to publish web sites on freenet
 Bookmarks.Defaults.ShortDescription.FreesiteHOWTO=Another guide to publishing a website to Freenet
 Bookmarks.Defaults.ShortDescription.Sone=Social chat over Freenet
 Bookmarks.Defaults.ShortDescription.freenetdocwiki=Freenet documentation wiki
+Bookmarks.Defaults.ShortDescription.mempo=Hardened debian served over Freenet
```
Thynix: "Debian" should be capitalized in the short description too.

bertm:
I'm not in favor of merging this. I'm not convinced we need yet another default bookmark. Yes, this is a cool usage of Freenet, but I don't think the average Freenet user is interested in Mempo at all.

If the Mempo project somehow needs their Freesite in the bookmarks for part of their project to work, they should ask their users to add this bookmark themselves instead.

Thynix:
I agree with bertm - it makes more sense to me to ask Mempo users add the bookmark. I don't get the sense that most Freenet users are interested in finding a new distro necessarily, and including it in the default bookmarks is something of an endorsement.

ArneBab:
ok - in addition I found a way to ensure that this is no longer necessary for a well-working mempo (thanks to stackoverflow).

Thynix:
What is that way?

ArneBab:
The way is putting a script into /usr/local/bin/apt-get and making that script call apt-get after wget -O /dev/null http://127.…/USK@…/mempo/-42/

See http://unix.stackexchange.com/a/206784/34896

Thynix:
It's certainly a hack but it's easier than writing a site-fetching-and-hosting plugin.

PR #367, opened 2015-05-23
Remove xml index plugins

Thynix:
The XML index plugins have long been obsolete and deprecated, are no longer maintained, and the time has come to remove them. (Rather than spend time updating them for purge-db4o.)

While this implementation will work, perhaps a better way to do it would be avoid the multiple maintenance points of `NoLongerOfficialPluginNames`, introduce a `.removed()`, and change the XMLLibrarian and XMLSpider `.deprecated()` to it.


Thynix:
We also need to remove the localization for them.

Thynix:
Introducing a `.unsupported()` in the official plugin description works much more nicely.

Thynix:
This is superseded by #368.

PR #368, opened 2015-05-23
Mark XMLLibrarian and XMLSpider as unsupported

Thynix:
These plugins have long been obsolete and deprecated and are no longer
maintained. This also removes the localization because these are not
just unsupported - they are superseded - and it's difficult to see
someone picking them up for official maintenance again.


src/freenet/pluginmanager/OfficialPlugins.java
```diff
@@ -262,6 +274,7 @@ private OfficialPluginDescription createOfficialPluginDescription() {
 			this.deprecated = deprecated;
 			this.experimental = experimental;
 			this.advanced = advanced;
+      this.unsupported = unsupported;
```
Bombe: Indentation seems off here.

Bombe:
I support this whole-heartedly. Get rid of the cruft!

ArneBab:
looks good!

src/freenet/pluginmanager/OfficialPlugins.java
Thynix: It sort of is - all added lines in this follow the [style guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.2-block-indentation). Does it make sense to follow the guide regardless of surrounding whitespace for new lines and automatically reindent the codebase once all pull requests are either merged or closed?

Thynix: Now that I've typed it I'm not so sure that makes sense - why be out of line with existing code if we're going to autoconvert? That said there's so much code that has yet to be merged that it may never be reasonable to autoconvert.

PR #371, opened 2015-05-30
Fix: Transfer backoff was never being reset.

TheSeeker:
This bug has been present since introduction of different backoff types in 2008.
See commit: 514b85cb197777345c451342b33dd2f85b20747f


Thynix:
Nice find!

PR #373, opened 2015-06-11
This should resolve bug 0006510 (CSS3 Issues)

drcrane:
I notice more properties related to box-sizing (more detail here: http://www.w3.org/TR/css3-ui/) should be added to CSSTokenizerFilter, I am happy to add these if my work so-far is satisfactory.

I notice from the flexbox example page: http://www.w3.org/TR/css3-flexbox/ is a curious value for the "min-width" attribute: "min-width: min-content;". I cannot seem to find where in W3C documentation that this is defined but a quick web-search shows it is in use. Should this and some others be added as valid values to appropriate attributes? I will compile a list to add to the bug report.

On units for isLength() in FilterUtils: I thought about keeping the existing logic for 2 character strings and then only performing the back-tracking if no match but unfortunately "rem" ends in "em" and "vmin" ends in "in" and so there was a clash, not the end of the road but perhaps over-optimisation?


Thynix:
Thanks for posting this! I'm encouraged by the addition of tests! If someone else wants to review that's fine by me, but if not I hope to have time on Sunday.

In the meantime, the [style guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace) suggests whitespace before and after operators (like `==`) and between `if` and `(`. Could I trouble you to bring the other lines this diff modifies up to date as well?

drcrane:
Sorry, I was trying to keep with the existing style... certainly, would you also like me to make the if statements (the ones effected) conform to [K&R Style](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.1.2-blocks-k-r-style) as per the style guide?
Use this format:
```java
if ("blah".equals(str)) {
    System.out.println("str is blah");
}
```
rather than:
```java
if ("blah".equals(str))
{
    System.out.println("str is blah");
}
```
in CSSTokenizerFilter.java. Many other files in the project seem to follow K&R style.

Thynix:
Sure - the [braces section](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.1-braces) of the guidelines specifies K&R, as well.

drcrane:
That should do it, I look forward to any feedback.

Thynix:
Sorry, change of plans: while others are very welcome to review this in the interest of avoiding scope creep for 1468 I will hold off on reviewing / merging this until 1468 is released.

Thynix:
1468 will be released 2015-07-11. Once that happens unless critical bugs that escaped testing necessitate another immediate bugfix release (like 1467 after 1466) I can review this.

Thynix:
1468 was released on-time, but I'm moving to a new apartment this week so review time has been scarce. If someone like @nextgens would like to review this it'd be very much appreciated, otherwise I expect to be able to get to this on Sunday at the earliest.

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -657,6 +664,10 @@
 		propertyTests.put("#xy34 { border-style: solid dotted }", "#xy34 { border-style: solid dotted }");
 		propertyTests.put("h1 { border-bottom: thick solid red }", "h1 { border-bottom: thick solid red }");
 		propertyTests.put("h1[foo] { border: solid red; }", "h1[foo] { border: solid red; }");
+		propertyTests.put("div { box-sizing: content-box; }", "div { box-sizing: content-box; }");
+		propertyTests.put("div { box-sizing: border-box; }", "div { box-sizing: border-box; }");
+		propertyTests.put("div { box-sizing: padding-box; }", "div { box-sizing: padding-box; }");
+		propertyTests.put("div { box-sizing: inhemerit; }", "div { }");
```
ArneBab: is this a typo? 

    propertyTests.put("div { box-sizing: inhemerit; }", "div { }");

drcrane:
It is, but it is intentional to test for a value that is invalid.
The W3C standard defines `content-box` and `border-box` but Firefox also supports `padding-box` and `inherit`. I know I have accepted `padding-box` as valid so I should probably include `inherit` too?

http://www.w3.org/TR/css3-ui/#propdef-box-sizing
https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

Thynix:
I plan to review this Monday evening. Sorry for the delay!

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -371,8 +385,16 @@ public boolean isValidURI(String URI)
 	 */
 	private static void addVerifier(String element)
 	{
-		if("azimuth".equalsIgnoreCase(element))
-		{
+		if ("align-content".equalsIgnoreCase(element)) {
+			elementVerifiers.put(element, new CSSPropertyVerifier(Arrays.asList("flex-start", "flex-end", "center", "space-between", "space-around", "stretch"), ElementInfo.VISUALMEDIA));
```
Thynix: Does this prevent specifying `auto` for `align-content`?

Thynix: Why are [`baseline` and `last-baseline`](http://www.w3.org/TR/css3-align/#content-distribution) omitted?

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -371,8 +385,16 @@ public boolean isValidURI(String URI)
...
+			allelementVerifiers.remove(element);
+		} else if("align-items".equalsIgnoreCase(element)) {
+			elementVerifiers.put(element, new CSSPropertyVerifier(Arrays.asList("flex-start", "flex-end", "center", "baseline", "stretch"), ElementInfo.VISUALMEDIA));
```
Thynix: Why are [`auto`, `last-baseline`, positions other than `flex-start`, `flex-end`, and `center`,](http://www.w3.org/TR/css3-align/#typedef-baseline-position) and the [`overflow-position`](http://www.w3.org/TR/css3-align/#typedef-overflow-position)s omitted?

Thynix:
I'm beginning to suspect both that I'm missing something fundamental and that I would continue to find similar things in the rest, so I will wait for feedback at this point.

ArneBab:
@drcrane could you add a comment above the test for filtering out invalid values stating that that’s its intention? Or maybe rename the value from inhemerit (which looks as if it might be a strange but real identifier I just did not know) to something like invalidValueToTestTheFilter?

drcrane:
On `align-content` I was quite confused how I missed this because I used the document you referenced to implement the code. The addition of `auto` and `space-evenly` etc must be recent changes these are not present in [Mozila's documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/align-content) and also these values fail the [CSS Validator](http://jigsaw.w3.org/css-validator/validator?uri=http%3A%2F%2Fbengreen.eu%2Fcsstesting%2Findex.css&profile=css3&usermedium=all&warning=1&vextwarning=&lang=en) (see the end of the CSS source file [index.css](http://www.bengreen.eu/csstesting/index.css)). I was using an older version of the document it would seem.

There is also something that I missed, three special values accepted by all CSS attributes: `initial`, `inherit` and `unset` aka [W3C Defaulting Keywords](http://www.w3.org/TR/css3-cascade/#defaulting-keywords). I am not sure when these were introduced but as they are valid for all attributes it is likely inclusion will involve some pain.

I will review each tag in turn (including those not altered by me) and compare them to the W3C standard.

drcrane:
@ArneBab: I like the latter suggestion of using `invalidValueToTestTheFilter` I will be sure to do that.

drcrane:
I did not think this work would be quite so involved but I think this will work. One area of concern I do have is the definition of the `nav-*` attributes: `nav-left: div#someid 'http://example.org'` is valid and will pass through the filter. The second argument is for <target-name> [CSS Basic UI Level 3](http://www.w3.org/TR/css3-ui/#propdef-nav-up) which is a string and should not be used as a URL but could it be possible to compose string that a buggy browser would navigate to??

In summary, sorry for my first attempt, I hope this pull request will now do what I intended, **support for: CSS3 Display, UI and Flexbox attributes and values**.

Thynix:
No worries - thank you for doing this! If a browser uses a string which "indicates the target frame" as a URL to navigate to that's a bug in the browser and not our problem unless you see indications that mainstream browsers - Chrome, Chromium, Internet Explorer, Edge, Safari, Opera, Firefox - have such bugs.

drcrane:
I see no indication of a bug like that but the last time I looked at the source code for Firefox was 4 years ago!

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -315,11 +339,11 @@ public boolean isValidURI(String URI)
 		//<border-color>
 		auxilaryVerifiers[11]=new CSSPropertyVerifier(Arrays.asList("transparent"),Arrays.asList("co"),null,null,true);
 		//<border-style>
-		auxilaryVerifiers[13]=new CSSPropertyVerifier(Arrays.asList("none","hidden","dotted","dashed","solid","double","groove","ridge","inset","outset", "inherit"),Arrays.asList("le"),null,null,true);
+		auxilaryVerifiers[13]=new CSSPropertyVerifier(Arrays.asList("none","hidden","dotted","dashed","solid","double","groove","ridge","inset","outset"),Arrays.asList("le"),null,null,true);
 		//<border-width>
 		auxilaryVerifiers[14]=new CSSPropertyVerifier(Arrays.asList("thin","medium","thick"),Arrays.asList("le"),null,null,true);
 		//<border-top-color>
-		auxilaryVerifiers[15]=new CSSPropertyVerifier(Arrays.asList("transparent","inherit"),Arrays.asList("co"),null,null,true);
+		auxilaryVerifiers[15]=new CSSPropertyVerifier(Arrays.asList("transparent"),Arrays.asList("co"),null,null,true);
```
Thynix: Commenting here because this wasn't immediately obvious to me: in addition to adding support for CSS 3 things this also [separates support for defaulting keywords](https://github.com/freenet/fred/pull/373/files#diff-c78a1250177dae55aff5ca13c8e2a271R3883) from the explicitly allowed words for each tag.

Thynix:
Does anyone know enough to review this pull request? It adds useful things!

ArneBab:
review notes:

* in `test/freenet/client/filter/FilterUtilsTest.java` the line endings are \r\n instead of \n. I fixed that in https://github.com/ArneBab/fred-staging-1/commit/bde9bbe0ee997dff27d7901bdfdb74eeae4df0d3
* is 1. really not an allowed value in CSS? `test/freenet/client/filter/FilterUtilsTest.java`

```
+		assertFalse(FilterUtils.isLength("1.em", false));
+		assertFalse(FilterUtils.isLength("1.", false));
```
* ruby-base-container: wow, I had forgotten that that exists and thought about the programming language, not the typography system. http://www.w3.org/TR/2001/WD-css3-ruby-20010216/
* we’re actually redefining array elements by index?? wow… Kudos to you for taking up this task and navigating the code! 
* there is inconsistent use of spaces and tabs for whitespace here:

```
		allelementVerifiers.add("clip");
               allelementVerifiers.add("break-before");
```

The only real issue I hit upon:

* I’m a bit concerned about the selector and string value ("se" and "st"). Doesn’t that allow passing through anything?

ArneBab:
It looks like the code needs Java 7.

```
   [javac] fred/src/freenet/client/filter/CSSTokenizerFilter.java:4227: error: diamond operator is not supported in -source 1.6
    [javac]                     List<CSSPropertyVerifier> propertyVerifierList = new ArrayList<>();
```

@Thynix How far are we at requiring Java 7?

drcrane:
* You are correct, `1.em` and `1.` seem to be valid according to the W3C validator. I will see what amendments I can make.
* Thanks for making the changes to the line endings, I will make sure my git configuration is correct.
* Though I did notice the spaces and tabs inconsistency I took the same stance as with the K&R Style braces and only altered indentation if my code changes required me to do so.
* I also had the same concern of letting through any old string (see my earlier comment about `nav-*` attributes) what would you suggest?

drcrane:
@ArneBab I will remove the diamond operator (it only saves a few keystrokes anyway) sorry about that.

Thynix:
Java 7 can be required in 1472 with 1471 as a transition build.

ArneBab:
With Java7 the tests pass.

ArneBab:
@drcrane I would rather be too strict than too accommodating ⇒ I’d rather break some layouts. Are there other features than nav which are broken when filtering out string and selector values? 

string ("st") seems to be used in "none", "current", "root", "ellipsis" and "clip". At least for the latter two it might be possible to directly set a list of allowed values, while for current and root the url filter might work.

Full CSS is a Turing complete language (very nice example: http://silon.slaks.net/#chained-adder ), so we cannot afford exposing all its capabilities. An alternative could be to run the items through the url-filter — that could allow some selectors to survive.

ArneBab:
So text-overflow it accepts any string value: 

    string 	Render the given string to represent clipped text 

→ http://www.w3schools.com/cssref/css3_pr_text-overflow.asp

I think we should treat it the same as :before and :after.

I don’t think I understand the CSS filter well enough to really judge what it allows and what not…

Thynix:
Are we expecting people to be targeted based on browser window size / screen resolution? Is that something to guard against? I wonder if it costs too much prettiness power for that to be a good tradeoff.

drcrane:
@Thynix we could simply enforce convention and perhaps dictate that only strings matching:
`[A-Za-z_ 0-9]*` (that is composed entirely of space, underscore and alpha numeric characters) should be accepted? Even filter out characters that do not match that pattern. The strings `_parent` and `_root` are invalid anyway but remain in the spec for backward compatibility.

I have made the changes @ArneBab suggested and have been looking at incorporating the URL filter for checking the strings. I will commit the minor changes tonight.

Thynix:
That commit LGTM. Anyone have additional things, or is this ready to merge?

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -365,22 +389,57 @@ public boolean isValidURI(String URI)
...
+		
+		// <justify-items>
+		// auto | stretch | <baseline-position> | [ <item-position> && <overflow-position>? ] | [ legacy && [ left | right | center ] ]
+		auxilaryVerifiers[128] = new CSSPropertyVerifier(Arrays.asList("legacy"), null, null, null, true);
+		auxilaryVerifiers[129] = new CSSPropertyVerifier(null,ElementInfo.VISUALMEDIA,null,Arrays.asList("4 128", "128 4"), true,true);
+		auxilaryVerifiers[130] = new CSSPropertyVerifier(null,ElementInfo.VISUALMEDIA,null,Arrays.asList("129 126", "126", "126 129"), true,true);
+		
+		// used in nav-down, nav-left, nav-right and nav-up
+		auxilaryVerifiers[143] = new CSSPropertyVerifier(null, Arrays.asList("se"), null, null, true);
+		auxilaryVerifiers[144] = new CSSPropertyVerifier(Arrays.asList("current", "root"), Arrays.asList("st"), null, null, true);
```
ArneBab: is the use of "st" (unchecked string) safe here?

ArneBab: OK, I think fixing this is not your job. It has to be fixed here, but your change does not cause a regression. https://github.com/drcrane/fred/blob/next/src/freenet/client/filter/ElementInfo.java#L561

So from my side your change good to go. Thank you for persisting!

Thynix:
Thanks!

ArneBab:
Thank you!

PR #374, opened 2015-06-12
RequestClient builder

Bombe:
Adds a `RequestClient` builder that is easy to use and cuts down on the number of anonymous classes created.


Thynix:
Would it be reasonable to have the builder require an enum argument of realtime / bulk? Currently it's implicitly bulk and it seems like enums might be clearer.

Bombe:
In Fred a greater number of non-real-time than real-time request clients is used so I think it’s a sensible default.

Thynix:
Ah, yeah. Okay, that makes sense. Thanks!

PR #375, opened 2015-06-12
Try to reload seednodes if they are missing

Bombe:
When starting a new node without seednodes, Fred gives a warning about the missing seednodes in the web interface but never tries again to find them, so that even if you download the seednodes into the node directory you have to restart the node in order for it to find the seednodes.


Thynix:
Is a missing seed nodes file a thing that happens somehow? The installers ship with a copy.

Bombe:
It happens every time I start a node from a development environment. We don’t have a seednodes file in the repository (which is okay) and it’s simply a nuisance having to restart the node manually after having downloaded the seednodes file.

Thynix:
Thanks! I documented the reasoning we talked about on IRC as well: a790cc26d6b718373df9e50da549f80a81df4495

PR #381, opened 2015-07-11
Prioritise memory jobs

toad:
Priorities for MemoryLimitedJob's, i.e. FEC decode/encode jobs. Please check the call stack to ensure that there are no locking issues, but there shouldn't be any.

This should ensure that fproxy is reasonably responsive even when there are big files completing, and mitigate the damage caused by BC < 1.52 with encrypted disk enabled.


toad:
I have re-reviewed this and IMHO there are no hidden concurrency issues. This should be safe. It seems to work well in my local testing too.

toad:
It is however less important to get this out quickly provided we can get #382 out soon.

toad:
A full solution to "stalls for hours at 100% CPU and misses the update" does require both patches I guess...

Thynix:
Are people happy with this?

PR #383, opened 2015-07-12
Bug6591: improve compression algorithm choice

nextgens:
The compression algorithm choice is now dependant on number of blocks rather than compression ratio from byte size


xor-freenet:
```<nextgens> [12:24:42] not tested```

PR #385, opened 2015-07-15
Update bouncycastle location

RdrOfTheSt0rm:
Opening the link provided by ant when bouncycastle cannot be found does not work. I have modified it to what i believe to be more robust locations


RdrOfTheSt0rm:
also removed /latest/ from build.xml as this is no longer maintained
Please review

nextgens:
Hmm, this is the wrong fix; what went wrong is our release scripts

PR #386, opened 2015-07-16
Allow icons in <link> tags

Bombe:
This change allows freesites to set a favicon which is shown in the browser tab.

### DIFF
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -2518,6 +2518,7 @@ public class HTMLFilter implements ContentDataFilter, CharsetExtractor {
 			
 			String parsedRel = "", parsedRev = "";
 			boolean isStylesheet = false;
+			boolean isIcon = false;
 
 			if(rel != null) {
 				
@@ -2537,6 +2538,8 @@ public class HTMLFilter implements ContentDataFilter, CharsetExtractor {
 							if(tok.hasMoreTokens())
 								return null; // Disallow extra tokens after "stylesheet"
 						}
+					} else if (token.equalsIgnoreCase("icon")) {
+						isIcon = true;
 					} else if(!isStandardLinkType(token)) continue;
 					
 					i++;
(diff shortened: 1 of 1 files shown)


src/freenet/client/filter/HTMLFilter.java
```diff
@@ -2616,7 +2619,11 @@ void processStyle(HTMLParseContext pc) {
 			String href = getHashString(h, "href");
 			if (href != null) {
 				href = HTMLDecoder.decode(href);
-				href = htmlSanitizeURI(href, type, charset, maybecharset, pc.cb, pc, false);
+				if (isIcon) {
+					href = htmlSanitizeURI(href, type, null, null, pc.cb, pc, false);
+				} else {
+					href = htmlSanitizeURI(href, type, charset, maybecharset, pc.cb, pc, false);
+				}
```
ArneBab: Why don’t you set the charset and maybecharset to null if (isIcon) and then call htmlSanitizeURI(href, type, charset, maybecharset, pc.cb, pc, false)?

Bombe: a) it didn’t occur to me to clobber variables. :)
b) now that I thought about it I still think it would be inferior to what I actually did.

If I would have written it the way you suggest it would be easy for the two resulting pieces of code (the if-block with the null-setting inside and the call to sanitizeURI) to be moved away from each other which would make the intent of not having a charset on an image — which is what this is about! — a lot less clearer. The way I did it it’s always clear that something different happens for an icon.

ArneBab: I think you’re right. Changing the charset and maybecharset could result in problems if the code structure changes (for example is fomething comes after the current code).

Are there other content types which need a change in the charset?

Bombe: http://www.w3.org/TR/2014/REC-html5-20141028/links.html#linkTypes says there are only three values for “rel” which allow “href” to point to external resources. “stylesheet” is already handled by the code, support “link” I added with the PR, and “prefetch” is already removed by the logic involving `standardRelTypes` (see [HTMLFilter.java:2644](https://github.com/Bombe/fred/blob/1caf4622211ac5349a4ef4786df4265c4e3fb856/src/freenet/client/filter/HTMLFilter.java#L2644)).

PR #387, opened 2015-07-16
Add request identifiers to persistent requests

Bombe:
I’m currently writing a new high-level FCP client implementation and I’m facing a problem with processing the various replies to a `ListPersistentRequests`. Usually, when I send an Identifier with a request, the response carries the same Identifier; associating one with the other is a breeze. However, the replies to a `ListPersistentRequests` all have their own identifiers so it is impossible to associate the replies with a request. Using the approach of “use everything until you get an `EndListPersistentRequests`” will only work if all replies to a `ListPersistentRequests` are sent atomically. Unless that is guaranteed, a second `ListPersistentRequests` on the same connection will happily interleave its replies with the replies to the first request. Madness will ensue.

This pull request contains changes to all involved classes that will cause all messages sent as reply to a `ListPersistentRequests` to contains a `ListRequestIdentifier` field which works similar to the usual `Identifier` field; only that it does not contain the identifier of the request itself but the identifier of the request that caused the sending of the messages. The `EndListPersistentRequests` message contains the usual `Identifier` field, though.

### DIFF
--- a/src/freenet/clients/fcp/ClientGet.java
+++ b/src/freenet/clients/fcp/ClientGet.java
@@ -371,8 +371,8 @@ public class ClientGet extends ClientRequest implements ClientGetCallback, Clien
 			if(returnType == ReturnType.DIRECT)
 			    returnBucketDirect = data;
 		}
-		trySendDataFoundOrGetFailed(null);
-		trySendAllDataMessage(null);
+		trySendDataFoundOrGetFailed(null, null);
+		trySendAllDataMessage(null, null);
 		finish();
 		if(client != null)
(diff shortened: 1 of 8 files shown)


toad:
Does this make Identifier mandatory on ListPersistentRequests?

Bombe:
Hmm, it does. It could be made optional; it is already kind-of optional for messages that are sent due to other circumstances (such as a `WatchGlobal` with `Enabled=true`). I'll change that.

Thynix:
Is this ready to be merged?

toad:
Looks fine to me.

PR #388, opened 2015-08-02
Load management send slow down when close to limit (without strict fairness)

toad:
This makes fred send a slow-down signal when load is above 75% of the node's capacity, rather than only sending slow-down signals when we reject requests. The idea is to reduce the number of rejections (which can cause misrouting). The details are complicated and need reviewing.

Note that this is different to https://github.com/freenet/fred/pull/313 because it does NOT include the strict fairness code (and so doesn't need the associated changes to messages). IMHO that is probably not a good idea. The slow-down criteria are different too; it's a bit of a messy kludge, but strict fairness would probably cost us too much (at least if traffic is bursty).

### DIFF
--- a/src/freenet/node/CHKInsertHandler.java
+++ b/src/freenet/node/CHKInsertHandler.java
@@ -119,14 +119,6 @@ public class CHKInsertHandler implements PrioRunnable, ByteCounter {
 			return;
 		}
 		
-		if(tag.shouldSlowDown()) {
-			try {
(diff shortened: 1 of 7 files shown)


ArneBab:
why are the slow down criteria different — and how do they differ? (these were already reviewed — what do the reviewers have to look out for in addition to what they already read?)

ArneBab:
@toad I’d like to get this merged, but I need your answer for that.

PR #389, opened 2015-08-02
Lower backoff constants harvested from 'performance' patches.

TheSeeker:
Only contains lowered initial and maximum backoff values. The original patches
also included early-zeroing of the next-delay on success, and vastly reduced
delays for sending requests.  See branch: let-it-burn

### DIFF
--- a/src/freenet/node/PeerNode.java
+++ b/src/freenet/node/PeerNode.java
@@ -2951,21 +2951,21 @@ public abstract class PeerNode implements USKRetrieverCallback, BasePeerNode, Pe
 	long routingBackedOffUntilRT = -1;
 	long routingBackedOffUntilBulk = -1;
 	/** Initial nominal routing backoff length */
-	static final int INITIAL_ROUTING_BACKOFF_LENGTH = (int) SECONDS.toMillis(1);
+	static final int INITIAL_ROUTING_BACKOFF_LENGTH = (int) MILLISECONDS.toMillis(200); // mm and oo: 200ms, toad: 1s
 	/** How much to multiply by during fast routing backoff */
 
 	static final int BACKOFF_MULTIPLIER = 2;
 	/** Maximum upper limit to routing backoff slow or fast */
-	static final int MAX_ROUTING_BACKOFF_LENGTH = (int) HOURS.toMillis(3);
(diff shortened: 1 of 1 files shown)


Thynix:
These values look reasonable to me. Is your intent to keep the attribution comments around? What's the reasoning for that?

TheSeeker:
I just copied over the changed lines without editing, I'm not attached to any of the content of the patch.  When applying it, if you so choose, you might set the author to mmaalfdks@Da216sBMWMoFkqTiJvKZEg15Ero though, I'm not sure how to do that locally.

src/freenet/node/PeerNode.java
```diff
@@ -2951,21 +2951,21 @@ public boolean isRoutingBackedOffEither() {
 	long routingBackedOffUntilRT = -1;
 	long routingBackedOffUntilBulk = -1;
 	/** Initial nominal routing backoff length */
-	static final int INITIAL_ROUTING_BACKOFF_LENGTH = (int) SECONDS.toMillis(1);
+	static final int INITIAL_ROUTING_BACKOFF_LENGTH = (int) MILLISECONDS.toMillis(200); // mm and oo: 200ms, toad: 1s
```
nextgens: This is a terrible idea... 200ms is less than the expected RTT... meaning that it ignores the first few backoff rounds

Thynix:
It seems like taking just the lowered `MAX_ROUTING_BACKOFF_LENGTH` and `MAX_TRANSFER_BACKOFF_LENGTH` could be worthwhile? It might normalize and end up with no overall change, but I'd suspect it'd avoid long bulk queue hangs.

ArneBab:
I think we can use the average lowest remaining connection time of peers as max backoff-time (that’s when backoff conditions will change).

Looking at my current opennet stats and somewhat carelessly assuming that the connected time is a good proxy for the remaining time, 8 minutes doesn’t sound too bad. 

From our statistics, there are many 24/7 nodes and many nodes with uptime below 10%. If we assume that this is daily uptime, we have 2.4h uptime. Half of that is what would be remaining on average, so a reasonable max backoff time with 10 peers would be 1.2h, maybe going down to 30min. Since only the lowest counts, this is close enough to 8 min to make that useful.

@TheSeeker can you do the change? (only use the max times from the patches)

PR #390, opened 2015-08-08
MessageAuthCode: fix JVM-version-dependent test failure

Thynix:
This fixes the test failures discussed [on devl](https://emu.freenetproject.org/pipermail/devl/2015-August/038145.html) by preventing `MessageAuthCode.verify()` from throwing when given `null`. Is this an appropriate way to go about it? I didn't find any callers where this would change behaviour.


Bombe:
Looks good.

PR #394, opened 2015-08-11
Async chkinsertsender riskier cleanup

toad:
This includes some slightly more ambitious cleanups, hopefully mostly affecting logging. The async-chkinserthandler branch on the other hand is a bit weaker.


toad:
This seems to work on my node and also in synthetic tests (RealNodeRequestInsertTest and RealNodeULPRTest). Please consider for merging into a test release. It should greatly reduce the number of threads needed on high bandwidth nodes.

toad:
Unfortunately this might be affected by issues in #400. One possibility is to leave out the last commit but I'm concerned about memory usage.

ArneBab:
@toad Bombe said that #400 is OK now, is there something else blocking this?

toad:
Not IIRC, although it should be tested carefully, especially in vitro.

bertm:
Tested for days without problems *in vivo*. Number of running threads was reduced drastically by this changeset — well done.

I did not review the code except for quickly reading through, nor did I test anything *in vitro* except for unit tests.

toad:
Merged next into the branch. Please consider for 1472.

Thynix:
Whee, more conflicts.

PR #397, opened 2015-08-12
Recreate message, as required

toad:
Should be trivial log fix for swapping. But needs testing on actual darknet nodes, or at least on a hybrid node (such as mine).

### DIFF
--- a/src/freenet/node/LocationManager.java
+++ b/src/freenet/node/LocationManager.java
@@ -1090,10 +1090,11 @@ public class LocationManager implements ByteCounter {
         }
         item.lastMessageTime = System.currentTimeMillis();
         // Returning to source - use incomingID
-        m.set(DMT.UID, item.incomingID);
+        byte[] hisHash = ((ShortBuffer)m.getObject(DMT.HASH)).getData();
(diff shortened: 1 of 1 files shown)


Thynix:
I don't understand why this is doing what it does.

toad:
DMT/inter-node messages can have sub-messages, which can take up space and propagate information that ought not to be propagated (maybe even used for tracing across the network by a passive attacker). Recreating the message explicitly is therefore preferable to forwarding the original message. There is an existing warning message in the MessageCore code to check for this.

bertm:
This looks like duplication of logic. Why not use `Message.cloneAndDropSubMessages()`?

Thynix:
@toad Ping.

toad:
Because the ID changes?

Thynix:
The incomingID? I don't follow. Why does it change?

toad:
The ID on the message changes, because we change it from incoming to outgoing, or vice versa?

bertm:
Ah, of course.

In retrospect, this is more reasonable than I originally thought, as it is consistent with how swap replies are generated in `run()`, however relaying the swap reply (for a forwarded swap request) instead of generating it.

PR #398, opened 2015-08-14
Fix keepalive

toad:
Fixes keepalives in NewPacketFormat (regression with NewPacketFormat, years ago). This makes multi-node-one-VM simulations much faster and more reliable, but also should improve reliability of low traffic connections generally.

### DIFF
--- a/src/freenet/node/NewPacketFormat.java
+++ b/src/freenet/node/NewPacketFormat.java
@@ -992,7 +992,7 @@ addOldLoop:			for(int i = 0; i < startedByPrio.size(); i++) {
 	 * queued at plus the lesser of half the RTT or 100ms if there are acks queued. 
 	 * Otherwise Long.MAX_VALUE to indicate that we need to get messages from the queue. */
 	@Override
-	public long timeNextUrgent(boolean canSend) {
+	public long timeNextUrgent(boolean canSend, long now) {
 		long ret = Long.MAX_VALUE;
 		if(canSend) {
 			// Is there anything in flight?
@@ -1014,9 +1014,9 @@ addOldLoop:			for(int i = 0; i < startedByPrio.size(); i++) {
 		}
 		// Check for acks.
 		ret = Math.min(ret, timeCheckForAcks());
-		
+				
 		// Always wake up after half an RTT, check whether stuff is lost or needs ack'ing.
-		ret = Math.min(ret, System.currentTimeMillis() + Math.min(100, (long)averageRTT()/2));
+		ret = Math.min(ret, now + Math.min(100, (long)averageRTT()/2));
 		
 		if(canSend && DO_KEEPALIVES) {
 		    synchronized(this) {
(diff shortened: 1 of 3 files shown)


toad:
Use of synchronized(this) in timeNextUrgent should be safe as it's called more or less directly from PacketSender, and other methods use the same lock.

Thynix:
I don't understand why this does what it does. Why is the lock not held for checking canSend too?

toad:
canSend is a parameter/local variable here AFAICS?

toad:
The real question is why do we only update the time if canSend is true - the answer is keepalives need us to allocate a message number, which is what canSend tells us about.

Thynix:
What does this fix?

toad:
Keepalive packets were not being sent (they are supposed to be sent every 7 seconds, if there is no other traffic). Low traffic connections would therefore be lost (after 35 seconds). This caused problems for simulations.

src/freenet/node/NewPacketFormat.java
```diff
@@ -1015,8 +1015,17 @@ public long timeNextUrgent(boolean canSend) {
 		// Check for acks.
 		ret = Math.min(ret, timeCheckForAcks());
 		
-		// Always wake up after half an RTT, check whether stuff is lost or needs ack'ing.
-		ret = Math.min(ret, System.currentTimeMillis() + Math.min(100, (long)averageRTT()/2));
+		if(ret > now) {
+		    // Always wake up after half an RTT, check whether stuff is lost or needs ack'ing.
+		    ret = Math.min(ret, now + Math.min(100, (long)averageRTT()/2));
+		    
+		    if(canSend && DO_KEEPALIVES) {
```
Thynix: What does this block do? Can you add a comment at least?

toad: The if() means if we're going to wake up immediately we don't need to consider keepalives. If there is room to send in the message window, we check for whether we need to send a keepalive, i.e. using the time since we last sent a packet.

src/freenet/node/NewPacketFormat.java
```diff
@@ -992,7 +992,7 @@ public void checkForLostPackets() {
 	 * queued at plus the lesser of half the RTT or 100ms if there are acks queued. 
 	 * Otherwise Long.MAX_VALUE to indicate that we need to get messages from the queue. */
 	@Override
-	public long timeNextUrgent(boolean canSend) {
+	public long timeNextUrgent(boolean canSend, long now) {
 		long ret = Long.MAX_VALUE;
```
ArneBab: Do I understand it correctly that this changes the public API? Is it ensured that nothing uses this?

toad: This is true. However I can't imagine a plugin using NewPacketFormat. If it did it would mean it was doing *really* low level stuff e.g. transport plugins - and there isn't really usable infrastructure for that. Maybe forcing a packet send, maybe some ambitious low-level diagnostics...?

toad: It might make sense to look at whether we could make this package-level.

ArneBab: In short: We don’t know whether anyone uses this. But transport plugins just might — which are already bitrotting, so this could make the bitrot worse.

Is it necessary to do this by changing the function?

Thynix:
UOM tests continue to fail with this merged.

toad:
Nextgens ran into this recently. Is it merged?

ArneBab:
@toad now it is merged.

@Thynix Thank you!

PR #399, opened 2015-08-14
Fix some simulations

toad:


src/freenet/node/simulator/RealNodeRequestInsertTest.java
```diff
@@ -172,8 +176,8 @@ int insertRequestTest() throws CHKEncodeException, InvalidCompressionCodecExcept
         Node randomNode = nodes[node1];
         //Logger.error(RealNodeRequestInsertTest.class,"Inserting: \""+dataString+"\" to "+node1);
         
-        //boolean isSSK = requestNumber % 2 == 1;
-        boolean isSSK = true;
+        boolean isSSK = requestNumber % 2 == 1;
+        //boolean isSSK = true;
```
Thynix: Why keep both of these? The one switched to seems like a more general test - is that always preferable?

toad: The more general test is preferable here IMHO. But it's also useful to be able to toggle it, and this is testing code anyway, so having the commented out version is acceptable?

ArneBab: @toad could you add a note to the commented out code when it’s useful in testing? Aside from that I think it’s OK to keep the entry point for quick testing. All cleaner ways I see add multiple lines of overhead.

bertm: This comes to mind:
```java
static final boolean SSK_ODD_ONLY = true;
...
boolean isSSK = SSK_ODD_ONLY ? (requestNumber % 2 == 1) : true;
```

ArneBab: That looks great! I forgot about the ? operator — and this finally gives me a usecase where it provides better readability than the alternatives :)

Thank you!

PR #400, opened 2015-08-15
Add test for remove()

toad:
Clean up PrioritisedTicker a bit and add a test.

### DIFF
--- a/src/freenet/support/PrioritizedTicker.java
+++ b/src/freenet/support/PrioritizedTicker.java
@@ -194,10 +194,10 @@ public class PrioritizedTicker implements Ticker, Runnable {
 	}
 	
 	/** Queue a job at a specific absolute time. 
-	 * @param l The absolute time at which the job should run.
+	 * @param runJobAt The absolute time at which the job should run.
 	 * @param offset The offset in milliseconds from "now" (i.e. some recent call to 
 	 * System.currentTimeMillis()). */
-    private void queueTimedJobInner(Runnable runner, String name, long l, long offset, 
+    private void queueTimedJobInner(Runnable runner, String name, long runJobAt, long offset, 
             boolean runOnTickerAnyway, boolean noDupes) {
         if(noDupes) runOnTickerAnyway = true;
         if(offset <= 0 && !runOnTickerAnyway) {
@@ -208,29 +208,29 @@ public class PrioritizedTicker implements Ticker, Runnable {
         Job job = new Job(name, runner);
         synchronized(timedJobsByTime) {
             if(noDupes) {
-                Long t = timedJobsQueued.get(job);
-                if(t != null) {
-                    if(t <= l) {
+                Long alreadyQueuedAt = timedJobsQueued.get(job);
+                if(alreadyQueuedAt != null) {
+                    if(alreadyQueuedAt <= runJobAt) {
                         Logger.normal(this, "Not re-running as already queued: "+runner+" for "+name);
                         return;
                     } else {
                         // Delete the existing job because the new job will run first.
-                        removeQueuedJobInner(job, t);
+                        removeQueuedJobInner(job, alreadyQueuedAt);
                     }
                 }
             }
-            Object o = timedJobsByTime.get(l);
+            Object o = timedJobsByTime.get(runJobAt);
             if(o == null)
-                timedJobsByTime.put(l, job);
+                timedJobsByTime.put(runJobAt, job);
             else if(o instanceof Job)
-                timedJobsByTime.put(l, new Job[]{(Job) o, job});
+                timedJobsByTime.put(runJobAt, new Job[]{(Job) o, job});
             else if(o instanceof Job[]) {
                 Job[] r = (Job[]) o;
                 Job[] jobs = Arrays.copyOf(r, r.length+1);
                 jobs[jobs.length - 1] = job;
-                timedJobsByTime.put(l, jobs);
+                timedJobsByTime.put(runJobAt, jobs);
             }
-            timedJobsQueued.put(job, l);
+            timedJobsQueued.put(job, runJobAt);
         }
         if(offset < MAX_SLEEP_TIME) {
             wakeUp();
--- a/src/freenet/support/PrioritizedTicker.java
+++ b/src/freenet/support/PrioritizedTicker.java
@@ -186,7 +186,8 @@ public class PrioritizedTicker implements Ticker, Runnable {
 	 * passed already, then run the job ASAP. 
 	 * @param time The time at which to run the job. @see System.currentTimeMillis()
 	 */
-	void queueTimedJobAbsolute(Runnable runner, String name, long time, 
+	@Override
+	public void queueTimedJobAbsolute(Runnable runner, String name, long time, 
             boolean runOnTickerAnyway, boolean noDupes) {
 	    long now = System.currentTimeMillis();
 	    queueTimedJobInner(runner, name, time, time-now, runOnTickerAnyway, noDupes);
--- a/src/freenet/support/PrioritizedTicker.java
+++ b/src/freenet/support/PrioritizedTicker.java
@@ -250,13 +250,13 @@ public class PrioritizedTicker implements Ticker, Runnable {
 		return executor;
 	}
 
-	public int queuedJobs() {
+	int queuedJobs() {
 		synchronized(timedJobsByTime) {
 			return timedJobsQueued.size();
 		}
 	}
 
-    public int queuedJobsUniqueTimes() {
+    int queuedJobsUniqueTimes() {
         synchronized(timedJobsByTime) {
             return timedJobsByTime.size();
         }
--- a/src/freenet/support/PrioritizedTicker.java
+++ b/src/freenet/support/PrioritizedTicker.java
@@ -136,13 +136,8 @@ public class PrioritizedTicker implements Ticker, Runnable {
 			}
 
 		if(sleepTime > 0) {
-			// Update logging only when have time to do so
 			try {
-				if(logMINOR)
-					Logger.minor(this, "Sleeping for " + sleepTime);
-				synchronized(this) {
-					wait(sleepTime);
-				}
+			    sleep(sleepTime);
 			} catch(InterruptedException e) {
 				// Ignore, just wake up. Probably we got interrupt()ed
 				// because a new job came in.
@@ -150,8 +145,15 @@ public class PrioritizedTicker implements Ticker, Runnable {
 		}
 	}
 
+	protected void sleep(long sleepTime) throws InterruptedException {
+        if(logMINOR)
+            Logger.minor(this, "Sleeping for " + sleepTime);
+        synchronized(this) {
+            wait(sleepTime);
+        }
+    }
 
-	@Override
+    @Override
 	public void queueTimedJob(Runnable job, long offset) {
 		queueTimedJob(job, "Scheduled job: "+job, offset, false, false);
 	}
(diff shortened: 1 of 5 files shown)


test/freenet/support/PrioritizedTickerTest.java
```diff
@@ -55,6 +67,71 @@ public void testSimple() throws InterruptedException {
 		}
 	}
 
+    public void testRemove() throws InterruptedException {
```
Bombe: Don’t take this the wrong way but this test is horrible.

* It is unclear what is being tested.
* Multiple things seem to be tested at once. It’s better to have multiple tests that only test a single thing each.
* `Thread.sleep` is used. This will unnecessarily delay the test and make test execution slower for everyone else. It’s better to mock the part that requires the passage of time so that all required actions can run as fast as possible.
* You rely on an assert. It’s better to use special exceptions if you rely on the behaviour. If you only want to make sure that certain things happened, a verifying mock would be more appropriate.
* The test relies on certain conditions that it painstakingly tries to create itself. Again, a mock that is simply configured to match that condition would make this a lot clearer.

toad: Unfortunately most of these issues are hard to fix without changing the code being tested. I will look at it shortly, I don't think this test is doing what it's supposed to do at the moment - that is, I don't see why it is throwing an assertion error in these cases unless there is a leak...

toad: I think the timing issues will be hard to fix however without a lot of extra and probably costly abstraction. The two existing tests in this file both rely on Thread.sleep already.

toad:
This is a significant improvement on the existing code IMHO. It cleans up and documents the remove method, adds a test (admittedly potentially unreliable, but hard to avoid this in this class), and removes duplicate code.

Thynix:
@Bombe Is this ready to be merged?

Bombe:
@Thynix well, none of my points has really been addressed… in my opinion this is still a horrible test, for all the reasons I mentioned above. I’m not really sure about how to best enforce these things, though; we can not really refuse to merge people’s contributions as long as our tests are the way they currently are. :/

toad:
Well, I have addressed at least one of your points: It no longer relies on an assertion. Also I believe I have improved the quality of the code being tested.

IMHO refusing to accept a new test because it doesn't abstract out the clock from the component being tested is unreasonable. Hacking bytecode to allow test control over System.currentTimeMillis() is possible I guess, but probably pretty heavyweight, especially while we still haven't resolved the issues with Maven (i.e. with introducing nearly any additional third party dependencies). Granted in the further future it's worth seriously considering for high level tests and simulations (I will be doing a project which may include this), although that'll probably mostly be at the level of MessageCore.

Do you have any reason for abstracting the clock out from PrioritizedTicker other than to make it easier to test? In general refactoring for testing is acceptable, but only if it's not too invasive / too costly?

Thynix:
Is my understanding correct that this test relies on timing? In that case it's a race condition incarnate. (Not that these changes introduced it.) Nondeterministic tests are not acceptable.

I agree that it's not at all clear what this test is testing - it's doing many things all at once.

toad:
That's what the for loop is for. I agree it's not ideal, but it should work in practice unless the JVM is ridiculously slow. E.g. it won't work if single-stepped in a debugger.

toad:
Sorry, to clarify: It keeps on adding jobs until they were both added in the same millisecond. Then it can execute the test.

toad:
I guess another solution would be to change the Ticker interface to accept absolute rather than relative times...

toad:
I will rewrite this to avoid requiring a race condition.

toad:
Okay, this should address your remaining concerns. The new tests don't use sleep at all. Some of the old tests still do however.

toad:
Okay this is done. Please have a look! Thanks.

src/freenet/support/PrioritizedTicker.java
```diff
@@ -267,54 +250,81 @@ public Executor getExecutor() {
...
+	/** Remove a queued job from the internal structures other than timedJobsQueued. The
+	 * caller is responsible for that.
+	 * @param job The job to remove.
+	 * @param t The time at which is it scheduled.
+	 */
+	private void removeQueuedJobInner(Job job, Long t) {
+        Object o = timedJobsByTime.get(t);
+        if(o == null) {
+            Logger.error(this, "Job in timedJobsQueued but not in timedJobsByTime");
+            // Cleaned up so no reason to throw in this critical structure.
```
Thynix: Elseriseth [has opinions](SSK%40WPECGLtYbVi8Rl6Y7Vzl2Lvd2EUVW99v3yNV3IWROG8%2CMnwtfrASwEUVgXVCxNALLwCKqe6woGt3biDRVcBulkk%2CAQACAAE/fms|2015-08-17|Message-1) on this:

> GRRR!
> That's outright insane.
> Those asserts should be rather removed, than converted to
> Logger.error.
> If condition MUST NEVER EVER HAPPEN, it should be either ignored, or be
> HARD ERROR, not just one more random message. And released (non-debug,
> post-alpha) binary SHOULD NOT EVEN CHECK for those condition (by
> disabling asserts).

I agree that if the data structure has a bug we should have it fail loudly so we can fix it instead of adding another log message to the fire hose. Without benchmarks showing that asserts are a bottleneck I disagree that asserts should be disabled in release mode.

Bombe:
Well, old tests are old; while we should rewrite old tests to not suck anymore at some point in the future, not committing new bad tests is something we can and absolutely should do today already.

That being said, this does look a bit cleaner than it did before. I’m not quite sure that adding two jobs for the same time (`System.currentTimeMillis`) is not still subject to a race condition, though…

toad:
That's why I use the new API to add it at an absolute time. AFAICS the whole thing is fixed now, but maybe I've missed something?

src/freenet/support/PrioritizedTicker.java
```diff
@@ -175,82 +177,63 @@ public void queueTimedJob(Runnable job, long offset) {
...
+            boolean runOnTickerAnyway, boolean noDupes) {
+	    long now = System.currentTimeMillis();
+	    queueTimedJobInner(runner, name, time, time-now, runOnTickerAnyway, noDupes);
+	}
+	
+	/** Queue a job at a specific absolute time. 
+	 * @param l The absolute time at which the job should run.
+	 * @param offset The offset in milliseconds from "now" (i.e. some recent call to 
+	 * System.currentTimeMillis()). */
+    private void queueTimedJobInner(Runnable runner, String name, long l, long offset,
```
Bombe: Oh, by the way: `l`? Seriously? Are we paying github by the byte now? :)

Bombe:
Okay, the `time` parameter is now handed in from the outside but `offset` might still be different for two consecutive executions of `queueTimedJobAbsolute`… but as it can only be `0` (or smaller) and both cases are treated the same, I guess it’s okay now.

Thynix:
testDeduping fails:

    [junit] Running freenet.support.PrioritizedTickerTest
    [junit] Testsuite: freenet.support.PrioritizedTickerTest
    [junit] Tests run: 4, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.31 sec
    [junit] Tests run: 4, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.31 sec
    [junit] ------------- Standard Output ---------------
    [junit] Attempting to load the NativeThread library [jar:file:/home/steve/Documents/Coding/freenet/fred/lib/freenet/freenet-ext.jar!/freenet/support/io/libNativeThread-amd64.so]
    [junit] Using the NativeThread implementation (base nice level is 0)
    [junit] Starting Ticker
    [junit] Starting Ticker
    [junit] Starting Ticker
    [junit] Starting Ticker
    [junit] ------------- ---------------- ---------------
    [junit] 
    [junit] Testcase: testRemoveTwoInSameMillisecond took 0.029 sec
    [junit] Testcase: testDeduping took 0.208 sec
    [junit] 	Caused an ERROR
    [junit] null
    [junit] java.lang.AssertionError
    [junit] 	at freenet.support.PrioritizedTickerTest.testDeduping(PrioritizedTickerTest.java:230)
    [junit] 
    [junit] Testcase: testRemove took 0.002 sec
    [junit] Testcase: testSimple took 0.001 sec
    [junit] 	Caused an ERROR
    [junit] null
    [junit] java.lang.AssertionError
    [junit] 	at freenet.support.PrioritizedTickerTest.testSimple(PrioritizedTickerTest.java:139)
    [junit]

toad:
I think this might be fixed now.

toad:
No, it isn't. Sorry, working on it...

toad:
Works For Me. :)

toad:
Please re-test, review the latest changes, and re-consider. Thanks.

src/freenet/support/Ticker.java
```diff
@@ -6,9 +6,39 @@
...
+	/** Run a job at a specified absolute time.
+     * @param job The Runnable to execute.
+     * @param name The name of the thread to run it.
+	 * @param time The time at which to run the job.
+     * @param runOnTickerAnyway If true, start the job from the Ticker thread even if we could pass
+     * it directly through to the Executor. This is needed for increasing thread priorities, since
+     * the Ticker runs at a high priority, and for some tests.
+     * @param noDupes If true, ignore the job if it is already queued. Implies runOnTickerAnyway.
+	 */
+    public abstract void queueTimedJobAbsolute(Runnable runner, String name, long time,
```
bertm: This is used only supposed to be used in unit tests, am I right? If so, please make it package-private.

src/freenet/support/PrioritizedTicker.java
```diff
@@ -267,54 +252,81 @@ public Executor getExecutor() {
 
 	public int queuedJobs() {
 		synchronized(timedJobsByTime) {
-			return timedJobsByTime.size();
+			return timedJobsQueued.size();
 		}
 	}
 
+    public int queuedJobsUniqueTimes() {
```
bertm: I don't think this should be part of the public API.

Do we need this outside of unit tests? If not, please make it package-private.

src/freenet/support/Ticker.java
toad: IMHO this is more widely useful. I'm using it in some of my simulator enhancements, which are in a different package (mostly freenet.node). Not having to go back and forth between relative and absolute times all the time should result in fewer bugs, more consistency, and (marginally) improve efficiency AFAICS?

src/freenet/support/PrioritizedTicker.java
toad: This one is a fair point. Fixed.

Thynix:
Do you have reasons to want this in 1471?

toad:
Cleaned up error checks.

src/freenet/support/PrioritizedTicker.java
```diff
@@ -265,56 +250,69 @@ public Executor getExecutor() {
...
+
+	/** Remove a queued job from the internal structures other than timedJobsQueued. The
+	 * caller must check that it is present in timedJobsQueued, remove from that structure, and 
+	 * call this method, all inside the timedJobsByTime lock. 
+	 * @param job The job to remove.
+	 * @param t The time at which is it scheduled.
+	 */
+	private void removeQueuedJobInner(Job job, Long t) {
+        Object o = timedJobsByTime.get(t);
+        assert(o != null);
```
toad: I have cleaned up the error checks as requested.

toad:
No, this should be considered for 1472, not 1471. Please re-review latest changes, it should be correct now / I have taken all the feedback into account.

PR #401, opened 2015-08-17
add a NEWS file

ArneBab:
The goal of the NEWS file is to distribute the effort of gathering and summarizing noteworthy changes.

Ideally writing release notes would only consist of turning the list of changes into a sentence (adding “this release” and a few *and*’s and *also*’s).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,0 +1,73 @@
+next ():
+
+- 
+
+1470 (2015-08-15):
+
+- FIX Freemail problems that prevented sending mail
+- remove a compromised opennet seed node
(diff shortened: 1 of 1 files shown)


Thynix:
Would this be duplicating information already in the tags? In that case I'd
rather mention to see the build tags for release notes.

ArneBab:
I don’t think so: This rather collects the changes and would be a simpler source for writing release notes which then go into the tags.

The big advantage is that when someone creates a pull-requests we can say “please add a NEWS entry”. That distributes the work of creating a short overview of the changes (and as such complements the ChangeLog which provides the detailed list of the changes).

PR #403, opened 2015-08-27
themes: avoid styling headers as normal text

Thynix:
This overly broad directive caused headers in fniki and FlogHelper
previews to appear as normal text.

The only place Fred uses headers is page titles, so this moves its
styling to a selector specific to it.


Bombe:
Less styling is nice. Maybe we should use some widely-used reset.css (or something similar) and rebuild our themes based on that…

PR #404, opened 2015-08-27
Eleriseth pending keys

Thynix:
Eleriseth@WPECGLtYbVi8Rl6Y7Vzl2Lvd2EUVW99v3yNV3IWROG8.fms posted this patch set:

> This is a respin of forgotten two year old branch, pending-keys (6b30ce888c66c9ecca9e019d870559950e2df5c1).
> 
> After two years of keeping using my variant, I finally decided that toad design was better :-),
> dropped my customized hashmap, and rewritten/rebased/reordered/shuffled toad branch on top of
> 1468.
> 
> Let's rehash it from the start:
> # 
> 
> When running massive USK subscription, and I noticed interesting entry in logs:
> 
> (freenet.client.async.KeyListenerTracker, <noname>(XXX),
> NORMAL): Removed pending keys from freenet.client.async.KeyListenerTracker@XXX:SSK :
> size now 23837 : freenet.client.async.SingleKeyListener@XXX
> 
> KeyListenerTracker keeps all key listeners on ArrayList, and search it
> linearly (O(n)) on each added/removed/tripped-on-passing-key.
> 
> This worked well with "few big file downloaders (each splitfile uses single
> KeyListener to watch over all contained keys and uses bloom filter to
> pre-select "interesting" ones) and some dozens bookmarks subscriptions", but
> deals very badly with more massive USK (or other ULPR) subscriptions
> case, when there are thousands of individual listeners, and each passing
> key triggers linear search over huge KeyListener array.
> 
> This patch series replaces linear array with better structures for
> single-key listeners.
> 
> Notes:
> 
> 1) all multi-key (splitfile) listeners are kept in keyListener ArrayList, as before;
> 
> 2) if KeyListener and HasKeyListeners watch for a single route key (for CHK)
> or pubkey hash (for SSK), they should provide getWantedKey() method.
> 
> Requirements:
> 
> a) KeyListener.getWantedKey() and corresponding KeyListener.getHasKeyListener().getWantedKey()
> must return same/equal (as in "Arrays.equal") arrays (including null==null case).
> 
> b) getWantedKey() must return same/equal arrays (including null case)
> when called multiple times for given {Has,}KeyListener object.
> 
> (idea about adding HasKeyListener.getWantedKey() was taken from toad's
> 8778bc3126fb4f8d7d6cd6b07680982e5613d615 and integrated to initial
> patch, to avoid dancing forward and backward over same code).
> 
> Patch series is run-tested and works well for me (though, of course, I
> could've missed something, my testing is not really comprehensive, etc;
> careful review suggested).


src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
+				} else if(o instanceof KeyListener) {
+					if((ret = (listener == (KeyListener)o)))
+						singleKeyListeners.remove(wrapper);
+				} else {
+					@SuppressWarnings("unchecked")
+					KeyListener[] listeners = (KeyListener[])o;
+					KeyListener[] newListeners = new KeyListener[listeners.length-1];
+					int x = 0;
+					for (KeyListener l : listeners) {
+						if(listener == l) {
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
+		boolean ret = false;
+		byte[] wantedKey = hasListener.getWantedKey();
+		ByteArrayWrapper wrapper = wantedKey != null ? new ByteArrayWrapper(saltKey(wantedKey)) : null;
+		synchronized(this) {
+			if(wantedKey != null) {
+				Object o = singleKeyListeners.get(wrapper);
+				if(o == null) {
+					// do nothing
+				} else if(o instanceof KeyListener) {
+					if((ret = ((KeyListener)o).getHasKeyListener() == hasListener)) {
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
+						((KeyListener)o).onRemove();
+					}
+				} else {
+					@SuppressWarnings("unchecked")
+					KeyListener[] listeners = (KeyListener[])o;
+					KeyListener[] newListeners = new KeyListener[listeners.length-1];
+					int x = 0;
+					String msg = logMINOR ? "" : null;
+					for (KeyListener l : listeners) {
+						if(l.getHasKeyListener() == hasListener) {
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
+				return ret;
 			}
-			if(listener.getHasKeyListener() == hasListener) {
-				found = true;
-				i.remove();
-				listener.onRemove();
-				Logger.normal(this, "Removed pending keys from "+this+" : size now "+keyListeners.size()+" : "+listener);
+			for(Iterator<KeyListener> i = keyListeners.iterator();i.hasNext();) {
+				KeyListener listener = i.next();
+				if(listener.getHasKeyListener() == hasListener) {
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
 		return Math.max(0, retryCount-MIN_RETRY_COUNT);
 	}
 
+	private boolean contains(KeyListener[] listeners, KeyListener listener) {
+		for(KeyListener l : listeners) {
+			if(l == listener) return true;
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
 			// We have to register before checking the disk, so it may well get registered twice.
-			if(keyListeners.contains(listener))
-				return;
-			keyListeners.add(listener);
+			if(wantedKey != null) {
+				Object o = singleKeyListeners.get(wrapper);
+				if(o == null) {
+					singleKeyListeners.put(wrapper, listener);
+				} else if(o instanceof KeyListener) {
+					if(listener == (KeyListener)o) return;
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
 		synchronized (this) {
-			ret = keyListeners.remove(listener);
-			while(logMINOR && keyListeners.remove(listener))
-				Logger.error(this, "Still in pending keys after removal, must be in twice or more: "+listener, new Exception("error"));
+			if(wantedKey != null) {
+				Object o = singleKeyListeners.get(wrapper);
+				if(o == null) {
+					// do nothing
+				} else if(o instanceof KeyListener) {
+					if((ret = (listener == (KeyListener)o)))
```
bertm: `.equals()` should be used here instead of comparing `KeyListener`s by identity.

src/freenet/client/async/KeyListenerTracker.java
```diff
@@ -95,61 +102,197 @@ protected static int fixRetryCount(int retryCount) {
...
+		Object o = singleKeyListeners.get(wrapper);
+		if(o == null) {
+			// do nothing
+		} else if(o instanceof KeyListener) {
+			KeyListener listener = (KeyListener)o;
+			do {
+				if(!listener.probablyWantKey(key, saltedKey)) continue;
+				if(matches == null) matches = new ArrayList<KeyListener> ();
+				matches.add(listener);
+			} while(false);
```
bertm: This seems to be an overcomplicated way of writing
```java
if (listener.probablyWantKey(key, saltedKey)) {
    if(matches == null) {
        matches = new ArrayList<KeyListener>();
    }
    matches.add(listener);
}
```
Please fix.

bertm:
This change set is duplicating implementation logic of a multi-value map all over the place. I suggest to move that logic to a separate class (for it to be reusable), or at least an inner class of the KeyListenerTracker (for keeping the code simple).

What I still don't get after reading the code, is why we'd want to store CHKs in a salted `HashMap` and SSKs in an unsalted `TreeMap`. The latter seems superior here, given it is collision-free and it actually shrinks when items are removed.

Apart from the remarks above and the comments in the code, the gist behind this looks rather sensible to me. Note that these commits don't seem to comply with the coding standards.

src/freenet/client/async/KeyListenerTracker.java
ArneBab: From Eleriseth: See no point. None of KeyListener implementation overrides equals(), they
are not *supposed* to override equals(), and the use of == avoids
unneccessary null pointer checks.

src/freenet/client/async/KeyListenerTracker.java
ArneBab: From Eleriseth: Previous code used "==" too => changing that don't belong to my series,
won't change.

src/freenet/client/async/KeyListenerTracker.java
ArneBab: (same as above)

src/freenet/client/async/KeyListenerTracker.java
ArneBab: (same)

src/freenet/client/async/KeyListenerTracker.java
ArneBab: (same)

src/freenet/client/async/KeyListenerTracker.java
ArneBab: (same)

src/freenet/client/async/KeyListenerTracker.java
ArneBab: (same)

src/freenet/client/async/KeyListenerTracker.java
ArneBab: From Eleriseth: It was coded this way to match the loop below.

ArneBab:
From Eleriseth (re duplicating logic): Exactly same trick with T/[] was used everywhere in freenet, and
for a good reason (memory efficiency; given the number of KeyListener
required for full WebOfTrust and Sone subscription is quite large, it
applies here too).
Generalizing and replacing them does not belong to this patch series.

ArneBab:
I think the replies from Eleriseth address all comments, so from my side this is ready to merge.

bertm:
My usual stance would be "you touch it, you fix it", but I think we can agree to disagree on this one.
I can live with the proposed changes from a functional point of view, so lacking volunteers to address my comments, this has my ack.

ArneBab:
so all that’s needed is resolving the conflicts due to bitrot…

ArneBab:
the conflicts shouldn’t be too hard to solve, but require more testing than I could do in the time slot I have for that right now (it’s mostly just copying for loops into additional checks). It’s not obvious enough, though, that I would attempt it without knowing the reasons for the changes on both sides.

PR #405, opened 2015-08-30
Delete unused code.

toad:
Trivial.


Thynix:
It seems reasonable enough to assume no plugins are using this, and if they are they're probably doing something silly.

PR #406, opened 2015-08-30
Cleanup peernode

toad:
Slightly larger cleanups, mostly organised around "PeerNode constructor takes too many arguments". Some minor performance gain on seednodes.


Bombe:
I really like how much stuff is removed here. :)

PR #407, opened 2015-09-07
l10n: reduce configuration description aggression

Thynix:
Warnings about not to change a setting "unless you know what you are
doing" make the software seem hostile. Options are already in advanced
mode if they shouldn't typically be changed, and if someone who doesn't
know what they're doing is changing settings they are liable to do so
even if the description advises against it. If a setting really should
never be changed it should be read-only in the interface.


Bombe:
If a setting should never be changed it should be completely removed from the configuration and replaced by the values, hardcoded.

Thynix:
You're right. s/really should never/really shouldn't typically/

Thynix:
@Bombe Until we get to that, though, do you think doing this is reasonable?

Bombe:
@Thynix yep.

PR #410, opened 2015-10-10
Disallow blank bookmark names

Thynix:
_xiaoyu on IRC reported problems using blank bookmark names. When a bookmark with a blank name was added, it could not be navigated to but did not cause other problems. After restart when parsing the blank name the bookmark state became malformed somehow and prevented any operations on the list of bookmarks. These changes do two main things:

* Disallow new bookmarks with blank names to prevent the problem from occurring anew.
* Load bookmarks with blank names as "Unnamed Bookmark" to fix the problem where it was already occurring.

I'm not certain this is the best way to go about it because [existing code](https://github.com/freenet/fred/blob/caba7734a583aafc55ebe229a270f3a70eed1876/src/freenet/clients/http/bookmark/BookmarkItem.java#L60) seems like it was at least intended to [do that already](https://github.com/freenet/fred/blob/caba7734a583aafc55ebe229a270f3a70eed1876/src/freenet/clients/http/bookmark/BookmarkItem.java#L187). I didn't try diagnosing it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 next ():
 
 - add a NEWS file
+- Fix problems with blank bookmark names. Thanks to _xiaoyu for the report!
 
 1470 (2015-08-15):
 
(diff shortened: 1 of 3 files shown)


ArneBab:
Looks good; the minimal change to fix the book if I understand it correctly.

PR #411, opened 2015-10-10
fix the default bookmarks (category software not shown)

ArneBab:
This fixes mistakes I made while reorganizing the bookmarks.


Thynix:
Thanks! It's too bad we didn't catch this before release. :/

ArneBab:
Thank you for merging!

PR #412, opened 2015-10-17
improve our READMEs with build instructions

ArneBab:
This basically fixes what I complained about in devl today.

### DIFF
--- a/README.building.md
+++ b/README.building.md
@@ -2,18 +2,20 @@
 
 Install junit 4, Apache ant and Hamcrest. For Debian Wheezy these are `junit4`, `ant`, `ant-optional` and `libhamcrest-java`.
 
-If sufficient packages are not available, put these jars into the `lib/` directory:
+Freenet also requires `freenet-ext.jar` in the `lib/freenet/` directory. To get it, [download](https://downloads.freenetproject.org/latest/freenet-ext.jar) or copy from the freenet folder of an existing installation or build the `contrib/` submodule.
+
+Then just execute `ant`.
+
+You can override values of variables set in build.xml by putting them into the file override.properties in the format `variable = value`. To build Freenet without running the tests, use `ant -Dtests.skip=true`.
+
+
+If sufficient packages for the dependencies are not available, put these jars into the `lib/` directory:
 
 * `junit4.jar` and `hamcrest-core.jar`: See the [download instructions](https://github.com/junit-team/junit/wiki/Download-and-Install)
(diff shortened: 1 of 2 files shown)


README.building.md
```diff
@@ -2,18 +2,20 @@
...
-Freenet also requires in the `lib/freenet/` directory:
+You can override values of variables set in build.xml by putting them into the file override.properties in the format `variable = value`. To build Freenet without running the tests, use `ant -Dtests.skip=true`.
 
-* `freenet-ext.jar`: Build the `contrib/` submodule or [download](https://downloads.freenetproject.org/latest/freenet-ext.jar) for convenience.
 
-The dependencies could also be copied from an existing Freenet installation.
+If sufficient packages for the dependencies are not available, put these jars into the `lib/` directory:
 
-## Building Freenet from source: Using Eclipse etc
+* `junit4.jar` and `hamcrest-core.jar`: See the [download instructions](https://github.com/junit-team/junit/wiki/Download-and-Install)
```
Bombe: This is horrible. Dependencies should always be listed with _exact_ version information, otherwise developers will have a very shitty time trying to track down the correct versions.

Bombe:
This is as far away from a fix as possible, it’s just a band aid. However, barring real progress it has to do. :/

ArneBab:
This fixes the most important part: It actually gives the build command and tells people who see the README (=everyone who sees the github project) where to find build instructions.

We did not do this before, and that’s truly horrible.

README.building.md
ArneBab: I won’t add exact version information if there’s no one who maintains them. If you’re willing to maintain the information, feel free to add them.

Bombe:
The fix is a real dependency management and build system. Simply listing all the arcane incantations necessary to get Freenet compiling is not a fix.

ArneBab:
That’s what I’m not doing here, partly because we’re not even sure what we want (see the discussion in devl).

The problem you see is still different from the problem I want to fix here. I simply want to ensure that potential developers find the existing instructions with the least possible hassle.

The steps aren’t that complicated, they simply weren’t even documented sufficiently to get a jar-file which can be dropped into the freenet folder. … that’s something I forgot to add. Give me a few minutes.

Thynix:
I think we're all aware that this is a band aid; I'm hoping we can move to Gradle or something in the near term. Until that happens, this is still an improvement.

ArneBab:
Thank you!

PR #413, opened 2015-10-19
removed explicit recommendation of Truecrypt, which is discontinued and/or compromised

ploopkazoo:


Thynix:
Thanks for doing this! We discussed this a little on the mailing list
already, and were talking about linking to Wikipedia's list of FDE software
as well as explicitly suggesting a specific one like VeraCrypt. [0]
Thoughts?

I haven't heard of on-the-fly encryption before as an alternative to FDE.
What is that?

[0] https://emu.freenetproject.org/pipermail/devl/2015-October/038263.html

ploopkazoo:
@Thynix in my experience, on-the-fly crypto refers to an encrypted "container" (truecrypt volume, dm-crypt in a file, etc) whereas "full disk" tends to refer to encrypting one's OS. There isn't technically a need for the distinction as they are both mountable volumes, but people tend to make the distinction.

Linking to the Wikipedia list sounds like a good idea.

PR #414, opened 2015-10-19
changed recommendation of Truecrypt to Wikipedia's list of FDE software

ploopkazoo:


Thynix:
Thanks! I've committed this as 260c8d8a2d84abe483b7df821a84d48c7b97f63d to meet the "Write it as a command to the codebase" part of the [commit message guideline](https://github.com/freenet/fred/blob/next/CONTRIBUTING.md#commit-messages).

Before release I'd like to have these things happen:

1. Link to the wiki article ([like "TrueCrypt" did](https://github.com/freenet/fred/blob/next/src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java#L61)) instead of asking users to copy-paste.
2. Include something like "such as VeraCrypt" to avoid throwing users at a giant list without any recommendations.
3. Add a note to this string asking translators to link to the equivalent Wikipedia article in their language if it exists. This requires access to Transifex because the localization format we use doesn't support translator comments natively.

Are you interested in the first two?

Thynix:
@ploopkazoo Do you plan to make those changes, or do you want someone else to?

bertm:
> Add a note to this string asking translators to link to the equivalent
Wikipedia article in their language if it exists. This requires access to
Transifex because the localization format we use doesn't support translator
comments natively.

Would it make sense to have the URL as a separate translation unit? In the
future, we might want to use the URL elsewhere as well.

Thynix:
Sure, that would work. In the medium term I'd like to move from SFS to
gettext, which does have native comments.

ploopkazoo:
@Thynix I have no experience in Java or this templating engine, or even really git for that matter. Probably best if I'm not involved.

PR #415, opened 2015-10-20
switch the default theme to sky static.

ArneBab:
That’s the cleanest theme I found. It still looks like 1990,
but like the good 1990, not the ugly 1990.


Thynix:
Merged as a9b16b23017dfead86ec4a7904fccaafc4c992c5 to meet the "Do not put a period at the end" part of the commit message guidelines. Thanks!

ArneBab:
Thank you!

PR #417, opened 2015-10-29
Change default in test.haltonfailure to false

llamora:
When building fred, continue the junit tests once a failure is encountered by default.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -74,7 +74,7 @@ test.verbose=false
 test.benchmark=false
 test.extensive=false
 test.xml_output=true
-test.haltonfailure=yes
+test.haltonfailure=false
 
 # select a single test to run
 #test.class=


Thynix:
Sounds good to me. Any objections?

Maybe a more specific commit message can both be shorter and avoid the need for a message body - something like:

    Change default test.haltonfailure to false

Thynix:
I'll cherry-pick c044fb3.

xor-freenet:
Argh, no, please do not do this.
Nobody will notice tests failing if the stuff still spits out a JAR even though they failed: Tests produce insane amounts of text, people will not read it.
And the tests are useless if people ignore failures.

xor-freenet:
(Notice: My previous comment is written under the assumption that this change causes build.xml to produce a freenet.jar even if unit tests failed. It is possible that the particular semantics of the test.haltonfailure option actually do *not* imply this: It could merely be meant as a means to cause the junit target to run all tests even if one failed, but then still *not* spit out a freenet.jar ... If that is the case, my comment could be ignored. Albeit to be honest I still wouldn't do this then: The name of the option "test.haltonfailure" really sounds like it will ignore failures and spit out a JAR. So if it does not do that, someone in the future might think it is bugged and then "fix" it to spit out a JAR nevertheless - together with pull requests which makes haltonfailure=false the default, unit tests would then be completely useless.)

Thynix:
Whoops! Right you are. I thought this was equivalent to Python unittest's [`--failfast`](https://docs.python.org/3.3/library/unittest.html#cmdoption-unittest-f), where it stops running tests on the first failure. I don't see a way to stop that in ant's [junit support](https://ant.apache.org/manual/Tasks/junit.html), and you are right that `haltonfailure` is "Stop the build process if a test fails (errors are considered failures as well)."

Turns out this isn't what we want.

infinity0:
Please be more careful reviewing stuff like this in the future. There need not even have been a discussion; I would have close this immediately on sight. This sort of change is exactly what makes it easier for security holes to sneak in. Please be aware that intelligence services are actively trying this sort of bullshit. I am not saying the original requester is that; just that *similar shit* is being done, that is often indistinguishable from **very bad engineering** and that one must be very critical of everything.

infinity0:
One standard should be that you don't even bother looking at pull requests UNLESS the requester writes a FULL JUSTIFICATION on why they are making the change. This pull request contained absolutely no justification, so there is no need for *US* to mentally masturbate about the semantics of `haltonfailure`. IT IS UP TO THE REQUESTOR TO DO THIS THEMSELVES. This helps to defend against time-wasting attacks.

Thynix:
We did discuss the motivation on IRC; I suggested changing the default, and
it's my fault for perpetuating the misunderstanding about haltonfailure
being equivalent to --failfast.

infinity0:
If it was discussed on IRC, the justification should be transferred to the commit message. The commit logs are the only thing that is readable by reviewers after a few days.

Thynix:
The commit message does mention "When building fred, continue the junit tests once a failure is encountered by default."

infinity0:
That is not a justification - that is a redundant *description* of what the commit *does*. A justification would talk about *why* this change was made. *Why* continue once a failure is encountered by default?

infinity0:
Of course, some descriptions have pretty obvious justification like fixing typos. But this commit is about literally changing something to the *exact opposite behaviour*. There needs to be good justification for that. This applies even when changing `true` to `false` to fix a bug - one should describe what the bug *is* and why the fix is the better/correct behaviour.

Thynix:
Agreed. The reasoning was that instead of getting information on at most one test failure, one could from a single run observe all the failures to gain more information about the problem.

PR #418, opened 2015-10-31
Order alerts by type and sent time (+polish)

ArneBab:
(just what the title says)


src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -158,6 +158,13 @@ public int compare(UserAlert a0, UserAlert a1) {
 			int classHash1 = a1.getClass().hashCode();
 			if(classHash0 > classHash1) return 1;
 			else if(classHash0 < classHash1) return -1;
+			// then go by sent time, if both are n2ntms
+			boolean isN2NTM0 = a0 instanceof N2NTMUserAlert;
+			boolean isN2NTM1 = a1 instanceof N2NTMUserAlert;
+			if(isN2NTM0 && isN2NTM1) { // newest first
+				if(a0.getUpdatedTime() < a1.getUpdatedTime()) return 1;
+				else if(a0.getUpdatedTime() > a1.getUpdatedTime()) return -1;
+			}
```
bertm: Ordering by updated time is always more sensible than ordering by hash code. I don't see the need to restrict this to `N2NTMUserAlert`. Apart from that, I'd prefer the `UserAlertManager` to be as agnostic as possible of the kind of `UserAlert`.

src/freenet/node/useralerts/N2NTMUserAlert.java
```diff
@@ -117,6 +117,11 @@ public FCPMessage getFCPMessage() {
 				composedTime, sentTime, receivedTime, messageText);
 	}
 
+    @Override
+    public long getUpdatedTime() {
+        return sentTime;
```
bertm: It might be slightly more sensible to order by receive time: sent time is set by the sender, so does not necessarily lead to "newest first" behaviour (consider a friend being disconnected for some time).

bertm:
Are you aware of the possibilities `git-rebase` offers to just reword commit messages? (`git rebase -i HEAD~2`, mark all commits with `r`eword, rephrase commit messages one by one, done)

src/freenet/node/useralerts/UserAlertManager.java
Thynix: Agreed on both.

ArneBab: so I should just leave out the (isN2NTMx)?

src/freenet/node/useralerts/N2NTMUserAlert.java
ArneBab: that’s a good point… thanks!

ArneBab:
I’m using Mercurial to interface with github, there the command is called histedit, but I wasn’t in my source tree when I filed the pull request — and as you saw it took me 3 days to get into it again, so I decided to say sorry instead of delaying the pull request for a longer time ☺

bertm:
No problem ☺

Looks good to me.

ArneBab:
Thank you for reviewing!

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1039,7 +1039,7 @@ N2NTMToadlet.tooLarge=The file you have uploaded through your browser is ${attem
 N2NTMToadlet.sizeWarning=Please note that browser-based uploading has a limit of ${limit}.
 N2NTMToadlet.uploadFailed=Failed to upload file.
 N2NTMUserAlert.delete=Delete
-N2NTMUserAlert.header=From: ${from} (composed ${composed} | sent ${sent} | received ${received})
+N2NTMUserAlert.header=From: ${from} (sent ${sent})
```
Thynix: Composed might not be especially useful information, but I think both sent and received are. I know when I want to know about the timing of texts when I come out of a basement without cell coverage or something I'll want to know when it was sent, but if I come back to my phone to something like "ready to go in an hour" I'll want to know when it was received to know if I ever had a chance of responding in time. Thoughts?

I'm tired so this is too rambly but hopefully it makes sense anyway.

src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -158,6 +158,9 @@ public int compare(UserAlert a0, UserAlert a1) {
 			int classHash1 = a1.getClass().hashCode();
 			if(classHash0 > classHash1) return 1;
 			else if(classHash0 < classHash1) return -1;
+			// Then go by sent time
```
Thynix: The comment says sent time; the getUpdatedTime() implementation says receivedTime.

src/freenet/l10n/freenet.l10n.en.properties
bertm: My apologies for not catching this earlier, but `sentTime` can apparently be deceiving: it is set as the time at which the N2N message is dispatched to the Node's MessageCore ([source][0]) where it is queued for transmission over the network, if and only if the peer is currently connected. It should be really close to `receivedTime` if we ignore clock skew problems.

`composedTime` might be the one you are after knowing.

[0]: https://github.com/freenet/fred/blob/next/src/freenet/node/DarknetPeerNode.java#L610

ArneBab:
This should fix both (and the comment is now actually useful, stating that it orders newest first).

I cut out the “sent” from the alert, because I think that it’s not actually a relevant information for users, so it just adds to the line noise.

ArneBab:
thank you!

PR #419, opened 2015-11-08
Much less recentlyfailed

ArneBab:
Recently failed is one of the great annoyances when using Freenet, so I think it would be great to merge and use this.

The change likely makes it cheaper to create load by firing requests which never finish, but given that recently failed hits me at least once a week, I think this change should go in (I checked the diff and it looks sane).

### DIFF
--- a/src/freenet/node/PeerManager.java
+++ b/src/freenet/node/PeerManager.java
@@ -1160,7 +1160,7 @@ public class PeerManager {
 		
 		if(recentlyFailed != null && logMINOR)
 			Logger.minor(this, "Count waiting: "+countWaiting);
-		int maxCountWaiting = maxCountWaiting();
+		int maxCountWaiting = maxCountWaiting(peers);
 		if(recentlyFailed != null && countWaiting >= maxCountWaiting && 
 				node.enableULPRDataPropagation /* dangerous to do RecentlyFailed if we won't track/propagate offers */) {
 			// Recently failed is possible.
@@ -1247,11 +1247,12 @@ public class PeerManager {
 	}
 
 	/**
+	 * @param peers 
 	 * @return The minimum number of peers which are waiting for timeouts due to RecentlyFailed or 
 	 * DNF's for which we will terminate the request with a RecentlyFailed of our own.
 	 */
-	private int maxCountWaiting() {
-		int count = countConnectedPeers();
+	private int maxCountWaiting(PeerNode[] peers) {
+		int count = countConnectedPeers(peers);
 		return Math.max(3, count / 2);
 	}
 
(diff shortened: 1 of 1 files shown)


Juiceman:
Holy bitrot Batman! Imho useful code that has been languishing for over two years...

ArneBab:
yes…

bertm:
@ArneBab Does this patch work for you / does it actually reduce any problems you experience with RecentlyFailed? I don't usually experience any of those, except when fetching content that does indeed not exist (yet), so I can't judge.

ArneBab:
I notice much higher load on my node, though this might also be due to reverting back to WoT16 (stable Freenet). Do you have an idea how often this code is called?

I often get recently failed on content I inserted myself.

ArneBab:
I don’t think that the patch would reduce any issues with recently failed when I only run it on my own node: recently failed is sent back by other nodes, and this patch should reduce the number of recently failed we send to others.

ArneBab:
a current example is this: http://127.0.0.1:8888/freenet:USK@0iU87PXyodL2nm6kCpmYntsteViIbMwlJE~wlqIVvZ0,nenxGvjXDElX5RIZxMvwSnOtRzUKJYjoXEDgkhY6Ljw,AQACAAE/freenetproject-mirror/172

ArneBab:
@toad can this change increase the CPU load?

bertm:
This should only slightly increase the CPU load in general, but it does have some potential to increase the CPU load in the worst case.

To gain some insight in how often this change affects routing, here's the result of some instrumentation using this patch.

After 1 hr of node uptime:

countWaiting | events
---|---
0 | 65000
1 | 8630
2 | 3407
3 | 1604
4 | 932
5 | 656
6 | 487
7 | 413
8 | 427
9 | 296
10 | 229

ArneBab:
@bertm 4k recently failed vs. 77k not reporting recently failed should not be much: additional 5% of requests would error out with recently failed, if I read this right. These are, however, the cases when a site was just updated, and they block people from seeing it for half an hour, so their impact might be higher than what the statistic suggests.

toad:
Well ideally we'd like to test this in vitro. Maybe the ULPR test, and RequestInsertTest?

bertm:
This should either be reopened or be closed with a motivation.

Thynix:
This was closed inadvertently with an unintended mass branch deletion. If no one else does it I'll restore the remaining ones.

ArneBab:
I think we should deploy this. As shown by @bertm the risk is low, and the amount of recently failed creates large delays for automated near-realtime-communication like the tools by thesnark.

Thynix:
This passed when merged with next. I'll push it.

https://travis-ci.org/freenet/fred/builds/232083669

Thynix:
@nextgens expressed that this has not been tested, that there's a lot of significant changes pending in next already, and that this doesn't have clear reasoning for making the changes it makes. Therefore it is "replacing black magic with untested black magic." I agree with him that we should hold off on merging this for those reasons.

PR #420, opened 2015-11-08
enable activelinks by default

ArneBab:
Without those no web interface can give a good first time impression.
And the initial speed of Freenet is much better than 10 years ago.
Also this makes the bookmarks load faster when people click them.

### DIFF
--- a/src/freenet/clients/http/SimpleToadletServer.java
+++ b/src/freenet/clients/http/SimpleToadletServer.java
@@ -629,7 +629,7 @@ public final class SimpleToadletServer implements ToadletContainer, Runnable, Li
 		});
 		enableInlinePrefetch = fproxyConfig.getBoolean("enableInlinePrefetch");
 		
-		fproxyConfig.register("enableActivelinks", false, configItemOrder++, false, false, "SimpleToadletServer.enableActivelinks", "SimpleToadletServer.enableActivelinksLong", new BooleanCallback() {
+		fproxyConfig.register("enableActivelinks", true, configItemOrder++, false, false, "SimpleToadletServer.enableActivelinks", "SimpleToadletServer.enableActivelinksLong", new BooleanCallback() {
 
 			@Override
 			public Boolean get() {
--- a/src/freenet/clients/http/SimpleToadletServer.java
+++ b/src/freenet/clients/http/SimpleToadletServer.java
@@ -629,7 +629,7 @@ public final class SimpleToadletServer implements ToadletContainer, Runnable, Li
 		});
 		enableInlinePrefetch = fproxyConfig.getBoolean("enableInlinePrefetch");
 		
-		fproxyConfig.register("enableActivelinks", true, configItemOrder++, false, false, "SimpleToadletServer.enableActivelinks", "SimpleToadletServer.enableActivelinksLong", new BooleanCallback() {
+		fproxyConfig.register("enableActivelinks", false, configItemOrder++, false, false, "SimpleToadletServer.enableActivelinks", "SimpleToadletServer.enableActivelinksLong", new BooleanCallback() {
 
 			@Override
 			public Boolean get() {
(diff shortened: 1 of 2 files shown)


ArneBab:
(also I found that Library is enabled by default, which should not be the case, but I did not find out how to disable that)

Thynix:
I dunno about this. For my node the activelinks aren't immediately available and even though everything but the activelinks themselves loads relatively quickly, it makes to me an impression of slowness.

xor-freenet:
IIRC, they used to be enabled by default, and I was rather frustrated when the default was changed to disable them. They do spice things up, and make Freenet feel alive. Since the web interface also doesn't contain any other graphical stuff, they're an eye catcher. And because the only content accessible with a default Freenet installation currently is Freesites reachable through the default bookmarks, catching the user's eyes on the bookmarks is good.

So the concept has my ACK. Code not reviewed.

About @Thynix criticism of load time: Can't we use HTML/CSS/JS tricks to hide the images until they're fully loaded, instead of displaying the "image loading" placeholder of the browser?

Thynix:
Are we guaranteed the CSS will load before the activelink images?

I'd be more interested in the medium-term fetching bookmarks with an activelink in the background than hiding browser placeholders. I've been testing with activelinks enabled and I ran into my node's web UI being very unresponsive until I restarted it. I don't know how to reproduce the problem; it correlated with an increased probe overload rate, which I think here was caused by lower peer count. I don't know what caused that.

ArneBab:
When CSS is cached, it is loaded before the activelinks. the bugfig in 423 should fix that. For the first access that’s different, though: with the fixes in #423 CSS and activelinks are loaded at the same time. Before that traces in Firefox devtools showed that activelinks mostly went first, delaying CSS a lot.

(to sum up: webdev isn’t easy as it looks…)

I’m always running with activelinks enabled (and have been doing so for years, I think). An important point is that once activelinks have been loaded, the bookmarks load quickly (because they have already been preloaded), giving the impression that Freenet is pretty fast.

toad:
Yes but the bookmarks should be preloaded anyway, unless we've messed up with priorities (which is usually true for active Freenet users, but less so for newbies). We *do* attempt to preload bookmarks, including content IIRC.

ArneBab:
The recent two commits prevent the reshuffling of the bookmarks which happened when activelinks failed to load.

Thynix:
Before merging this I'd like some way to ensure that loading activelinks will never block page CSS.

ArneBab:
There will not be a “never”, but when merging #423, CSS will be cached for new users when they start the wizard (no activelinks), and for all others the CSS cache will only expire once a month. In addition, the optimization of CSS loading done in #423 makes CSS load in parallel with the activelinks in my tests.

The one thing which can get slower is clicking reload for the browse-freenet page, because this makes the browser re-fetch the cached CSS files. In my test with roughly 100 activelinks, about half of them not existing (worst case), issuing a reload (F5) can take 2 seconds to get all CSS files again. In contrast, when navigating within fproxy, the CSS files are not accessed at all (not anymore with #423 ).

xor-freenet:
I wonder whether your impression of poor responsiveness of the web interface using this code can be explained by the recently fixed severe breakage of thread priorities in all of fred? (https://github.com/freenet/fred/pull/435)
Could you test this again with that code?

Thynix:
Good point. I will, but not until after 1471 is out.

bertm:
A good way to test if enabling activelinks by default is sensible is to set your client cache to "none", completely break your internet connectivity, and then again test whether the index page loads acceptably fast.

According to my experience, at this point we cannot enable activelinks by default.

Thynix:
Agreed.

ArneBab:
@bertm damn… that’s something to work on, then. Thank you for the test! I wonder whether we could proactively preload activelinks and send 404’s at once when they aren’t available.

nextgens:
We've talked about it in the past... we could also pre-load the default datastore install with freeblobs containing the content we expect new users to request...

There's no limit to how much we can cheat here. Does it help anything? I'm not sure.

ArneBab:
I think it would help a lot: It would make the first page users see look somewhat interesting and load fast. Pure text links are as boring as it gets, and the impression about the performance from the first clicked links is likely to stick.

ArneBab:
added a commit which removes the change in default, since we can activate activelinks per theme. Also activated it in next for winterfacey.

PR #422, opened 2015-11-13
pass the isOfficial flag to the plugin(api)

saces:
pass the isOfficial flag to the plugin so the official plugin can point to mantis as bugtracker and stuff.

### DIFF
--- a/src/freenet/pluginmanager/PluginRespirator.java
+++ b/src/freenet/pluginmanager/PluginRespirator.java
@@ -332,4 +332,8 @@ public class PluginRespirator {
 	public void storeConfig() {
 		pi.getConfig().store();
 	}
+
+	public boolean isOfficialPlugin() {
+	    return pi.isOfficialPlugin();
+	}
 }


bertm:
What's the point of having Fred tell plugins whether they are official?

The plugins probably know that themselves (i.e. the author declares them official —for whatever that may mean in the context of the particular plugin: Fred doesn't base it's behaviour on whether the plugin is official— in the plugin code). Or do I happen to miss some obvious use case for this?

Bombe:
I also can not see a valid reason for this to be included.

Thynix:
@saces Can you elaborate on what this is for, please? If it's to be used somewhere in Fred these changes should include something that uses it.

saces:
Its more eye candy than real function: the official KeyUtils (the plugin loaded through the Official* loader to be more precise) shows only the official bugtracker (bugs.freenetproject.org), all other versions show teh github.com/saces/KeyUtils/issues as primary bugtracker...
At leat this was the idea behind.

feel free to reject :tinfoil:

Bombe:
@saces “all other versions?” How many other versions are there? And why?

saces:
I don't like it anymore. kinda.

PR #423, opened 2015-11-18
Theme sky static optimize css parallel

ArneBab:
shaves off 0.1 to 0.5 seconds loading time for me.


ArneBab:
We could speed this up some more by always including base.css directly in the HTML header, but that would reduce the flexibility of themes to just not include base.css

bertm:
This feels like patchwork to me. Wouldn't it be more appropriate to have the `StaticToadlet` send headers that encourage the browser to cache the static files? That will shave off even more time, given the browser does not have to request the files at all, except for the first load.

ArneBab:
@bertm you’re right — and this is what I just did

bertm:
Nice catch! I'll be happy to review the pull request fixing that bug. ☺

I don't think merging the CSS files is necessary – the first load will be slow anyways due to the nature of Java, and I appreciate the semantic separation of having several files. However, I'm not strongly opposed to it either.

Thynix:
I'd appreciate keeping the files separate. Enabling caching seems the way
to go. We could also look into optimizing static asset response time.

ArneBab:
There’ll  be one more commit this evening or tomorrow which fixes the caching.

ArneBab:
it’s in now: https://github.com/ArneBab/fred-staging-1/commit/cf693b445c0fbc0932cd97255d1b7f7ea663bea5

An alternative to merging the files in basic would be to make base.css only include other files and reference those the files directly in the theme.css. This would turn base.css into starting point for themes.

The problem we face is that when we have chains of CSS includes, the browser needs to fetch them one after the other. As soon as they are cached that does not hurt anymore, but they won’t always be cached — especially not when users open the web interface the first time on a given day.

Juiceman:
From the 10 minutes of reading up on CSS I just did it looks like @import in general is a bad idea because each @import is called serially preventing parallel loading of CSS files. Add to that we have CSS calling other CSS which also does @import and no wonder it is slow.

bertm:
The caching patch looks good.

As for parallelism, can't we get away with using additional `<link rel=…>` constructs instead of `@import`s? This only requires a minor extension to `PageMaker`. I'll try and see if I can come up with something.

ArneBab:
Those <link rel=...> constructs would limit the flexibility of themes (currently they can control all included CSS). A one-step import still works quite well from what I saw. Two-step import (theme (@import base (@import other stuff))) causes the imports to be done after the first batch of activelinks (=horribly slow).

bertm:
Hmm. Your caching patch does not cause Firefox to actually cache the data, despite the availability of a valid `Expires` header. Maybe `Cache-Control` is needed as well?

> Those constructs would limit the flexibility of themes (currently they can control all included CSS).

I'm not sure I follow. CSS is additive, how would changing the way of inclusion have any effect on the scope of the CSS? Surely the order of inclusion is important, but given that `@import` must be the first statement(s) of a CSS file (except for `@charset`), the proper inclusion order is easy to determine. Please see #425 for an implementation of my approach, which appears to work well.

bertm:
`Cache-Control: public`, together with your `Expires` fix, seems to solve the caching.

→ bertm/fred-staging@8cf7f6ae152569a2060e45e08a4b04087cd4abc6

ArneBab:
@bertm that Expired doesn’t suffice is strange. It works for me… Do you run private mode?

I’ll have a look into Cache-Control to see whether it has additional side effects.

Why did you add Cache-Control it at another place than the expired header?

ArneBab:
@Juiceman yepp. One level of imports gives themes flexibility (the themers decide what’s defined — and it’s explicit which definitions are tehre). Two levels mostly create problems.

ArneBab:
If we use cache-control, the clean way of working would be to specify a very long max-age along with a content-specific value as etag (as long as that does not change, the cache stays valid). It would be nice to use the build number, but I’m not sure whether we have that in the code.

Info: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching

ArneBab:
I now cleaned up the implementation. The etag is the simplest hash possible (just taking the contentLength of the file) which should work pretty well at avoiding refetching.

src/freenet/clients/http/ToadletContextImpl.java
```diff
@@ -376,15 +376,32 @@ static void sendReplyHeaders(OutputStream sockOutputStream, int replyCode, Strin
...
+		String expiresTime;
+		String cacheControl;
+		if (allowCaching) {
+			// use an expiry time of 30 day from now, about the frequency of Freenet releases
+			// Expires is needed for older browsers
+			expiresTime = TimeUtil.makeHTTPDate(System.currentTimeMillis() + DAYS.toMillis(30));
+			cacheControl = "public, max-age=" + String.valueOf(3600 * 24 * 30);
+			// Use contentlength as simplest possible hash function
+			// which needs no additional file access
+			mvt.put("ETag", String.valueOf(contentLength));
```
Bombe: You’d be better off using e.g. the start time of the servlet here. The length not really a good hash value.

bertm:
> @bertm that Expired doesn’t suffice is strange. It works for me… Do you run private mode?

I don't run private mode. I don't know why it didn't work either.

> Why did you add Cache-Control it at another place than the expired header? (or didn’t you? I only looked at the diff on github till now)

Not for any particular purpose as far as I remember. Probably didn't read the code surrounding your patch well enhough.

> [Re: using Cache-Control with max-age and ETag] I would not depend on this alone — we should also include expires, for example for older text browsers.

Agreed.

As for `ETag`, if we set that, we'd also need to ensure that `Toadlet`s (at least `StaticToadlet`) return `304 Not Modified` on a matching `If-None-Match`.

src/freenet/clients/http/ToadletContextImpl.java
bertm: Start time of the servlet sounds like a good idea. I don't think we should `put` here if `mvt` already contains an `ETag` header: `Toadlet`s might want to override it for cacheable yet non-static content.

ArneBab: Start time of the servlet would mean that the file would have to be re-loaded everytime freenet restarts, right? If that’s the case, then it would defeat the purpose: The CSS files would have to be re-requested in a situation where the user is most likely to open the browse page with the activelinks — that’s the page where cached CSS files help the most.

The length is not a good hash value, but we don’t need a good hash. All we need is that the value changes when the file is changed between releases. So it does not need to change on every possible file change, only on ones which are likely during development — and even if it does not change, the file should be re-requested once a month (though I’m not certain that I’m right on this). I did not use the md5 hash, because we don’t have the file content at this place in the code and I don’t want to do too invasive changes.

Another good value could be the build number of Freenet. If that isn’t known, the ETag should likely be avoided.

Please refrain from using a git hash: that would break my builds (and likely packages, too).

ArneBab:
The specs say that Expires should only be adhered to if Cache-Control is present, so my FF seems not to act according to spec.

A 304 is a “should” not a “must”, so we can leave it out. The intention of the caching is mainly to avoid a request in the first place: If we have to send headers, most of the latency cost is already paid (since the static assets are small). So I’m not sure whether the additional complexity is worth the potential gain.

It might be, if we restructure the response such that it does not access the file content if the ETag matches. But that would require some bigger changes (so I think we shouldn’t do it right now, I worry about the risk to break something unrelated and incurring a longer bugfixing phase).

src/freenet/clients/http/ToadletContextImpl.java
Bombe: There are virtually unlimited changes you can perform on a file without changing its length so the length of a file is pretty much the worst hash value you could ever have (except for a constant value).

What we need here is a value that does change if the file changes; servlet start time would be an option, build version is slightly worse (because of pre-releases which have the same build version as other builds and the files might have changed), even the real hash would be okay because we only need to calculate it once; static assets can additionally be cached because they don’t change at runtime: if they have already been requested once during Fred’s lifetime they are guaranteed to not have changed since then.

ArneBab: @Bombe do you know off-hand how to get the Freenet version string? That’s something I’d love to use here. If there isn’t a version string, it’s a development version.

Can we access the checksum of the Freenet jar file?

(for the sake of the argument: There are also unlimited changes you can perform to a file which do change its length — many more than the changes you can perform which do not change the length, so the chance of hitting one which does not change the length is pretty slim. Algebra of infinites is fun ☺)

Bombe: The node’s version can be found in `freenet.node.Version`.

ArneBab: @Bombe thank you! I’ll have a look at using that as ETag later.

Bombe: @ArneBab why would you want to use something that is not guaranteed to change when a file changes when you have something available that is guaranteed to change when the file changes? That’s what the ETag is about: detecting changes. If you are really worried about caching files across restarts, use the hash. Stop trying to find supposedly cheaper shortcuts that will only make things worse by royally messing up the caching!

bertm:
Wouldn't it be pointless to send an `ETag` without listening to `If-None-Match`? How would an `ETag` improve caching beyond `Expires` and `Cache-Control`?

src/freenet/clients/http/ToadletContextImpl.java
ArneBab: I want to avoid messing too much with the structure of the code. We have neither the file content nor the file path at this point. This is not about cheaper, it is about avoiding to restructure code for this change.

I want to keep this pull-request short, and rather file a second one to improve on it.

Bombe: So you want to forcibly break it by using a non-unique ETag and then fix it later? I can not agree with this course of action.

ArneBab: no, I want to improve it over the current state of affairs without reaching for the perfect on one step.

ArneBab:
@bertm you’re right. Let’s just leave it out. That also saves us from all the complexities I discussed with @Bombe

Bombe:
@ArneBab you need to learn to differentiate between “not making something perfect” and “breaking something.” ;)

ArneBab:
@bertm could you check whether the PR is merge-ready now? I’d like to see this in 1471 (I only now realized that it didn’t go in because deleting n2n text messages is awfully slow — due to having to load every asset every time).

bertm:
Yes, this looks good. I'll give it a final test tonight.

bertm:
According to my testing this does indeed work as intended—but it does not load all of the CSS before the activelinks are loaded in Firefox, which matters when loaded for the first time.

Thynix:
I don't think that's a regression? Hence activelinks being disabled by
default. If that's the only issue then this sounds ready to merge. Any
other reservations?

ArneBab:
It’s not a regression, no. @bertm created a pull request which would ensure that all CSS is loaded before the activelinks, but it hardwires the imported paths, so theme creators would have to also adjust Java code to get their theme to work well (entangling themes and toadlets), so I asked to put that back for the moment and first try other ways to improve performance.

Not loading CSS before the activelinks should not matter for new users, because they will have first seen the wizard, so all CSS should already be cached when they see the activelinks.

Thynix:
@ArneBab If we're looking to merge this in the next release can you please rebase it onto `1471-stable`? It is currently based on `next` so merging it brings changes from `next`.

ArneBab:
@Thynix is it ok to just merge 1471-stable into the version before I merged next?

ArneBab:
thank you!

PR #424, opened 2015-11-20
Add warning message about leftover plugin files

vwoodzell:
Added a warning that files stored by a plugin are not removed when the plugin is unloaded, as requested in bug 6558. I didn't add the FredPluginWithUninstallation interface, because it doesn't make sense until we have functionality for it.

### DIFF
--- a/src/freenet/clients/http/PproxyToadlet.java
+++ b/src/freenet/clients/http/PproxyToadlet.java
@@ -198,6 +198,9 @@ public class PproxyToadlet extends Toadlet {
 				HTMLNode infoboxContent = infobox.addChild("div", "class", "infobox-content");
 				infoboxContent.addChild("#", l10n("pluginUnloadedWithName", "name", pluginThreadName));
 				infoboxContent.addChild("br");
+                infoboxContent.addChild("#", l10n("pluginFilesWarning"));
+                infoboxContent.addChild("br");
+                infoboxContent.addChild("br");
 				infoboxContent.addChild("a", "href", "/plugins/", l10n("returnToPluginPage"));
 				writeHTMLReply(ctx, 200, "OK", pageNode.generate());
 				return;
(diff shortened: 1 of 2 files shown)


Bombe:
Good choice to not add yet _another_ interface to Fred’s already-messed-up-enough plugin system!

bertm:
> I didn't add the FredPluginWithUninstallation interface

Thanks for not doing that. We wouldn't be able to guarantee that the plugin had properly erased its traces anyhow.

That said, after unloading, we actually don't know whether or not the plugin has performed any cleanup. Your proposed text implies it has not—which is indeed quite likely, but it feels somewhat over-generalizing.
Lacking suggestions for a better phrased alternative, I think this is good to go.

xor-freenet:
Thank you for implementing this! I'm very glad if people deal with long term security issues which don't cause any immediate pressure, this is wise behavior :)

> Good choice to not add yet another interface to Fred’s already-messed-up-enough plugin system!

Interfaces have been our standard way of telling fred what plugins do, we have over a dozen of them IIRC.
So I don't see any other solution than adding an interface.
Do you?

> I didn't add the FredPluginWithUninstallation interface, because it doesn't make sense until we have functionality for it.

The point of having a whitelist WITHUninstallation interface (blacklist would be "WITHOUTUninstallation") which only displays the warning if the interface is *not* implemented is: Plugins which were written before the whitelist interface was added will yield correct behavior without any changes: They don't implement the WithUninstallation interface yet, so the node will warn that they don't have uninstallation.

The opposite of a blacklist WITHOUTUninstallation which would only display the warning if the interface *is* implemented would be: We would have to adapt ALL plugins now when trying to release this, to ensure the warning is displayed for plugins which do not implement uninstallation yet.
So this would cause a lot of maintenance overhead just for the initial release.

So the suggested whitelist approach is the ideal migration scheme as it causes the least work for the initial release.
Conclusively, your decision to not add it yet "because it doesn't make sense until we have functionality for it." does not fit the reasons behind this interface choice: The goal with this choice is to not require any changes while we not have functionality for it. So it doesn't make sense to use the "not have functionality" part for a decision against it, as that is the basic assumption behind having that interface in the first place. 

> Thanks for not doing that. We wouldn't be able to guarantee that the plugin had properly erased its traces anyhow.

By design decision plugins are trusted code: They get full access to the fred API. Also, we only ever add plugins as official if we did review their code.
And since we've decided they're trusted code, we could trust them to actually delete stuff it they did implement the interface to say that they do. Why would someone implement an interface which says it has uninstallation, but then not actually write the code anyway? Easier to just not implement that interface.

> That said, after unloading, we actually don't know whether or not the plugin has performed any cleanup. Your proposed text implies it has not—which is indeed quite likely, but it feels somewhat over-generalizing.

Which is why I suggested the interface to ensure that plugins which do delete their stuff can disable the warning :)

vwoodzell:
Sorry, the functionality to which I was referring was fred calling uninstall(). If that's not called, then it's impossible for a plugin to clean up its files, unless I'm misunderstanding something. And if the plugin can't clean up its files, then clearly it shouldn't disable the warning.

toad:
That's a good idea IMHO.

PR #425, opened 2015-11-21
<link> to CSS dependencies

bertm:
This replaces the CSS `@import` declarations with top-level `<link>` declarations of unique includes in the HTML in the correct order of appearance. To be able to do this, we could either parse/modify the CSS (which is doable but hard), or hard-code the dependency graph in Java. Since themes are hard-coded in `PageMaker` anyhow, I opted for the latter approach.

The original `@import` statements have been omitted from the CSS files accordingly, and replaced by a comment explaining where the import has gone.

This can be seen as another approach to #423.

### DIFF
--- a/src/freenet/clients/http/PageMaker.java
+++ b/src/freenet/clients/http/PageMaker.java
@@ -2,7 +2,10 @@ package freenet.clients.http;
 
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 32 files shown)


ArneBab:
This closely couples themes and code, so themers would have to change Java code, while before they could do all the changes in pure css. From a performance perspective it’s fastest (all parallel), but it’s much stronger coupling of different kinds of code.

bertm:
You're absolutely right about the coupling added by this changeset.

I'm curious what approach the others prefer and why. I won't feel offended if this doesn't get merged in the end. ☺

ArneBab:
Thank you for that! I appreciate the amount of work which went into this change, but from a project planning viewpoint I think it goes into the wrong direction — though it would give the fastest load time of the first page. I think that maintenance is more important here (that’s also why I split up the base.css in #423 again), because that first page will likely be the wizard, so when users first see a page with activelinks (were loading a second level of CSS in serial can be very slow), the CSS should already be cached.

ArneBab:
I wonder whether it would be easy to parse the theme.css file and add all imports of the current theme file as <link>. It could mess up the ordering, though…

bertm:
That was my first approach, the code for generating the dependencies in the right order is actually quite simple. In fact, the order is unimportant, as long as we link the original stylesheet (with all its `@import`s left intact) last.

Then I decided that writing a CSS parser that correctly parses all possible forms of `@import` (excluding those in comments) was hard, and that the code would unnecessarily bloat `PageMaker`. Now I think of it again, it could of course as well be a separate class, but I don't particularly feel like writing a CSS parser.

ArneBab:
that could hurt, yes… maybe we can leave it for later, if with the parallel CSS fix the performance is still not good enough?

bertm:
Alright then, no reason to merge this (or the proposed idea) now. Let's first see what improvements some caching and your proposed fix bring.

PR #426, opened 2015-11-25
BaseL10n: fix Bokmål slug

Thynix:
Thanks to vwoodzell for the report.


Thynix:
@vwoodzell Is this right?

vwoodzell:
I don't know anything about the language name and code, but it loads the properties file now.

PR #428, opened 2015-12-01
Fix improper indenting & logging bug caused by it

Pascal666:
Not found SSKs were always being logged as found in cache.


Bombe:
Well, yes. That logging statement was definitely not in the correct place. :)

bertm:
Nice find. May I suggest adding braces around the if's effect (in accordance with the coding standard)? This would prevent future introduction of similar bugs.

PR #429, opened 2015-12-01
filter: add minimal support for audio tag

ArneBab:
We have an mp3 filter and current browsers support mp3 in the audio-tag, so this provides actual capabilities to users.

The mp3 filter might be too strict (it butchers my test file), but that’s a separate fix.

### DIFF
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -1255,6 +1255,16 @@ public class HTMLFilter implements ContentDataFilter, CharsetExtractor {
 				new String[] { "href" },
 				emptyStringArray,
 				new String[] { "onfocus", "onblur" }));
+		allowedTagsVerifiers.put(
+			"audio", // currently just minimal support
+			new CoreTagVerifier(
+				"audio",
+				new String[] { // allowed tags
+					"preload",
+					"controls"},
+				emptyStringArray, // uris
+				new String[] { "src" }, // inline uris
+				emptyStringArray));
 		allowedTagsVerifiers.put("style", new StyleTagVerifier());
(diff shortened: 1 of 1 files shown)


bertm:
This looks good.

Would you care to open a bug for your failing file?

Thynix:
I'd like to hold off on trying to get this into 1471.

ArneBab:
holding off sounds good. There are already changes for filters in the next release.

toad:
Should we attempt to force the MIME type in src= ? Some browsers may still do odd things e.g. with plain text, application/octet-stream etc?

ArneBab:
For images that’s no issue, I’m not sure here. The content filter should garble any non-mp3 file which claims to be a mp3 beyond recognition, right?

And any other file should not get served (the audio-tag does not show the “filter failed” site).

The following is the test code which can play the audio-file, but garbled.

    <audio src="/CHK@VGUnIhnnceDNObxaaYH8XxCLYAJhJnFBHjZoir81PeQ,W1oSdk31ibN8sdpvbe3MDudsjCLZ~qeW63JHaDOQ0c4,AAMC--8/das-klein-wild-voegelein.mp3" type="audio/mpeg" style="height: 20px" controls="controls" preload="auto" ></audio>

ArneBab:
here’s the bug with the failing file: https://bugs.freenetproject.org/view.php?id=6787

ArneBab:
The bug for mp3s is fixed in #478 

So once me merge both, people can use the audio-tag on Freesites.

ArneBab:
@bertm it’s now forcing MIME to audio/mpeg, since browsers would try to play text/plain as mp3, but the filter wouldn’t fix attacks.

PR #430, opened 2015-12-02
HTMLFilter: fix processing of reset input element

bertm:
This fixes an obvious typo in HTMLFilter that prevented `<input type="reset">`-style form elements from passing the filter.


Bombe:
Well spotted.

PR #432, opened 2015-12-07
Random fixes

Bombe:
Some easy fixes for a couple of problems detected by IDEA.


src/freenet/client/async/SplitFileFetcher.java
```diff
@@ -451,14 +447,8 @@ public void onResume(ClientContext context) throws FetchException {
...
-        try {
-            if(storage.start(resumed))
-                getter.schedule(context, storage.hasCheckedStore());
-        } catch (KeyListenerConstructionException e) {
-            Logger.error(this, "Key listener construction failed during resume: "+e, e);
-            fail(new FetchException(FetchExceptionMode.INTERNAL_ERROR, "Resume failed: "+e, e));
-            return;
-        }
+        if(storage.start(resumed))
+            getter.schedule(context, storage.hasCheckedStore());
```
bertm: While you're touching this line, you might as well want to add braces.

bertm:
Assuming unit tests still pass, for which I see no reason they would not, this looks good. Thanks!

src/freenet/client/async/SplitFileFetcher.java
Bombe: So true. :)

PR #435, opened 2015-12-17
Fix bug #6623 : Wrapper.log: Lots of "setNativePriority(X) has failed!"

nextgens:
This is something I've introduced in 7487c51c32e673bde5e89a148b449179d8fc53b8

### DIFF
--- a/src/freenet/support/compress/RealCompressor.java
+++ b/src/freenet/support/compress/RealCompressor.java
@@ -112,7 +112,7 @@ public class RealCompressor {
     public static class CompressorThreadFactory implements ThreadFactory {
         @Override
         public Thread newThread(Runnable r) {
-            return new NativeThread(r, "Compressor thread", NativeThread.MIN_PRIORITY, false);
+            return new NativeThread(r, "Compressor thread", NativeThread.MIN_PRIORITY, true);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


nextgens:
@toad thinks it's a must-have for next release

toad:
Yes, but we should fix the API too before closing #6623.

toad:
This is the correct fix for build 1471, i.e. a one line fix and javadocs. I have filed a bug for a longer term solution.

Thynix:
Bug in question is https://bugs.freenetproject.org/view.php?id=6743

This fix is the same as the patch in https://bugs.freenetproject.org/view.php?id=6623 and as bertm says

> This patch does not address the underlying problem causing the messages. It merely disables priority checking of the new thread, which would fail otherwise, because apparently we are creating this NativeThread from a low-priority NativeThread (that appears as having been reniced).

I'd rather not merge a half-measure like this that just disables the logging.

nextgens:
Respectfully, @bertm is wrong :)

This is not a half-measure, it's the real fix. I'm the one who introduced the said buggy code and we've extensively discussed it with toad already.

The check behind the boolean doesn't make much sense (and yes it's also my code) and it definitely doesn't in this specific instance.

Thynix:
Oh. Alright.

bertm:
You all seem to be reading my comment out of context. This does not fix the logging behaviour bug 6623 reported. **If** NativeThreads are somehow disabled—i.e. Fred is reniced, which it shouldn't be, but heh—it will still spit out hundreds of `setNativePriority(…) has failed!` lines. However: that is *only a logging bug* that this pull request does not seem to intend to, nor has to, fix.

I do however agree that this fixes this particular instance where NativeThreads are disabled inappropriately, and give my full ack for merging.

Thynix:
Whoops, I was not intending to quote you out of context. Sorry about that.

PR #437, opened 2015-12-19
Fix JVM version comparison for version >1.9 and build >99

bertm:
We now properly parse JVM version numbers, instead of comparing them
lexicographically.

This resolves bug [6390](https://bugs.freenetproject.org/view.php?id=6390).


bertm:
Fixed trailing whitespace.

PR #439, opened 2015-12-19
Remove DiffieHellman related code (pre-neg9)

nextgens:
This has been dead code for a while now... It doesn't affect functionality and should be safe to merge as is.

I've just bootstrapped an opennet node with it (it has been tested).


bertm:
This does break bootstrapping for pre-neg9 nodes to seednodes once they have updated. Do we consider that a problem? Those nodes are ancient anyhow, so that might not be a problem at all (I recall some recent discussion on this subject, but I don't know what the final consensus was).

Apart from the above, looks good to merge from reading, will test as well in a minute.

Thynix:
pre-neg9 nodes are years old and not worth maintaining (if we can call it that) this cruft for. It's putting in a lot of effort for non-users of the software. negtype 9 has been mandatory since 1448 went mandatory on July 23rd, 2013, and support for negtypes before 9 was disabled in build 1465, released on August 9th, 2014.

bertm:
Alright! ☺

Bootstrapping tested OK, just as expected. This has my ack.

PR #440, opened 2015-12-20
Cleanup noderefs

nextgens:
This is preparing for next phase: 623a239 should be cherry-picked even if the rest isn't

Odds are we can't remove more of the DSA code until this has been deployed and is mandatory (without breaking opennet)

### DIFF
--- a/src/freenet/clients/fcp/AddPeer.java
+++ b/src/freenet/clients/fcp/AddPeer.java
@@ -156,7 +156,7 @@ public class AddPeer extends FCPMessage {
 			} catch (ReferenceSignatureVerificationException e) {
 				throw new MessageInvalidException(ProtocolErrorMessage.REF_SIGNATURE_INVALID, "Error adding ref: "+e.getMessage(), identifier, false);
 			}
-			if(Arrays.equals(pn.getPubKeyHash(), node.getOpennetPubKeyHash()))
+			if(Arrays.equals(pn.peerECDSAPubKeyHash, node.getOpennetPubKeyHash()))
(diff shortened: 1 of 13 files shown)


nextgens:
this needs testing, including whether seednodes running the new code can be used by old nodes

toad:
I agree that we should cherry-pick the above fix. However for the rest, I don't understand why this works at all. E.g. even when announcing / path folding, PeerNode will expect a signature, and we don't include one.

toad:
The consequence of this is that a new node can't be added to an old node via announcement or path folding. Unfortunately that means UOM won't work! We do need symmetry, in the short term.

toad:
The main problem with this is that innerProcessNewNoderef() needs the same changes made to the constructor in the last commit: Identity is optional for opennet nodes.

toad:
AFAICS the worst case consequence of this is opennet reconnections not working, since we call innerProcessNewNoderef() in completedHandshake(). Also breaking ARKs, and lots of log spam.

nextgens:
This fixes the obvious (log-spam); ARKs will still be broken... but eventually the node will just re-announce or the peer will upgrade and publish an ARK with an identity line

I've tested:
- starting a fresh node running PR440 and boostraping onto the existing network
- starting a fresh node running PR440 and boostraping on a fresh seednode also running PR440; this was slow and the first peer it got was the seednode itself!
- starting an old node and boostrapping it on the fresh, updated seednode

This should cover most of what we care about.

nextgens:
12:16 <@toad_> okay so it doesn't even break ARKs
12:16 <@toad_> awesome

IMHO it's good enough to receive testing in the pre-release

Thynix:
Alright, sounds good. I don't yet know if I'll have time this evening to
merge it.

PR #442, opened 2015-12-24
Tell user when we drop old darknet peers

toad:
Thanks to nextgens' code, 1471 will drop a large number of very old darknet peers for many long-term users. We should tell the user about this. Unfortunately this is a relatively large patch, because of the need to introduce a new exception.


toad:
Also it may be a bit ugly in terms of creating UserAlert's directly from PeerManager. But IMHO we should consider having something like this in 1471. The alternative is a bunch of NullPointerException's and an opaque warning only in the logs (and the peers disappearing).

Bombe:
What about not yet merging @nextgens’ code, or is it already too late for that? :)

Thynix:
The changes this is in response to was merged in #440.

toad:
Note that this is not a short term fix to be reverted soon. This is a long term solution to a problem that will likely recur in future: We do need to be able to change peer crypto. E.g. we may want to use djb ciphers in future, replacing or in addition to the current ECDSA.

src/freenet/node/PeerManager.java
```diff
@@ -303,7 +346,39 @@ private boolean readPeers(File peersFile, NodeCrypto crypto, OpennetManager open
 		return !someBroken;
 	}
 
-	public boolean addPeer(PeerNode pn) {
+	private void reportDroppedOldPeers(List<String> droppedOldPeers, int droppedOldPeersBuild,
```
nextgens: Can't you stick the user-alert creation method elsewhere? it belongs to freenet/node/useralerts/ imo

src/freenet/node/PeerNode.java
```diff
@@ -489,6 +493,11 @@ public PeerNode(SimpleFieldSet fs, Node node2, NodeCrypto crypto, boolean fromLo
 
 		/* Read the ECDSA key material for the peer */
 		SimpleFieldSet sfs = fs.subset("ecdsa.P256");
+		if(sfs == null) {
+		    GregorianCalendar gc = new GregorianCalendar(2013, 6, 20);
```
nextgens: This should be static... but I guess it's a rare-enough occurrence

nextgens:
I agree with @toad ; It's not specific to #440 ... we will need it down the line too.

toad:
I have factored this out into a separate useralert in the proper place.

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -475,6 +475,9 @@ DownloadFeedUserAlert.delete=Delete
 DownloadFeedUserAlert.fileDescription=Description:
 DownloadFeedUserAlert.fileURI=URI:
 DownloadFeedUserAlert.title=${from} recommends you a file
+DroppedOldPeersUserAlert.droppingOldFriendTitle=Warning: Dropping ${count} connection(s) due to new encryption protocol (your friend(s) haven't upgraded or connected since ${buildDate}).
+DroppedOldPeersUserAlert.droppingOldFriendFull=Freenet has dropped ${count} friend connection(s) because they haven't updated since ${buildDate}. Build ${buildNumber} included an important change to Freenet's encryption, and we cannot maintain backwards compatibility to really old nodes because older encryption protocols are vulnerable to attack. The old node references will have been saved in ${filename}.
```
Thynix: s/will have been/are/ ?

Thynix:
I've merged this in 6e2434991b11bfe116030b811eb1eb5bb3c4fa66, and we inadvertently did duplicate work, so I took my ".broken.broken" fix - 4acd93fbf977435d49ee4fe8caa362dfb791af33 - and your a5f9532a9ba93e3bebd42572658b5be710700784 and eb399504aa8acc9069675a9a41f81168223d8a51.

PR #443, opened 2015-12-24
Re-check the stores list periodically

toad:
Fixes a nasty race condition which could happen with late initialisation
of the client-cache (e.g. due to password encryption), would cause an
infinite loop.

Please consider for 1471 (as this fixes a regression in next).

### DIFF
--- a/src/freenet/store/caching/CachingFreenetStoreTracker.java
+++ b/src/freenet/store/caching/CachingFreenetStoreTracker.java
@@ -130,11 +130,12 @@ public class CachingFreenetStoreTracker {
 	
 	private void pushAllCachingStores() {
 		CachingFreenetStore<?>[] cachingStoresSnapshot = null;
-		synchronized (cachingStores) {
-			cachingStoresSnapshot = this.cachingStores.toArray(new CachingFreenetStore[cachingStores.size()]);
-		}
 		
 		while(true) {
+		    // Need to re-check occasionally in case new stores have been added.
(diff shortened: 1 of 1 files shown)


src/freenet/store/caching/CachingFreenetStoreTracker.java
```diff
@@ -130,11 +130,12 @@ public void run() {
 	
 	private void pushAllCachingStores() {
 		CachingFreenetStore<?>[] cachingStoresSnapshot = null;
```
Bombe: Can you move this line, too?

src/freenet/store/caching/CachingFreenetStoreTracker.java
```diff
@@ -130,11 +130,12 @@ public void run() {
...
-		synchronized (cachingStores) {
-			cachingStoresSnapshot = this.cachingStores.toArray(new CachingFreenetStore[cachingStores.size()]);
-		}
 		
 		while(true) {
+		    // Need to re-check occasionally in case new stores have been added.
+	        synchronized (cachingStores) {
```
Bombe: Indentation seems off…

toad:
Now the diff looks even worse, but at least it's small.

Do you really want me to reindent the whole function? That's the only way Eclipse will actually use spaces consistently. :(

toad:
Also somebody should configure the github viewer to use 4-space tabs to avoid such confusion :)

PR #444, opened 2015-12-26
Localise FDE hyperlink

toad:
Do we want to localise documentation hyperlinks? This may increase the need for reviewing translations slightly? Or it may not - malicious translators can do all sorts of bad things anyway?


bertm:
In this particular case I think it is sensible to allow translator to substitute a native language Wikipedia article. Not tested but looks good.

And yes, translators can also insert instructions to `format c:` or `delete system32`, but we'll probably catch such translations soon-ish.

PR #445, opened 2015-12-26
Fix CSS filter being more generous than it should be in 1471-pre

toad:
next is too generous with CSS selectors in nav-*. I doubt that this is exploitable but we should not allow more than the spec unless there is a good reason. Also, we don't sanitize/rebuild the selectors here (which we do in other cases), which might make any possible exploit easier.

To be clear, this should be considered for 1471 since it's a possible (but unlikely) security regression in next/pre-1471 relative to 1470.


toad:
The last commit fixes another issue where the filter is more permissive than it should be.

PR #446, opened 2015-12-26
CSS parser expression fixes

toad:
This goes a bit further than #445. It documents the syntax and limitations and also fixes some potentially confusing bugs in future parsing rules. It is not urgent. CSS tests pass.


unixninja92:
Inconsistant indentation

PR #421, opened 2015-11-11
L10n tests and better error handling

vwoodzell:
Added some unit tests for l10n. Also changed addL10nSubstitutionInner() to address some problems like the one in bug 6386. Added a test that validates the l10n files, and fixed problems revealed by that test.

PR #447, opened 2015-12-27
Setup Travis at https://travis-ci.org/freenet/fred/

nextgens:
Setup Travis at https://travis-ci.org/freenet/fred/

... It highlights that #437 uses Java7-only Integer.compare(int, int); thanks @Bombe  for noticing!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: java
+
+cache:
+  apt: true
+  directories:
+  - lib/
+
+before_install:
+  - sudo hostname freenet-build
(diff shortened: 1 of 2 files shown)


nextgens:
If we merge it we can get rid of the travis branch @Thynix has started

Thynix:
Thanks! The l10n tests were already failing on Java 6 too. It seems like some other thing changes in the string API between 6 and 7. I've posted about it [here](https://github.com/freenet/fred/pull/421#issuecomment-167704194).

PR #448, opened 2015-12-28
CSS: more 1471 fixes

toad:
A few minor fixes to the CSS changes. Removes some keywords that shouldn't be allowed, adds some that should. Can wait.

### DIFF
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -411,7 +411,7 @@ class CSSTokenizerFilter {
 		// auto | stretch | <baseline-position> | [ <item-position> && <overflow-position>? ] | [ legacy && [ left | right | center ] ]
 		auxilaryVerifiers[128] = new CSSPropertyVerifier(Arrays.asList("legacy"), null, null, null, true);
 		auxilaryVerifiers[129] = new CSSPropertyVerifier(null,ElementInfo.VISUALMEDIA,null,Arrays.asList("4 128", "128 4"), true,true);
-		auxilaryVerifiers[130] = new CSSPropertyVerifier(null,ElementInfo.VISUALMEDIA,null,Arrays.asList("129 126", "126", "126 129"), true,true);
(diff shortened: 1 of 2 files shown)


Thynix:
What does it add? I only see removing "space-between", "space-around", and "space-evenly" from "align-items". What does the "justify-items" change fix?

Thynix:
It removes left, center, and right - the strings are references to other array indexes.

Thynix:
Er, I'm wrong. I still don't understand what exactly it changes.

Thynix:
I think I got it now?

For <justify-items> it used to accept "left legacy", "center legacy", "right legacy", "legacy left", "legacy center", or "legacy right" before or after "center", "start", "end", "self-start", "self-end", "flex-start", "flex-end", "left", or "right". Instead of the former set it now only accepts "true" or "safe".

I still don't understand why though.

Thynix:
Why change `justify-items`? [The spec](http://www.w3.org/TR/css3-align/#justify-items-property) indicates the previous values were accepted, does it not?

toad:
align-items doesn't accept the 3 keywords I deleted.

Thynix:
Okay, but what about `justify-items`?

toad:
justify-items accepts any one of the auxiliary verifiers 125, 129 and 130.

125 is auto, stretch or baseline-position.
129 is the last bit: legacy && [left|right|center]: legacy left, right legacy, center etc.
- This is made of 128 (legacy) and 4 (left|right|center)
That leaves 130 which should be "item-position && overflow-position?" e.g. "start true" etc.

The new code has 130 = combinations of 122 and 126. 122 is overflow-position, 126 is item-position.
The old code had 130 = combinations of 129 and 126. This is incorrect - we don't want the legacy keyword here, and there are more item-positions than left, right, center.

Thynix:
Alright, so. We finally got me to understand this:

`<justify-items>` used to accept `<item-position>` along with `legacy && [ left | right | center ]`, which is incorrect. It now accepts `<item-position>` along with `<overflow-position>`, to match http://www.w3.org/TR/css3-align/#justify-items-property

PR #450, opened 2015-12-29
FNP fixes post 440

toad:
Fixes a bug nextgens missed in #440 and cleans up slightly. 

**Please consider for 1471** because the bug affected announcement performance: Checking for peers we had already announced to failed because we were putting ECDSA pubkey hashes into the exclude set, but then comparing them against obfuscation keys (PeerNode.identity).


nextgens:
Nice catch! Agreed, that should be in 1471 too

PR #452, opened 2015-12-30
Throttle unmatched scan

toad:
This needs more testing and probably more work. However review would be helpful.
See https://bugs.freenetproject.org/view.php?id=6773

### DIFF
--- a/src/freenet/node/FNPPacketMangler.java
+++ b/src/freenet/node/FNPPacketMangler.java
@@ -221,8 +221,15 @@ public class FNPPacketMangler implements OutgoingPacketMangler {
 				}
 			}
 		}
-		PeerNode[] peers = crypto.getPeerNodes();
+
 		if(node.isStopping()) return DECODED.SHUTTING_DOWN;
+		
+		if(wantAnonAuth) {
+		    if(tryProcessAuthAnon(buf, offset, length, peer))
+		        return DECODED.DECODED;
+		}
+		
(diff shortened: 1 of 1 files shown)


src/freenet/node/FNPPacketMangler.java
```diff
@@ -150,6 +145,28 @@ public void run() {
 
         private long lastConnectivityStatusUpdate;
         private Status lastConnectivityStatus;
+        private AtomicInteger unmatchedCount = new AtomicInteger(0);
+        // FIXME make configurable.
+        static final int MAX_UNMATCHED_PACKETS_IN_INTERVAL = 10;
+        private int CLEAR_UNMATCHED_INTERVAL = (int) MINUTES.toMillis(5);
```
bertm: This interval should be way smaller, and the limit way higher. I receive about 20 unmatched packets each minute on a moderately connected peer. We only need to throttle when we're under attack; presumably in such a way that our performance would severely suffer.

Receiving a up till some tens of unmatched packets each second is handled just fine (just a minor nuisance). Receiving 1000 unmatched packets per second for a few a second, followed by several minutes of no unmatched packets, still makes us largely unreachable for those seconds, possibly causing backoff.

nextgens:
This is something I'd like to see in 1475

ArneBab:
@nextgens what’s missing too get this done?

PR #453, opened 2015-12-30
Largish changes to in-RAM temp buckets

toad:
Major changes here:
- Track last access to any temporary bucket/RAF.
- Always migrate buckets to disk [configurable period] after last access.
- Use less memory for temp buckets by default.

(On next we migrate buckets to disk in order of creation when we run out of disk space)

For consideration. Maybe we want to keep only some of this. This needs testing - porting it forward to next was fairly involved. It should improve performance on average new nodes; I believe I tested it quite a bit at the time, but there have been significant changes since then. There may be lock contention issues, or even full blown deadlocks etc; this is for discussion, not for immediate inclusion.


nextgens:
I'm not sure I get why it's better than the current behaviour. Why would we want to always migrate stuff to disk?

IMHO if there's no memory pressure it's just wasted disk IO

PR #455, opened 2015-12-30
Simplify disconnected()

toad:
Simplifies disconnection code and fixes minor concurrency issues. Not urgent. Old.


unixninja92:
Looks good to me

src/freenet/node/NodeStats.java
```diff
@@ -473,10 +473,16 @@ public void set(Long val) throws InvalidConfigValueException, NodeNeedRestartExc
...
+				if(newLevel == NETWORK_THREAT_LEVEL.MAXIMUM) {
+					synchronized(NodeStats.this) {
+						ignoreLocalVsRemoteBandwidthLiability = true;
+					}
+				}
+				if(oldLevel == NETWORK_THREAT_LEVEL.MAXIMUM) {
+					synchronized(NodeStats.this) {
+						ignoreLocalVsRemoteBandwidthLiability = false;
+					}
+				}
```
bertm: While correct as long as `oldLevel != MAXIMUM || newLevel != MAXIMUM`, it feels weird to first update based on the new value, and then on the old value.

src/freenet/node/NodeStats.java
```diff
@@ -1459,6 +1465,7 @@ private String checkBandwidthLiability(double bandwidthAvailableOutputUpperLimit
 			PeerNode source, boolean isLocal, boolean isSSK, boolean isInsert, boolean isOfferReply, boolean hasInStore, int transfersPerInsert, boolean realTimeFlag, int maxOutputTransfers, int maxOutputTransfersPeerLimit, UIDTag tag) {
 		String name = input ? "Input" : "Output";
 		int peers = node.peers.countConnectedPeers();
+		boolean ignoreLocalVsRemoteBandwidthLiability = ignoreLocalVsRemoteBandwidthLiability();
```
bertm: This duplication of names is confusing. Prepending `current` or something similar to the name would make its intended purpose more clear, and would allow the compiler to catch cases where one meant `ignoreLocalVsRemoteBandwidthLiability()` but wrote `ignoreLocalVsRemoteBandwidthLiability` (which are both of type `boolean`) or vice-versa.

src/freenet/node/NodeStats.java
```diff
@@ -3890,6 +3897,7 @@ public final double getBandwidthLiabilityUsage() {
 		long now = System.currentTimeMillis();
 		long limit = getLimitSeconds(false);
 		int transfersPerInsert = outwardTransfersPerInsert();
+		boolean ignoreLocalVsRemoteBandwidthLiability = ignoreLocalVsRemoteBandwidthLiability();
```
bertm: As above, this is confusing.

bertm:
Apart from the style issues mentioned above, this looks good to me.

src/freenet/node/NodeStats.java
Bombe: Also, whitespace after `if` and `synchronized`, please, as these are language constructs and not method calls.

PR #457, opened 2015-12-30
Cleanups around the time of 1471 (low priority)

toad:
Minor stuff, consider for 1472.

### DIFF
--- a/src/freenet/node/PeerNode.java
+++ b/src/freenet/node/PeerNode.java
@@ -13,7 +13,6 @@ import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
 import java.lang.ref.WeakReference;
-import java.math.BigInteger;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
 import java.net.UnknownHostException;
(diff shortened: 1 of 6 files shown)


unixninja92:
Looks good to me

src/freenet/node/PeerNode.java
```diff
@@ -5828,4 +5828,9 @@ private boolean verifyReferenceSignature(SimpleFieldSet fs) throws ReferenceSign
 	protected final byte[] getPubKeyHash() {
 	    return peerECDSAPubKeyHash;
 	}
+
+	// FIXME Move callers into freenet/node/ (or copy the byte[]).
+    public byte[] getPeerECDSAPubKeyHash() {
+        return peerECDSAPubKeyHash;
```
bertm: I don't see a need to add this as another FIXME, please fix it instead (by copying the `byte[]` and adding Javadoc, that is). It's not used frequently anyways.

src/freenet/node/PeerManager.java
```diff
@@ -252,6 +252,7 @@ private boolean readPeers(File peersFile, NodeCrypto crypto, OpennetManager open
 				// Read a single NodePeer
 				SimpleFieldSet fs;
 				fs = new SimpleFieldSet(br, false, true);
+				boolean broken = true;
```
bertm: This is never set to `false`. Did you intend `broken = false;` somewhere in the `try {}`?

bertm:
Yay for making the pubkey hashes package-private. Looks good except for the above comments and the terrible mess of whitespace in the parts touched.

Thynix:
@toad Do you plan to address the comments? If not someone else can volunteer to do so.

PR #458, opened 2015-12-30
Minor FNP cleanups

toad:
Mostly deleting obsolete code and updating comments.


nextgens:
I'd land that on 1471 too

PR #460, opened 2015-12-31
Plugin manager clean up

Bombe:
Lots of clean ups and refactorings in one of `PluginManager`’s most ugly methods.

### DIFF
--- a/src/freenet/pluginmanager/PluginManager.java
+++ b/src/freenet/pluginmanager/PluginManager.java
@@ -71,16 +71,7 @@ import freenet.support.io.NativeThread;
 
 public class PluginManager {
 
-	/*
-	 *
-	 * TODO: Synchronize
-	 * TODO: Synchronize
-	 * TODO: Synchronize
-	 * TODO: Synchronize
-	 * TODO: Synchronize
-	 *
-	 */
-	private final HashMap<String, FredPlugin> toadletList;
+	private final HashMap<String, FredPlugin> toadletList = new HashMap<String, FredPlugin>();
 
 	/* All currently starting plugins. */
 	private final OfficialPlugins officialPlugins = new OfficialPlugins();
@@ -111,7 +102,6 @@ public class PluginManager {
 		logDEBUG = Logger.shouldLog(LogLevel.DEBUG, this);
 		// config
 
-		toadletList = new HashMap<String, FredPlugin>();
 		pluginWrappers = new ArrayList<PluginInfoWrapper>();
 		pluginsFailedLoad = new HashMap<String, PluginLoadFailedUserAlert>();
 		this.node = node;
--- a/src/freenet/pluginmanager/PluginManager.java
+++ b/src/freenet/pluginmanager/PluginManager.java
@@ -54,6 +54,7 @@ import freenet.node.RequestStarter;
 import freenet.node.useralerts.AbstractUserAlert;
 import freenet.node.useralerts.UserAlert;
 import freenet.pluginmanager.OfficialPlugins.OfficialPluginDescription;
+import freenet.pluginmanager.PluginManager.PluginProgress.ProgressState;
 import freenet.support.HTMLNode;
 import freenet.support.HexUtil;
 import freenet.support.JarClassLoader;
@@ -431,7 +432,7 @@ public class PluginManager {
 		PluginInfoWrapper pi = null;
 		try {
 			plug = loadPlugin(pdl, filename, pluginProgress, alwaysDownload);
-			pluginProgress.setProgress(PluginProgress.PROGRESS_STATE.STARTING);
+			pluginProgress.setProgress(ProgressState.STARTING);
 			pi = new PluginInfoWrapper(node, plug, filename, pdl.isOfficialPluginLoader());
 			PluginHandler.startPlugin(PluginManager.this, pi);
 			synchronized (pluginWrappers) {
@@ -1539,7 +1540,7 @@ public class PluginManager {
 	 */
 	public static class PluginProgress {
 
-		enum PROGRESS_STATE {
+		enum ProgressState {
 			DOWNLOADING,
 			STARTING
 		}
@@ -1547,7 +1548,7 @@ public class PluginManager {
 		/** The starting time. */
 		private long startingTime = System.currentTimeMillis();
 		/** The current state. */
-		private PROGRESS_STATE pluginProgress;
+		private ProgressState pluginProgress;
 		/** The name by which the plugin is loaded. */
 		private String name;
 		/** Total. Might be bytes, might be blocks. */
@@ -1571,7 +1572,7 @@ public class PluginManager {
 		 */
 		PluginProgress(String name, PluginDownLoader<?> pdl) {
 			this.name = name;
-			pluginProgress = PROGRESS_STATE.DOWNLOADING;
+			pluginProgress = ProgressState.DOWNLOADING;
 			loader = pdl;
 		}
 
@@ -1604,7 +1605,7 @@ public class PluginManager {
 		 *
 		 * @return The current state of the plugin
 		 */
-		public PROGRESS_STATE getProgress() {
+		public ProgressState getProgress() {
 			return pluginProgress;
 		}
 
@@ -1614,13 +1615,13 @@ public class PluginManager {
 		 * @param pluginProgress
 		 *            The current state
 		 */
-		void setProgress(PROGRESS_STATE state) {
+		void setProgress(ProgressState state) {
 			this.pluginProgress = state;
 		}
 
 		/**
-		 * If this object is one of the constants {@link #DOWNLOADING} or
-		 * {@link #STARTING}, the name of those constants will be returned,
+		 * If this object is one of the constants {@link ProgressState#DOWNLOADING} or
+		 * {@link ProgressState#STARTING}, the name of those constants will be returned,
 		 * otherwise a textual representation of the plugin progress is
 		 * returned.
 		 *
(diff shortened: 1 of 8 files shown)


src/freenet/pluginmanager/PluginManager.java
```diff
@@ -324,12 +308,14 @@ private static String pluginList(ArrayList<PluginInfoWrapper> wrappers) {
 	}
 
 	private String[] getConfigLoadString() {
-		ArrayList<String> v = new ArrayList<String>();
+		List<String> v = new ArrayList<String>();
+		for (PluginInfoWrapper pi : pluginWrappers) {
```
unixninja92: Why did you move this before the if(!started) check? I understand why you moved it out of the synchronized block, but if it's not started, why do you want to bother filling in v? Might that have some effect on the timing of wether started is true or not?

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1064,65 +975,35 @@ public String handleHTTPPost(String plugin, HTTPRequest request) throws PluginHT
...
-		boolean found = false;
-		synchronized(pluginWrappers) {
-			for(int i = 0; i < pluginWrappers.size() && !found; i++) {
-				pi = pluginWrappers.get(i);
-				if(pi.getThreadName().equals(name)) {
-					found = true;
-					break;
-				}
+		for (PluginInfoWrapper pluginInfoWrapper : pluginWrappers) {
+			if (pluginInfoWrapper.getThreadName().equals(name)) {
```
unixninja92: Why did you remove the break after stopping the plugin? I don't see why we should keep looking through threads.

unixninja92: Same question for the other killPlugin methods

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1183,266 +1064,246 @@ public File getPluginFilename(String pluginName) {
...
 	 * @throws PluginNotFoundException
 	 *             If anything goes wrong.
+	 * @throws PluginAlreadyLoaded if the plugin is already loaded
 	 */
-	private FredPlugin loadPlugin(PluginDownLoader<?> pdl, String name, PluginProgress progress, boolean ignoreOld) throws PluginNotFoundException {
+	private FredPlugin loadPlugin(PluginDownLoader<?> pdl, String name, PluginProgress progress, boolean alwaysDownload) throws PluginNotFoundException, PluginAlreadyLoaded {
 
 		pdl.setSource(name);
 
-		/* check for plugin directory. */
```
unixninja92: Why are we removing this check?

unixninja92:
@Thynix reviewed up to ```private FredPlugin loadPlugin```

src/freenet/pluginmanager/PluginManager.java
Bombe: Mostly because if I didn’t I would either have to let it stay in the `synchronized` block or would have to split up the `synchronized` block into two `synchronized` blocks… I’m not sure which solution is better. :)

src/freenet/pluginmanager/PluginManager.java
Bombe: Hmm, no reason, really. Re-added.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -280,41 +269,36 @@ public void stop(long maxWaitTime) {
 			progress.kill();
 		}
 		// Stop already loaded plugins.
-		ArrayList<PluginInfoWrapper> wrappers;
-		synchronized(pluginWrappers) {
-			wrappers = new ArrayList<PluginInfoWrapper>(pluginWrappers);
-		}
-		for(PluginInfoWrapper pi : wrappers)
+		for(PluginInfoWrapper pi : pluginWrappers)
```
bertm: This should have braces.

src/freenet/pluginmanager/PluginManager.java
bertm: Doing it this way has the potential (however unlikely) to return a wrong result due to a race condition where the snapshots of `pluginWrappers` and `pluginsFailedLoad`, not taken atomically as a pair, are inconsistent.

Including a little extra work in the `synchronized` block is not that bad, we don't expect this method to be called too often anyhow.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -409,17 +395,17 @@ private PluginInfoWrapper realStartPlugin(final PluginDownLoader<?> pdl, final S
...
-			if (plug == null)
-				return null; // Already loaded
-			pluginProgress.setProgress(PluginProgress.PROGRESS_STATE.STARTING);
+			plug = loadPlugin(pdl, filename, pluginProgress, alwaysDownload);
+			pluginProgress.setProgress(ProgressState.STARTING);
 			pi = new PluginInfoWrapper(node, plug, filename, pdl.isOfficialPluginLoader());
 			PluginHandler.startPlugin(PluginManager.this, pi);
+			pluginWrappers.add(pi);
 			synchronized (pluginWrappers) {
-				pluginWrappers.add(pi);
```
bertm: This should remain in the `synchronized` block, for `pluginWrappers` and `pluginsFailedLoad` to remain consistent.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1183,266 +1068,246 @@ public File getPluginFilename(String pluginName) {
...
 		/* check if file needs to be downloaded. */
 		if(logMINOR)
 			Logger.minor(this, "plugin file " + pluginFile.getAbsolutePath() + " exists: " + pluginFile.exists()+" downloader "+pdl+" name "+name);
 		int RETRIES = 5;
-		for(int i = 0; i < RETRIES; i++) {
-			if(!pluginFile.exists() || pluginFile.length() == 0)
+		for (int i = 0; i < RETRIES; i++) {
+			if (!pluginFile.exists() || pluginFile.length() == 0) {
 				try {
 					downloaded = true;
```
bertm: Why is this flag set here? It's clearly not downloaded at this point and the flag will never be reset to `false`.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1183,266 +1068,246 @@ public File getPluginFilename(String pluginName) {
...
+	private void verifyDigest(PluginDownLoader<?> pluginDownLoader, File pluginFile) throws PluginNotFoundException {
+		String digest = pluginDownLoader.getSHA1sum();
+		if (digest == null) {
+			return;
+		}
+		String testsum = getFileDigest(pluginFile, "SHA-1");
+		if (!(digest.equalsIgnoreCase(testsum))) {
+            Logger.error(this, "Checksum verification failed, should be " + digest + " but was " + testsum);
+            throw new PluginNotFoundException("Checksum verification failed, should be " + digest + " but was " + testsum);
+        }
```
bertm: This whitespace is inconsistent.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -820,66 +806,16 @@ public void unregisterPluginToadlet(PluginInfoWrapper pi) {
 		}
 	}
 
-	public void addToadletSymlinks(PluginInfoWrapper pi) {
```
bertm: Please deprecate this first, so we won't break plugins unnecessarily.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -820,66 +806,16 @@ public void unregisterPluginToadlet(PluginInfoWrapper pi) {
...
-					Logger.normal(this, "Removed HTTP symlink: " + target +
-						" => /plugins/" + pi.getPluginClassName() + '/');
-				}
-			} catch(Throwable ex) {
-				Logger.error(this, "removing Toadlet-link", ex);
-			}
-		}
-	}
-
-	public void removeToadletSymlinks(PluginInfoWrapper pi) {
```
bertm: As above, please deprecate this first.

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1706,39 +1559,35 @@ public static void setLanguage(LANGUAGE lang) {
...
+						} catch (Throwable t) {
+							Logger.error(this, "Cought Trowable in Callback", t);
+						}
+					}
+				}, "Callback");
 			}
 		}
 	}
 
-	public THEME getFProxyTheme() {
```
bertm: Please deprecate this before removing it entirely.

src/freenet/pluginmanager/PluginManager.java
Bombe: It should. It will.

src/freenet/pluginmanager/PluginManager.java
Bombe: If they need to be synchronized together, maybe they should be their own class.

src/freenet/pluginmanager/PluginManager.java
Bombe: The point of this flag is not to shows that the plugin has already been successfully downloaded, it’s for showing that an attempt to download was made. On the first run of the loop the plugin may already exist; errors on loading will then trigger a re-download. Only if this run (with `i = 1`) fails, an exception is thrown.

So, yes, the flag’s name is stupid, and it took me a while to understand. And yes, its name should be better. Will fix.

src/freenet/pluginmanager/PluginManager.java
Bombe: It indeed is. Will fix.

src/freenet/pluginmanager/PluginManager.java
Bombe: Okay, though I think that plugins that ask the plugin manager for the current fproxy theme should be taken out back and shot. Should we schedule a date for removal? In six months, maybe? A year?

src/freenet/pluginmanager/PluginManager.java
Bombe: I did a search with grep in my work directory and even though I don’t claim to have all plugins’ source code I got no hits in anywhere but the plugin manager. I will deprecate them but I would like to schedule date for removal.

src/freenet/pluginmanager/PluginManager.java
Bombe: ACK.

src/freenet/pluginmanager/PluginManager.java
Bombe: Actually, take a look at this method: it does not even pretend to do what it’s named for. I’m more in favor of simply removing this method before anybody notices its complete brokenness. :)

bertm:
Given the state of the methods now deprecated, I suggest we remove them entirely in 1473.

bertm:
As I see it, `startingPlugins`, `pluginWrappers` and `pluginsFailedLoad` should be kept consistent. Mostly for presentation purposes: if we fail doing so, the downside would be the possibility of a condition where the user would temporarily be presented with incorrect information on the plugin state. There are a couple more issues with synchronization (such as access to `startingPlugins` being synchronized on either `startingPlugins`, `this` or not at all) that will become apparent when doing so.

This pull request does not appear to make things worse than they were before. So if you'd rather not do that at this time, it can be done in a separate pull request.

src/freenet/pluginmanager/PluginManager.java
bertm: Yes, it's a great example of copy-pasting gone wrong ;)

Bombe:
@bertm I agree on keeping those three together. I am already working on something for that which should be finished in a couple of hours, hopefully. Also, I am completely fine with removing all those now-deprecated methods in 1473.

bertm:
Apparently I missed the additional progress made.

This looks good to me so far, but  we have merge conflicts now. @Bombe will you resolve those, or should I do it?

xor-freenet:
@Bombe It seems bertm has ACKed this a long time ago and it merely lacks merging of next into it. Can you do that?

ArneBab:
@Bombe this has been marked as "after next release" for quite some time. I’d say with 1483 having only few regressions compared to 1480, this qualifies as after next release. Once you get to fix the conflicts, I’d like to merge this.

Bombe:
> @Bombe this has been marked as "after next release" for quite some time. I’d say with 1483 having only few regressions compared to 1480, this qualifies as after next release. Once you get to fix the conflicts, I’d like to merge this.

I had some more commits lying around in that branch with some more minor fixes; they all have been rebased on `next` so should be good to go.

ArneBab:
Cool! Thank you!

PR #461, opened 2015-12-31
Deterministic ECDSA

nextgens:
Upgrade to BC1.54 and make deterministic (RFC6979) ECDSA signatures

It's definitely safer and probably faster too

This should probably wait *after* 1471 is out.


toad:
Code looks good, but verify the hashes.

Security is improved if:
- We don't care about side channels (it reuses key material!).
- The HMAC input to the signature is sufficiently strong.
- We have doubts about our PRNG.

So I guess I'm a bit sceptical here. This improves things if we had strong randomness at the time of generation but we're not sure at time of signing. But it introduces additional side-channels against the private key. Which may be OK if we're sure our HMAC doesn't have side-channels (but this seems highly unlikely as it is implemented in Java with no hardware support). See section 3.6 and 5.

Making it deterministic will make it produce the same SSK if the same content is inserted twice. This is a good thing, although should be rare in practice.

toad:
ECDDSA itself introduces a timing side channel. See the loop in section 3.2. However, this is probably only exploitable if the HMAC is broken.

nextgens:
530e418 diverges from the RFC... in any case, the property we're after here is "ensuring that for a given key and message, we never do two distinct signatures with the same K"

Failing to do so (bad randomness with the old code, side-channels with the new one) would expose the private key

Assuming that I didn't screw it up, we do deterministic signatures (avoiding the problem altogether)

nextgens:
In any case, that legacy DSA API is terrible; We take BigIntegers as input; it will never be safe to use.

nextgens:
@toad it only loops if the first hit isn't in range; the old code loops too in that case...

FYI I haven't implemented the looping stuff in the legacy-DSA stuff (we mask one bit) so we shouldn't have to

unixninja92:
Looks good to me

nextgens:
@toad in fact there's a very good reason on why we shouldn't add state (the node-shared key you suggested we mix in) to our signatures

We don't want two signatures to be "related" even post-key compromise. If you insert two SSKs they should share as little state as possible for obvious reasons (attribution, FS, ...)

toad:
Is there a proof that this only introduces new side-channels if the hash function is broken?

nextgens:
HMAC has a security proof dependant on the security proof of the underlying hash function. In our case that's SHA256 here... that doesn't have a security proof (and never will); but if that's broken we have other problems (what we sign has been digested by it too!).

In any case I don't think that what we should be worried about here is side channels. We don't let an attacker force us to sign (no oracle he can directly control; we pick the private key).

Verification is another matter altogether; but I haven't changed any of it.

Keep in mind that the value K has to be kept secret (whether it's random or not) and is never communicated to the attacker. Which side channel are you worried about? Timing attacks? Something else?

Since e76aa0d52a5d79a1be84a94b5886fcd94866f034 we're doing an extra round of hashing.

PR #463, opened 2015-12-31
test: fix resource path on Java 6

Thynix:
On Java 6 the classloader path may return multiple colon-delimited paths, which makes using the classloader path not work as intended. Using the class path requires more backing out of the build directory.

In my testing this allows the tests from #421 to pass under Java 6, but I really hope there's a more idiomatic / less fragile way to do this.

### DIFF
--- a/test/freenet/l10n/BaseL10nTest.java
+++ b/test/freenet/l10n/BaseL10nTest.java
@@ -215,9 +215,9 @@ public class BaseL10nTest extends TestCase {
     
     private BaseL10n createTestL10n(LANGUAGE lang) {
         String testL10nPath = new File(TEST_PATH, L10N_PATH).getPath();
-        URL classLoaderUrl = getClass().getClassLoader().getResource(".");
+        URL classLoaderUrl = getClass().getResource(".");
         File classLoaderDir = new File(classLoaderUrl.getPath());
-        File overrideFile = new File(new File(classLoaderDir, testL10nPath),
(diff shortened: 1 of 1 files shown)


Thynix:
I'd prefer to add the `../../../` to `TEST_PATH`, but for reasons I don't understand although the constant is not used elsewhere it's somehow used as a resource path, and so causes test failure if it's changed.

Thynix:
I'll merge this for 1471 and revert it in 1472.

toad:
This is a nasty hack because it climbs out of build/main/ and up to test/. In other words it reads it from the source code, because for some reason it's not being copied.

I've investigated in a bit more detail. We should be able to use the classloaders directly without the need for such hacks. Unfortunately we can't because even for BaseL10nTest, the src/ dir is on the classpath BEFORE the test/ dir. So even if we use the filename to load the test/ l10n, we end up with the wrong file. The solution is to simply rename the files.

Bombe:
@toad the _real_ solution would be to do it correctly, according to the 15-year-old standard of storing source code and resources separately, in `src/main/java` and `src/main/resources`. And as the _source path_ does not ever belong anywhere close to the classpath, the problem would never have occured in the first place. Please stop piling hacks onto hacks.

Thynix:
Yep, @Bombe's point is the way to go. Making that change will simplify the
Gradle setup as well. Merging follows renames, right? I'd be concerned
about making bitrot worse.

PR #465, opened 2015-12-31
Fix l10n tests properly

toad:
Better solution. Improvement on #463


test/freenet/l10n/BaseL10nTest.java
```diff
@@ -207,19 +205,19 @@ public void testStrings() throws Exception {
...
     
+    static final String FILENAME = "test.l10n.${lang}.properties";
+    static final String OVERRIDE_FILENAME = "test.l10n.${lang}.override.properties";
+    
     private BaseL10n createL10n(LANGUAGE lang) {
-        File overrideFile = new File(L10N_PATH, "freenet.l10n.${lang}.override.properties");
-        return new BaseL10n(L10N_PATH, "freenet.l10n.${lang}.properties",
-                overrideFile.getPath(), lang);
+        return new BaseL10n("freenet/l10n/", "freenet.l10n.${lang}.properties",
+                "freenet.l10n.${lang}.override.properties", lang);
```
Thynix: Why doesn't this use the constants? (The removed `L10N_PATH` would be useful here too.)

Thynix:
What are the advantages of this approach?

Thynix:
Given that what we currently have functions, and we no longer need to support Java 6 on next, I don't see a need for this.

PR #467, opened 2016-01-01
Replace our HMAC implementation by whatever JCA provides

nextgens:
Odds are it won't make the code any faster... but it's more readable and might be better equipped than us to deal with side channels.

It's a lot less ambitious than #258


nextgens:
It's worth noting that there is a few differences in between the old and the new HMAC:
- the new implementation won't stretch the key; it will throw if the key isn't of the expected length. I might be convinced to replace the check with another ensuring that the keylength is "big enough" and "not default/zeros"
- the new implementation dropped support for output truncation; we were not using it anywhere and it simplifies the API

PR #470, opened 2016-01-01
Commit my configuration file for IDEA

nextgens:
A very basic config file ... ensuring it "just works" when you check it out


Bombe:
If we would use a real build system, IDEA integration would basically be free and we wouldn’t have to resort to shit that will break every other developer’s development environment.

Thynix:
Yep. I'm inclined to wait for Gradle instead.

PR #472, opened 2016-01-01
remove legacy updater for build 1421

Thynix:
Build 1421 was released in October of 2012, and build 1431 has been
mandatory since February 2013. This is old enough it can be reasonably
removed.


Thynix:
If I understood Matthew correctly most of the changes to NodeUpdateManager and the change to MainJarUpdater should be reverted, and the sections updated to serve 1471 signed with the old key. The resulting UOM should be tested to work with {windows, linux} * {old to new, new to new} * {ordinary fetch, UOM}.

PR #474, opened 2016-01-02
Mark all FNP traffic with Diffserv AF13

nextgens:
We aim for lowest priority, highest throughput possible here. I'm not sure on whether we should make it configurable or not


nextgens:
I've tested it, it works on linux with both openjdk and oracle; For windows we need to set some magic registry keys...

To allow "applications" to set the bits:
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\Tcpip\Parameters] "DisableUserTOSSetting"=dword:00000000

In case the user doesn't have windows pro or isn't joined to a domain he'll also need:
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\Tcpip\QoS] "Do not use NLA"="1"

Thynix:
What is its behavior on Windows without the registry keys? Does it not set the priority class and otherwise function normally, or does it prevent or interfere with normal operation?

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1143,6 +1143,8 @@ Node.probeUptimeShort=Respond to probe requests for uptime
 Node.probeUptimeLong=If true freenet will respond to probe requests for 48-hour or 7-day uptime percentage. This is to improve knowledge of network conditions, and allow simulations to better reflect the actual network. It will add a small amount of random noise to make the information less identifiable but still useful for diagnostics. If false it will respond with a refusal.
 Node.probeRejectStatsShort=Respond to probe requests for proportion of requests of various kinds rejected.
 Node.probeRejectStatsLong=If true, Freenet will respond to probe requests for the proportion of various kinds of requests which are being rejected.
+Node.trafficClass=The TrafficClass value to pass to Socket.setTrafficClass() for both darknet and opennet
```
unixninja92: Do you want a period at the end here?

nextgens: Done in  d5f1ce9

nextgens:
Without the registry keys it will revert back to the previous behaviour; outgoing packets won't be marked (no ill-effects).

nextgens:
One valid concern that has been raised is that it might allow Eve to fingerprint our traffic more easily.

I think that the benefits out-weight the risks here... and to be fair, most routers/ISPs will reset the mark at the border (after the upstream congestion point), making it a non-problem.

bertm:
How about making this non-configurable, but disabling it on maximum network security?

nextgens:
I think that there is value in having it configurable; what matters to me is the default. I've seen users uninstall because freenet was preventing them from doing VOIP/gaming... this might fix it

As for disabling it at higher security levels... I don't care :)

src/freenet/node/Node.java
```diff
@@ -1312,13 +1320,37 @@ public Short get() {
 
 					@Override
 					public void set(Short val) throws InvalidConfigValueException {
-						if(maxHTL < 0) throw new InvalidConfigValueException("Impossible max HTL");
+						if(val < 0) throw new InvalidConfigValueException("Impossible max HTL");
```
ArneBab: how did that bug survive tests? good catch!

xor-freenet: Seems like this should be in a different PR?

unixninja92: @xor-freenet I believe @nextgens and @Thynix were discussing it earlier and decided it wasn't worth the effort to separate it out. IIRC it was on IRC.

Thynix: Correct.

src/freenet/io/comm/TrafficClass.java
```diff
@@ -0,0 +1,60 @@
...
+  CS6(0xC0),
+  CS7(0xE0);
+
+  public final int value;
+  TrafficClass(int tc) {
+    value = tc;
+  }
+
+  public static TrafficClass getDefault()
+  {
```
Bombe: Braces are inconsistent.

src/freenet/io/comm/TrafficClass.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public static TrafficClass fromNameOrValue(String tcName) {
+    int tcParsed = -1;
+    try {
+      tcParsed = Integer.parseInt(tcName);
+    } catch (NumberFormatException e){
+      // it's fine; we might have a name here
+    }
+
+    for(TrafficClass t : TrafficClass.values()) {
```
Bombe: Whitespace between language constructs (`if`, `for`) and opening parens, please.

src/freenet/node/Node.java
Bombe: @ArneBab tests? You’re cute. :)

src/freenet/io/comm/TrafficClass.java
```diff
@@ -0,0 +1,60 @@
...
+      if(t.toString().equalsIgnoreCase(tcName) ||
+         t.value == tcParsed)
+      {
+        return t;
+      }
+    }
+    return getDefault();
```
Bombe: I’d prefer to throw an `IllegalArgumentException` here…

src/freenet/io/comm/TrafficClass.java
nextgens: fixed in 3e75d835d98ce2470489a97c37c63983ef78ebc5

src/freenet/io/comm/TrafficClass.java
nextgens: that's d115b1e1f07fcf83224f633c6abb79c8f82b4a11

src/freenet/io/comm/TrafficClass.java
nextgens: done in f25f9a22590c56ff383836130447bcb50faaf397

src/freenet/node/Node.java
```diff
@@ -1312,13 +1320,47 @@ public Short get() {
...
+					 trafficClass = TrafficClass.fromNameOrValue(tcName);
+				 } catch (IllegalArgumentException e) {
+					 throw new InvalidConfigValueException(e);
+				 }
+			 }
+
+			 @Override
+			 public String[] getPossibleValues() {
+				 ArrayList<String> array = new ArrayList<String>();
+				 for(TrafficClass tc : TrafficClass.values())
```
Bombe: Sorry for the nitpickery but… whitespace after `for`? :)

nextgens: there has to be a better way to do this... a git-pre-commit hook? a stylecheck thingy that travis would run?

hernic: Checkstyle could help on this

Le sam. 9 janv. 2016 10:45, Florent Daigniere <notifications@github.com> a
écrit :

nextgens:
I think that this is ready to make it to next

src/freenet/io/comm/UdpSocketHandler.java
```diff
@@ -64,6 +64,7 @@ public UdpSocketHandler(int listenPort, InetAddress bindto, Node node, long star
 		try {
 			// Exit reasonably quickly
 			_sock.setReuseAddress(true);
+			_sock.setTrafficClass(node.getTrafficClass().value);
```
bertm: From the `setTrafficClass` Javadoc:
> Setting bits in the precedence field [the 3 most significant bits, red.] may result in a `SocketException` indicating that the operation is not permitted. 

Failing to set the traffic class is not a fatal error, but failure can be expected on some systems because of the above. This instruction should have its own try/catch that at most logs a warning if setting the traffic class fails, but should certainly not throw a fatal `RuntimeException` as per the current try/catch.

nextgens: Taken care of in 0619f04

bertm:
Setting a new traffic class through `TrafficClassCallback`:

* does not instruct the `UdpSocketHandler` to reset the traffic class on the socket,
* nor does it throw an appropriate `NodeNeedRestartException` to indicate that this is indeed not done.

The callback should do either one of both, but not none.

bertm:
This looks good to me.

toad:
Does this still crash routers? I guess not because it has to be turned on manually in the OS?

nextgens:
I don't think it ever did... in any case, without the registry changes it will most likely be a NOP on windows.

It's ready to get some real world testing in next.

PR #475, opened 2016-01-02
Revocation checker reduction

toad:
I need this for my simulations: Don't request the revocation key if auto-update is turned off.

Low priority, careful testing needed. In particular, it's essential that after the revocation certificate is found, we still show the useralert even after restarting.

This is really a partial solution. The full solution is to only do any of the checks, including the blob file, if auto-update is enabled. But in that case, we need blowing the revocation key to *not turn off auto-update* (for purposes of configuration). In which case, we need to change isEnabled()... That would need more careful testing, although the code changes should be small.

### DIFF
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -509,6 +509,7 @@ public class NodeUpdateManager {
 
 		node.clientCore.alerts.register(alert);
 		
+		revocationChecker.checkForBlobOnDisk();
 		enable(wasEnabledOnStartup);
 
 		// Fetch 3 files, each to a file in the runDir.
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -682,13 +682,6 @@ public class NodeUpdateManager {
 		// }
 		NodeUpdater main = null;
 		Map<String, PluginJarUpdater> oldPluginUpdaters = null;
-		// We need to run the revocation checker even if auto-update is
-		// disabled.
-		// Two reasons:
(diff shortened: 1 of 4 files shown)


unixninja92:
The use of tabs vs spaces is inconstant

src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -509,42 +510,51 @@ public void start() throws InvalidConfigValueException {
...
+	 * and the IP to country file. */
+	private void startSimpleFetches() {
+	    // Fetch 3 files, each to a file in the runDir.
+	    synchronized(this) {
+	        if(startedSimpleFetches) return;
+	        startedSimpleFetches = true;
+	    }
 
-		if (updateSeednodes) {
+        if (updateSeednodes) {
```
Bombe: This line is using spaces for indentation but the method header uses a tab.

src/freenet/support/Waiter.java
```diff
@@ -0,0 +1,23 @@
+package freenet.support;
+
+/** Waits for some other thread to signal to proceed. */
+public class Waiter {
```
Bombe: Why are you not using e.g. a `Semaphore` or a `CountDownLatch` here? Please don’t add custom implementations of classes that are already in the JDK.

toad: Semaphore is over-complex, to the point of creating new bugs. CountDownLatch looks more promising. Will fix.

Bombe: `CountDownLatch` would be fine, too, we just need to stop rolling our own versions of stuff that is already in the JDK and has been for the last ten years. :)

PR #476, opened 2016-01-06
Travis CI: Enable extensive unit tests

xor-freenet:
As Travis is a dedicated server for running tests, it is just the right place for running tests which take longer, so let's enable them.

NOTICE: You check whether this works by searching the stdout on Travis for "test.extensive".
Please do not only check the Ant command line, but also the actual output of the tests: They will print the content of the environment variables they have received, as implemented by commit https://github.com/xor-freenet/fred-staging/commit/1a5731c39d5b6e08c2a525d1b176ae90cd8952f0


xor-freenet:
Hmm apparently this does NOT work currently, the tests do not seem to run the test freenet.support.TestProperty.
I'll investigate and re-open.

PR #477, opened 2016-01-06
build: make Bouncy Castle path configurable

Thynix:
Debian Jessie (current stable) packages Bouncy Castle 1.49, but Freenet
requires later versions. The Debian build is configured to search
/usr/share/java for its libraries, but a usable version cannot
practically be there because it is not packaged. Therefore the build
must be able to search another directory.

I'm submitting this because I'm trying to get some work done on the Debian package, and this change is 1) already done 2) easier than finishing the move to Gradle.

### DIFF
--- a/build-clean.xml
+++ b/build-clean.xml
@@ -26,6 +26,7 @@ maintainers may prefer to use this instead of build.xml.
 	<path id="lib.path">
 		<fileset dir="${lib.contrib.dir}" includes="${lib.contrib.jars}"/>
 		<fileset dir="${lib.dir}" includes="${lib.jars}"/>
+		<fileset dir="${bcprov.dir}" includes="${bc.jar}"/>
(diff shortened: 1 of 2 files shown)


nextgens:
As far as I know it does already... it looks for ${lib.dir}/${bc.jar} and will use that over /usr/share/java/${bc.jar}

infinity0:
Yes, I believe the existing variables should already do what you need. BTW, bouncycastle is now 151 in Debian testing/unstable. I spoke to the #debian-java team on OFTC IRC today and apparently it's hard to upgrade it to 154 (the latest version) because they keep breaking their API so they have to test it against all other packages first. Is 151 OK for Freenet or does it need 152?

Thynix:
Ah, the Debian build is explicitly specifying /usr/share/java as its ${lib.dir}, but that's no longer necessary.

Freenet needs 152 - 151's skipping ciphers are too slow. See https://freenetproject.org/news.html#20150713-1468-known-issues

PR #479, opened 2016-01-08
let the JVM handle the SHA256 pool

nextgens:
Also, update the l10n strings suggesting that plugins require a newer version of the JVM


bertm:
Was b345567 intended to be included in this pull request?

src/freenet/crypt/SHA256.java
```diff
@@ -101,9 +87,9 @@ public static MessageDigest getMessageDigest() {
 		try {
 			MessageDigest md = null;
 			synchronized(digests) {
-				int x = digests.size();
-				if(x == 0) md = null;
-				else md = digests.remove(x-1);
+				while (digests.size() > 0 && md == null) {
+					md = digests.remove(0).get();
+				}
```
bertm: This change increases the worst-case complexity from O(n) to O(n²). You seem to be emulating queue behaviour here to reuse the least recently used digest, why not use a `Queue` instead?

bertm:
While you're touching the l10n, please also remove references to Sun in favor of Oracle. Sun Java is no more.

src/freenet/crypt/SHA256.java
nextgens: I don't think that the complexity matters here; it's meant to be small (we had 16 elements max).
We'll never get around clearing the queue of SoftReferences that are gone if I do LIFO... is it worth having a cleanup job on the ticker?
What type of non-blocking Queue do you have in mind? a LinkedList? that's a waste of memory IMHO

bertm: From an algorithmic perspective I was thinking of `ConcurrentLinkedQueue`, but you are right in that it wastes some memory with an upside of possibly reducing thread contention. But just a simple ArrayDeque with synchronization will probably do as well, I don't think locking is a true bottle-neck here.

bertm:
All changes look good to me. I'm not sure of the relevance of the changes to `KeyAgreementSchemeContext` in this pull request, but I have no objections on having them merged.

Thank you for dealing with this @nextgens!

PR #480, opened 2016-01-09
Fix peerstuff

nextgens:
Current next doesn't write to the peer-file when a new darknet peer is added... if you restart the node you loose the connection.

I haven't managed to track down when this was introduced yet... but we want that merged sooner rather than later :/

### DIFF
--- a/src/freenet/node/DarknetPeerNode.java
+++ b/src/freenet/node/DarknetPeerNode.java
@@ -212,7 +212,6 @@ public class DarknetPeerNode extends PeerNode {
 
 		// Setup the queuedToSendN2NMExtraPeerDataFileNumbers
 		queuedToSendN2NMExtraPeerDataFileNumbers = new LinkedHashSet<Integer>();
-		writePeers();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


nextgens:
Experimentally, I've found out that this affects 1471-pre3 too... so it's a good candidate for backport

Spawn a new node up, add a darknet peer... shutdown the node ... has a peer file been created?

nextgens:
from @bertm on IRC: "With 1440 I can't even get it to save darknet peers at all"

@toad is that the right fix? any idea of what/when it was introduced?

bertm:
Since 038ed47 peers get saved in 5-second intervals (5 minutes since ad6170c), but not on shutdown AFAICS.

bertm:
The write-peers-on-shutdown bug was fixed years ago in `064db87ee`, but apparently there are still code paths where we fail to flag that the peers file needs updating. 
`0e8618b` properly ensures we always write peers on shutdown, even if we failed to indicate this was necessary, and fixes the single *known* code path that failed to indicate this as well.

Tested, this works! :+1:

Thynix:
Is this ready for 1471 then?

bertm:
Affirmative.

Thynix:
Alright. Thanks! Merged in 51feed3a5a6c2e0363417107794d91d24ba1426f from a rebase of these commits onto 1471-stable which was also merged into 1471-stable.

PR #481, opened 2016-01-09
Remove the DSA params from the noderefs

nextgens:
This is the second part of #440 

DO NOT MERGE before 1471 is mandatory!

### DIFF
--- a/src/freenet/node/NodeCrypto.java
+++ b/src/freenet/node/NodeCrypto.java
@@ -67,12 +67,6 @@ public class NodeCrypto {
 	byte[] identityHashHash;
 	/** Nonce used to generate ?secureid= for fproxy etc */
 	byte[] clientNonce;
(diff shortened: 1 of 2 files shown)


src/freenet/node/NodeCrypto.java
```diff
@@ -495,18 +443,11 @@ private String ecdsaSignRef(String mySignedReference) throws NodeInitException {
 
 	void addPrivateFields(SimpleFieldSet fs) {
 	    // Let's not add it twice
-	    fs.removeSubset("ecdsa");
-	    fs.put("ecdsa", ecdsaP256.asFieldSet(true));
```
unixninja92: Why did you delete these tabs?

nextgens: I've replaced the four spaces with two tabs... to be consistent with the rest of the file

PR #482, opened 2016-01-12
Compare the right number of bytes

toad:
Please merge into next *right now*. Nextgens' hash refactoring broke basic connectivity.

### DIFF
--- a/src/freenet/crypt/HMAC.java
+++ b/src/freenet/crypt/HMAC.java
@@ -6,11 +6,10 @@ package freenet.crypt;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
-
-import org.bouncycastle.util.Arrays;
 
 import freenet.support.Logger;
 
--- a/src/freenet/crypt/HMAC.java
+++ b/src/freenet/crypt/HMAC.java
@@ -10,6 +10,8 @@ import java.security.NoSuchAlgorithmException;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 
+import org.bouncycastle.util.Arrays;
+
 import freenet.support.Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


unixninja92:
Looks good to me

src/freenet/node/NewPacketFormat.java
```diff
@@ -398,7 +398,7 @@ private NPFPacket decipherFromSeqnum(byte[] buf, int offset, int length, Session
 		byte[] payload = Arrays.copyOfRange(buf, offset + hmacLength, offset + length);
 		byte[] hash = Arrays.copyOfRange(buf, offset, offset + hmacLength);
 
-		if(!HMAC.verifyWithSHA256(sessionKey.hmacKey, payload, hash)) return null;
+		if(!HMAC.verifyWithSHA256(sessionKey.hmacKey, payload, hash, HMAC_LENGTH)) return null;
```
bertm: The rest of the code uses `hmacLength`, including the lines above this one. Please use that one instead.

toad:
What I don't understand is why the original breakage passed unit tests. Or did it? Am looking into this...

Thynix:
It did pass.

toad:
See #483

src/freenet/node/NewPacketFormat.java
```diff
@@ -398,7 +398,7 @@ private NPFPacket decipherFromSeqnum(byte[] buf, int offset, int length, Session
 		byte[] payload = Arrays.copyOfRange(buf, offset + hmacLength, offset + length);
 		byte[] hash = Arrays.copyOfRange(buf, offset, offset + hmacLength);
 
-		if(!HMAC.verifyWithSHA256(sessionKey.hmacKey, payload, hash)) return null;
+		if(!HMAC.verifyWithSHA256(sessionKey.hmacKey, payload, hash, hmacLength)) return null;
```
nextgens: Why the hell aren't we logging anything when it fails?

src/freenet/crypt/HMAC.java
```diff
@@ -53,6 +54,10 @@ public static boolean verify(HMAC hash, byte[] key, byte[] data, byte[] mac) {
     return MessageDigest.isEqual(mac, mac(hash, key, data));
   }
 
+  public static boolean verify(HMAC hash, byte[] key, byte[] data, byte[] mac, int length) {
```
nextgens: This isn't fool-proof; I wanted to get rid of the length parameter to avoid that type of problems.
IMHO we should enforce a length check on the key here ... and thow an IllegalArgumentException in case it's less than crypto-safe.

nextgens:
This was introduced by me in #467 ... congratz on catching it :)

src/freenet/crypt/HMAC.java
nextgens: Or better yet, keep the truncated key in the sessionKey object and don't change the API

nextgens:
I've submitted #484 to address the above; thanks again @toad for debugging it!

PR #483, opened 2016-01-12
Implement test for packet encryption

toad:
This fails on next. It passes with fix-hash-regression. It is no more ugly and timing-sensitive than any other test in the same file, and works for me. Also note that it is single-threaded; while there are sleep()'s, it should at least be close to deterministic.


toad:
Note that this test fails without #482 
This is intentional.

nextgens:
I've submitted #484 to address the above; thanks again @toad for debugging it!

PR #484, opened 2016-01-13
An alternative to #482 and #483

nextgens:
This fixes the indentation/formatting problems as well as resolve the problem without changing the HMAC API

### DIFF
--- a/src/freenet/node/NewPacketFormat.java
+++ b/src/freenet/node/NewPacketFormat.java
@@ -30,7 +30,6 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 
 public class NewPacketFormat implements PacketFormat {
 
-	private final int hmacLength;
 	private static final int HMAC_LENGTH = 10;
 	// FIXME Use a more efficient structure - int[] or maybe just a big byte[].
 	// FIXME increase this significantly to let it ride over network interruptions.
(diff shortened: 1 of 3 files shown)


src/freenet/node/NewPacketFormat.java
```diff
@@ -30,7 +31,7 @@
 public class NewPacketFormat implements PacketFormat {
 
 	private final int hmacLength;
-	private static final int HMAC_LENGTH = 10;
+	public static final int HMAC_LENGTH = 10;
```
bertm: What justifies making this public? This should be of no interest to classes other than `NewPacketFormat`.

src/freenet/node/NewPacketFormat.java
```diff
@@ -397,8 +398,14 @@ private NPFPacket decipherFromSeqnum(byte[] buf, int offset, int length, Session
 
 		byte[] payload = Arrays.copyOfRange(buf, offset + hmacLength, offset + length);
 		byte[] hash = Arrays.copyOfRange(buf, offset, offset + hmacLength);
+		byte[] localHash = Arrays.copyOf(HMAC.macWithSHA256(sessionKey.hmacKey, payload), HMAC_LENGTH);
```
bertm: The rest of the code, including the lines directly above this one, use `hmacLength` instead of `HMAC_LENGTH`. Please choose one.

nextgens: done in 6789a1b
You're reviewing faster than I can push :)

nextgens:
I'm happy with the above; I suggest we merge/consider that instead of #483  and #482

PR #485, opened 2016-01-17
SSK inserts: commit before reply and collisions

toad:
For review and careful testing. Two changes:
1) Always commit before sending InsertReply, just like we do in CHKInsertHandler (this is IMHO needed for consistency but it's also relevant to bug #3338, see the big comment).
2) Fix collisions not working correctly in some corner cases / race conditions.

### DIFF
--- a/src/freenet/node/SSKInsertHandler.java
+++ b/src/freenet/node/SSKInsertHandler.java
@@ -201,22 +201,10 @@ public class SSKInsertHandler implements PrioRunnable, ByteCounter {
 		SSKBlock storedBlock = node.fetch(key, false, false, false, canWriteDatastore, false, null);
 		
 		if((storedBlock != null) && !storedBlock.equals(block)) {
-			try {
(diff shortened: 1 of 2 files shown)


toad:
I would like this to be merged eventually. But it needs testing first, including in vitro. I will try it out in simulations...

toad:
One thing that needs careful testing is whether this results in more forced disconnections. :(

toad:
So far this works fine in one-request-at-a-time simulations. It will need testing more widely. This is not urgent.

toad:
Works fine in sequential simulations. Haven't tested more widely yet.

bertm:
Except for some formatting issues this looks reasonable.

toad:
I'm afraid I'm likely to ignore complaints that amount to "don't use Eclipse". :(

Juiceman:
On Jan 31, 2016 12:39 PM, "Matthew Toseland" <notifications@github.com>
wrote:
>
> I'm afraid I'm likely to ignore complaints that amount to "don't use
Eclipse". :(
>
> —

Whatever happened to formatting the code base en masse?

Thynix:
Automatic formatting isn't viable when there's a lot of existing code we
still want to merge, which is currently the case for Fred.

PR #486, opened 2016-01-18
Fix race condition in MultiMessageCallback

toad:
If message.sent() is called before cb.arm(), then we only call cb.sent() after completion. We should call it on arm()ing. This probably only affects simulations, but possibly also very busy nodes.

This is a minor accounting glitch, low priority, but might conceivably caused forced disconnects.

### DIFF
--- a/src/freenet/node/MultiMessageCallback.java
+++ b/src/freenet/node/MultiMessageCallback.java
@@ -6,7 +6,21 @@ import freenet.io.comm.AsyncMessageCallback;
  * You should add messages with make() and then call arm(). sent(boolean)
  * will be called when all the messages have been sent (or failed e.g. 
  * disconnected) and finish(boolean) will be called when all the messages 
- * have been acknowledged (or failed). */
+ * have been acknowledged (or failed). Typical usage:
+ * <pre>Message m1 = ...;
+ * Message m2 = ...;
+ * PeerNode pn = ...;
+ * MultiMessageCallback mcb = new MultiMessageCallback() {
+ *   protected void finish(boolean success) {
+ *     // Messages have finished.
+ *   }
+ *   protected void sent(boolean success) {
+ *     // Messages have been sent.
+ *   }
+ * }
+ * pn.sendAsync(m1, mcb, ctr);
+ * pn.sendAsync(m2, mcb, ctr);
+ * mcb.arm();</pre> */
 public abstract class MultiMessageCallback {
 	
     /** Number of messages that have not yet completed */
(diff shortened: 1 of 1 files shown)


ArneBab:
it looks like indentation is inconsistent here — mixing tabs and spaces?

ArneBab:
Could you add a comment in line 11 of the file what the variable armed means?

toad:
Indentation is evil. That is, Eclipse is evil. What do you suggest I do about it? Reindent the whole file with Eclipse? (Big diff). Replace spaces with tabs with a script? (Big diff but tiny with --ignore-space-change)?

ArneBab:
@toad just make the change fit the surrounding code. You could try emacs (M-x tabify or M-x untabify).

src/freenet/node/MultiMessageCallback.java
```diff
@@ -2,14 +2,37 @@
...
 public abstract class MultiMessageCallback {
 	
+    /** Number of messages that have not yet completed */
 	private int waiting;
+	/** Number of messages that have not yet been sent */
 	private int waitingForSend;
 	
+	/** True if arm() has been called. finish(boolean) and sent(boolean) will 
+	 * only be called after arming the callback. */
 	private boolean armed;
```
ArneBab: followup question: what does arm() do? I would guess that it prepares the message?

toad: See below.
	/** Enable the callback. The callbacks sent(boolean) and finish(boolean)
	 * will only be called after this method has been called, so you should 
	 * use this to indicate that you won't add any more messages. */ 
Is this unclear?

ArneBab: Matthew Toseland writes:

>>  	private int waitingForSend;
>>  	
>> +	/** True if arm() has been called. finish(boolean) and sent(boolean) will 
>> +	 * only be called after arming the callback. */
>>  	private boolean armed;
>
> See below.
> 	/** Enable the callback. The callbacks sent(boolean) and finish(boolean)
> 	 * will only be called after this method has been called, so you should 
> 	 * use this to indicate that you won't add any more messages. */ 
> Is this unclear?

I missed that — I think it’s pretty clear. So arming means something
like finalizing a packet?

Best wishes,
Arne

-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

toad: A group of packets. The whole point here is we get a callback when we've sent all of a bunch of messages. And then another callback when they're all "finished" (i.e. acknowledged). But not before we call arm(), for the sake of consistency of the caller. See the example in the top comment.

toad: I have improved the top-of-file class javadoc a little to reflect this confusion.

ArneBab: thank you!

PR #487, opened 2016-01-18
MemoryLimitedJobRunnerTest: avoid high thread counts

Thynix:
With high thread counts on some systems this can intermittently result in the JVM hitting a thread count limit and being unable to create another thread.

That said, what thread limit is this, where is it set, and can the test ensure it is high enough before starting? xor recommended breaking on OOM in a debugger and checking if the thread count really is high.


bertm:
I think the newly proposed thread counts still qualify as "many threads" so I'm fine with the changes, if that solves your problem.

> That said, what thread limit is this, where is it set, and can the test ensure it is high enough before starting?

To start, we would need a reliable way of triggering this behaviour. I've not encountered it first-hand yet.

PR #488, opened 2016-01-18
updater: change old update key to new for Java 7+

Thynix:
I'm posting this for review as requested but have yet to test it. As Matthew said previously:

> The resulting UOM should be tested to work with {windows, linux} * {old to new, new to new} * {ordinary fetch, UOM}.

I don't know if I'm going to bother to test {Windows, Linux} because the fetching doesn't differ between the two.

### DIFF
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -47,6 +47,7 @@ import freenet.node.useralerts.UserAlert;
 import freenet.pluginmanager.OfficialPlugins.OfficialPluginDescription;
 import freenet.pluginmanager.PluginInfoWrapper;
 import freenet.support.HTMLNode;
+import freenet.support.JVMVersion;
 import freenet.support.Logger;
 import freenet.support.api.BooleanCallback;
 import freenet.support.api.Bucket;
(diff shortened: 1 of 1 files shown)


Thynix:
To detail the testing plans, I'll work from this commit, create new temporary testing update keys, and build:

* 1470 to request UOM from 1471 (with self-mandatory) and normal key (without self-mandatory)
* 1471 to serve updates to 1470 and update from "1472" (UOM / key)
* "1472" (with self-mandatory) to serve UOM to 1471

Anything else to check?

ArneBab:
should 1472 also serve an update to 1470?

Thynix:
I don't think that tests anything new? That would already include all the code changes in 1471 - just that 1472 serves an update over UOM to 1471. I can look into it anyway. I'm hoping to write a script which tests all this.

toad:
That sounds good.

Re Windows, the deployment is different (we *always* have to change stuff in wrapper.conf, we can't rename files), but yes the fetching isn't much different (there can be Windows-specific components).

toad:
You need to test that 1471 updates the key and can then upgrade to 1472 on the new key? Hopefully this is just clarifying one of your points above.

Thynix:
Not just clarifying but it shouldn't have been. I'll adjust the test to do that.

Thynix:
I wrote [a script](https://github.com/freenet/scripts/commit/7455d640f13106c184e416b76f3e3edf15bf2421) to make the requisite test builds. I'll stop short of making a script to also do the test; this has been slower going than I had hoped. Next is actually doing the test, and, assuming it works and there are no further reservations, merging this and releasing 1471.

Thynix:
I tested updating from the keys (not UOM). It was successful. Notes / directions:

> Run build_autoupgrade. Set up two fresh nodes installed from 1470's installer, designating one as "upstream" and the other as the "upgrade target." Connect them over darknet. Shut both down. Copy the 1470 autoupgrade jar to the upgrade target, and the non-mandatory 1472 to upstream. Run "upstream," connect it to opennet, insert 1471 to the old key/jar-1471 and new key/jar-1471. Insert 1472 (non-mandatory) to new key/jar-1472. Modify freenet.ini for the upgrade target to update from the old key, and not automatically install new versions. (node.updater.URI to the old key; node.updater.autoupdate to false.) This will ensure that it checks the intended key instead of the stable 1470 key, and that one does not miss any upgrades happening.
>
> Key-based upgrade testing:
> Connect the upgrade target to opennet and ensure that it discovers 1471, and after upgrading discovers and upgrades to 1472.

Next is to test updating over UOM.

Thynix:
Testing UOM was also a success: *

> UOM upgrade testing:
> Set the upgrade target back to 1470, (in my case the wrapper was now loading from `freenet.jar.new` instead of `freenet.jar`, and remove its node.updater.URI setting so it goes back to its default of the old key. Disconnect both nodes from opennet, and check that 1470 UOMs to 1471, then 1472.

* The node wasn't able to fetch 1472, but I think it may have been that the "upstream" node was running mandatory, while the non-mandatory was inserted. The upgrade target logged (and mentioned in the UI) that it was fetching the main jar, but kept failing. I've updated the key testing instructions to insert mandatory 1472 so that this part of the test will be consistent.

toad:
See #494. I seriously doubt that your UOM testing was correct. It looks like we announce the transition build (1471) via the wrong kind of announcement message, which will prevent 1470 from using it.

Thynix:
I'm confused how it was successful in my testing then. I will check again
with only UOM.

Thynix:
I got to step 6 and wasn't able to get the insert to complete on the single node. Out of time for now.

> 1. Set up Upstream and UpgradeTarget 1470 nodes from the published installer.
> 2. Configure them to darknet mode with 512 MiB stores.
> 3. Connect them over darknet at high trust.
> 4. Shut them both down; remove the `node.updater.URI` setting and replace the `freenet.jar` with > the Upstream and UpgradeTarget with the 1471 self-mandatory and 1470 customized update URI builds respectively. Add "node.writeLocalToDatastore=true" to the Upstream.
> 5. Start up the Upstream node.
> 6. Insert the 1471 build to the old update URI (with LocalRequestOnly set to True; see insert_autoupgrade)
> 7. Start the UpgradeTarget.

toad:
Send the FCP request with LocalRequestOnly=true and PriorityClass=0. That should work IIRC.

Thynix:
I did send it with LocalRequestOnly=true, and whatever the default priority
class is. Do you recall how long the limiting is? Can I disable it? Why
does it exist?

toad:
5 minutes, but it only applies if opennet is turned on. I will test locally, 1mo...

toad:
Possibly it could still get stuck in a queue if it's low priority? Testing here...

toad:
ClientHello
Name=Toad
ExpectedVersion=2
End

ClientPut
URI=CHK@
UploadFrom=disk
Filename=/home/toad/Music/Non-classical/FreeMusic/wired-creative-commons-cd/01-beastie-boys-now-get-busy.mp3
Verbosity=1023
Identifier=test
LocalRequestOnly=true
PriorityClass=0

->

PutFailed
ExtraDescription=Got metadata from unknown state
Fatal=true
CodeDescription=Internal error
Identifier=test
Code=3
Global=false
ShortCodeDescription=Internal error
EndMessage

Argh!

toad:
And of course when I turn on logging, it works. :(

Anyway, the point is, it works, at least sometimes. And there's no delay.

toad:
With default priority, it immediately reaches 100%, but it looks like it may stall inserting the last (top) block.

Thynix:
I've resumed testing and narrowed down why I wasn't able to get it to insert: pyFreenet did not propagate otherwise unrecognized keywords to `put()` like I thought it did, so once I added `LocalRequestOnly` support the insert started. I also set the priority to 0 and enabled "Write local and nearby requests to the datastore?" but my assumption is those were not necessary. I'll update with more details as they happen.

Thynix:
Lovely, I've run into an "Internal error" like you did. Perhaps LocalRequestOnly inserts should have a unit test?

Thynix:
I'm not able to get a UOM-only update from 1470 to 1471. I wonder why it succeeded. Maybe I made a mistake or the nodes updating previously with the key allowed them to cache the updates. I'll look into #494.

PR #489, opened 2016-01-18
Opennet RequestSender cleanup

toad:
This makes RequestSender responsible for actually sending node references (and acknowledgements) to the data source. It should make opennet behave more consistently. It needs careful testing.


src/freenet/node/RequestHandler.java
```diff
@@ -777,7 +777,7 @@ private void finishOpennetInner(OpennetManager om) {
 			noderef = rs.waitForOpennetNoderef();
 		} catch (WaitedTooLongForOpennetNoderefException e) {
 			sendTerminal(DMT.createFNPOpennetCompletedTimeout(uid));
-			rs.ackOpennet(rs.successFrom());
+			// RequestSender will have already acknowledged.
```
bertm: This would be a good place to add an assertion on a package-private getter for `RequestSender.opennetFinishedRelaying`.

PR #490, opened 2016-01-18
No path folding at high HTL

toad:
For consideration. This would improve security (e.g. no possibility of attacker data source getting directly connected to request originator just by path folding), but it may have negative performance impacts, especially on new nodes who mostly get their peers from doing local requests.

### DIFF
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1707,8 +1707,10 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
         		return false;
         	}
         	
-            if(!node.canWriteDatastoreRequest(origHTL))
+            if(!node.canWriteDatastoreRequest(origHTL)) {
                 ackOpennet(next);
+                return false;
+            }
         	
 			if(node.addNewOpennetNode(ref, ConnectionType.PATH_FOLDING) == null) {
 				if(logMINOR) Logger.minor(this, "Don't want noderef on "+this);
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1716,7 +1716,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
 				// RequestHandler will send a noderef back up, eventually, and will unlockHandler() after that point.
 				// But if this is a local request, we need to send the ack now.
 				// Serious race condition not possible here as we set it.
-				if(source == null)
+				if(!node.canWriteDatastoreRequest(origHTL))
 					ackOpennet(next);
 				else if(origTag.shouldStop()) {
 					// Can't pass it on.
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1708,6 +1708,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
         	}
         	
             if(!node.canWriteDatastoreRequest(origHTL)) {
+                // Do not path fold at all at high HTL.
                 ackOpennet(next);
                 return false;
             }
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1720,8 +1720,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
 					opennetNoderef = noderef;
 				}
 				// RequestHandler will send a noderef back up, eventually, and will unlockHandler() after that point.
-				// But if this is a local request, we need to send the ack now.
-				// Serious race condition not possible here as we set it.
+				assert(source != null);
 				if(origTag.shouldStop()) {
 					// Can't pass it on.
 					origTag.finishedWaitingForOpennet(next);
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1707,6 +1707,9 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
         		return false;
         	}
         	
+            if(!node.canWriteDatastoreRequest(origHTL))
+                ackOpennet(next);
+        	
 			if(node.addNewOpennetNode(ref, ConnectionType.PATH_FOLDING) == null) {
 				if(logMINOR) Logger.minor(this, "Don't want noderef on "+this);
 				// If we don't want it let somebody else have it
@@ -1716,9 +1719,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
 				// RequestHandler will send a noderef back up, eventually, and will unlockHandler() after that point.
 				// But if this is a local request, we need to send the ack now.
 				// Serious race condition not possible here as we set it.
-				if(!node.canWriteDatastoreRequest(origHTL))
-					ackOpennet(next);
-				else if(origTag.shouldStop()) {
+				if(origTag.shouldStop()) {
 					// Can't pass it on.
 					origTag.finishedWaitingForOpennet(next);
 				}


toad:
This may also have an impact on the number of long links, and it was originally proposed to deal with that. However, we have code to deal with that directly, enforcing 30% of the routing table to be long and 70% short. With this patch, path folding will probably produce fewer long link candidates.

toad:
Unfortunately I am not yet in a position to test opennet in multi-node simulations. Maybe I will be in coming months.

src/freenet/node/RequestSender.java
```diff
@@ -1716,9 +1722,7 @@ private boolean finishOpennet(PeerNode next) {
 				// RequestHandler will send a noderef back up, eventually, and will unlockHandler() after that point.
 				// But if this is a local request, we need to send the ack now.
```
ArneBab: is this comment still up to date with this change, or should the “local request needs ack” line  be removed?

toad: Good point, I have moved that up above this line.

toad:
Note that #491 may be a more appropriate fix.

src/freenet/node/RequestSender.java
ArneBab: thank you!

ArneBab:
returning less long link candidates is likely a good thing. The 30/70 thing is a brute force method, this one should make it work smoother (since the links returned from path folding will be a better match for what the node requires).

toad:
It might. Or there might be too few. Either way, newbie nodes are likely to take longer to bootstrap, unless we take other measures such as announcing through any peer (I think there's a pull request for this?). Agree we should probably try it at some point, but we need some systematic stats first.

ArneBab:
Matthew Toseland writes:

> It might. Or there might be too few. Either way, newbie nodes are likely to take longer to bootstrap, unless we take other measures such as announcing through any peer (I think there's a pull request for this?). Agree we should probably try it at some point, but we need some systematic stats first.

There isn’t yet a pull request, I think, but there is a bug report. And
I don’t think we need additional stats to know that it would be useful
to announrce through any peer, since the seednodes are a single point of
failure.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

toad:
I thought the opennet-changes branch did this?

ArneBab:
if that’s the case, that’s great!

toad:
So in conclusion, this is longer term ... #491 should be considered for 1472...

ArneBab:
Why should there be too few nodes? We route with HTL 18, so we only lose about 10% of the possible folding options — mostly the long ones (of which we get too many anyway) — which means that bootstrap should become a bit slower, but not seriously slower. And the link length distribution could improve seriously.

toad:
Maybe. I'd want to simulate it first, and have much better stats gathering (digger3's stats are down).

Bootstrapping would be dramatically slower because new nodes would get no nodes from path folding from successful local requests - they'd have to wait until they have successful relayed requests. However, the opennet-changes branch may offset this.

toad:
We could test this without a full blown opennet simulation. The necessary changes will work fine on their own on the end-node.

bertm:
This looks good, except for formatting. I agree with @ArneBab that this is conceptually sound, but haven't looked at your alternative #491 yet.

ArneBab:
I’m testing this right now, and it fails to connect for two nodes here. One got one working connection, the other got none.

Time to find out why the concept doesn’t work…

ArneBab:
now I merged into build 1470, one node without folding at high HTL (Freenet1), one regular (Freenet2). Both nodes got 3 connections, but the one without folding at high HTL only reports 16 rejections, the regular one reports 36 rejections.

I’ll keep both running for a while.

toad:
Bear in mind that there is IP-address-based limiting on the seednodes, and announcement is rather hit-and-miss anyway. So bootstrapping more than one node on the same IP in the same day may work poorly regardless.

toad:
Also, this patch should have *no* effect on announcements - except to make new nodes more dependent on them.

ArneBab:
status: 6 connections with this pull-request, 11 without it.

ArneBab:
now both have 15-17 connections.

ArneBab:
canWriteDatastoreRequest is `HTL <= maxHTL - 2`, so it’s 16 or less. Given that the success rate is highest at HTL 17, this means most requests cannot fold, which looks excessive to me. 

Does this mean that it only starts folding after the first 2.5 hops? Or is it 1.5 hops (just skipping direct neighbors)?

If it is 2.5 hops, HTL 17 might suffice.

```
CHK                                                                SSK
18 	31,505% (1032,75121,241721) (,0006) 	1,342% (597,3075,273704) (,0043)
17 	34,155% (1227,69383,206735) (,0003) 	1,763% (549,2371,165639) (,0012)
16 	30,653% (1469,112647,372288) (,0002) 	0,775% (699,1167,240673) (,0004)
15 	26,292% (1111,109847,422017) (,0001) 	0,335% (378,491,259291) (,0002)
14 	15,796% (168,40717,258836) (,0002) 	0,084% (71,109,213087) (,0003) 
```

status: just increased upload bandwidth to 60 == 27 peers.

ArneBab:
* no high HTL: 17 peers
* regular: 24 peers

ArneBab:
Now 18 vs. 25. However the no high HTL peers are all connected while the regular peer has a quarter to half its peers busy.

ArneBab:
For analysis:

Node locations no high HTL:

```
0.16986777257793562
0.17901946183666562
0.5052537534294653
0.5086700871359677
0.6390555069251317
0.6822584345099271
0.8980318688923236
0.9450003036258698
0.9472578481300644
0.9475679196733452
0.9505918714537117
0.9512505057035953
0.9512848146581615
0.951332629568245
0.953138735932966
0.9534703560289312
0.9537457741318474
0.9612296077880356
0.18339809750763603
0.9517140598028574
0.9521872029906774
0.9531178654323221
```

Node locations, regular:

```
0.1683899971760181
0.5462565170155342
0.6584651359671296
0.6669163773467959
0.6752661195055957
0.8012936926640807
0.8020591962695599
0.8066928956581909
0.8081615954443561
0.8094734715697488
0.8100868389013971
0.8121570853807678
0.8136131131663924
0.8150852508230284
0.8164318649800096
0.9073881084564737
0.7200454861860763
0.8031756611247909
0.8047278039539334
0.8095277095841826
0.8103247800000128
0.8118534916279835
0.813718338962537
0.8137287779436605
0.8139409096231633
0.8184079088556605
0.7194572283696072
```

ArneBab:
now 23 vs. 27.

Peer distribution no high HTL:

```
0.17901946183666562
0.5052537534294653
0.8980318688923236
0.9008688888658157
0.9450003036258698
0.9472578481300644
0.9504251829525352
0.9505918714537117
0.9512505057035953
0.9512848146581615
0.951332629568245
0.953138735932966
0.9534703560289312
0.9537457741318474
0.9543209138274216
0.9553451901226157
0.9562020585339739
0.9564390915303991
0.9612296077880356
0.14636055660556124
0.8935836794924024
0.9509168180104608
0.9530186614237126
0.5086700871359677
0.9475679196733452
0.9502900387595263
0.9615630709874536
```

regular:

```
0.1683899971760181
0.40145412492581534
0.5462565170155342
0.6669163773467959
0.7200454861860763
0.7993934278340358
0.8003752187303285
0.8012936926640807
0.8018909930668665
0.8066928956581909
0.8081615954443561
0.8100868389013971
0.8118534916279835
0.8134545396525108
0.8136131131663924
0.8137287779436605
0.8139409096231633
0.8150852508230284
0.8164318649800096
0.8184079088556605
0.9073881084564737
0.9076888086179288
0.8031756611247909
0.8095277095841826
0.8103247800000128
0.813718338962537
0.8094734715697488
```

ArneBab:
26/27

ArneBab:
I’m shutting down the regular node. The no-high-HTL one will run a bit longer to see whether it settles.

ArneBab:
7 hours later (not yet analyzed):

```
0.0161567010603062
0.14977241274358533
0.2535342406046017
0.6398256029475275
0.7907590894018743
0.8991295287193819
0.9450003036258698
0.9472578481300644
0.9474802650784804
0.9477809820924051
0.9504251829525352
0.9505918714537117
0.9512848146581615
0.9517140598028574
0.9521872029906774
0.9530186614237126
0.9534703560289312
0.9535672941936432
0.9553451901226157
0.9579391787964676
0.959033783035655
0.9469400351870869
0.9482952331226397
0.9514408501738144
0.9543209138274216
0.686718197657147
```

ArneBab:
For comparing:

```
import pylab as pl                        
def plotfromtext(a, label):
   b = [float(i) for i in a.splitlines() if i]
   c = [(i - pl.median(b))%0.5 for i in b]
   pl.plot(sorted(c), label=label)
   pl.xscale('log')
```

toad:
Ok so you've confirmed that it works but it takes longer to bootstrap, which is pretty much what we expected???

Re hops taken, you can't deduce much from success rates... But you were looking at request counts?

Could such a bias be a good thing? If something is found quickly then presumably it's very close? Does that mean we'd tend to get more longer links from path folding??

toad:
Also, we spend 2 hops at HTL 18 on average.

ArneBab:
Yes, I confirmed pretty much what we expected.

I looked at request counts, too, and they were actually higher for the no-high-HTL-folding node.

If something is found quickly, it could also mean that it is popular and cached. Also the mean path length is likely pretty short, given that we have ~20% nodes with around 100 peers. The long paths might actually be an artefact of backoff or other structural problems.

Mathy: Every 10-peers node (the most common type) will on average know 2 100-peers nodes. These 100-peers nodes will know on average 20 other 100-peer nodes. So if the first step is to a 100 peer node (because of FOAF that’s very likely), the second step directly samples 2000 locations directly, or half the active network.

A 25 peer node (the second most common type) will know 5 100 peer nodes and already sample almost 1000 nodes directly via FOAF information, 1/5th of the active network.

So at least 20% of requests should succeed after two steps by plain greedy routing, regardless of the structure of the network and without taking caching into account.

http://127.0.0.1:8888/USK@pxtehd-TmfJwyNUAW2Clk4pwv7Nshyg21NNfXcqzFv4,LTjcTWqvsq3ju6pMGe9Cqb3scvQgECG81hRdgj5WO4s,AQACAAE/statistics/1115/plot_peer_count.png

ArneBab:
I’ll do a second test where I only fold at HTL17 or less.

ArneBab:
This is the adjustment I did: https://github.com/ArneBab/fred-staging-1/commits/no-path-folding-at-high-htl-stable

https://github.com/ArneBab/fred-staging-1/commit/7bb68248d756477bab8ebb096fab8c5499205237

ArneBab:
This one is at 22/26 connections now.

ArneBab:
HTL 17 after 3 hours:

```
0.10998207881876276
0.16670290904813168
0.1676192020810322
0.17001265429997814
0.17040095342436268
0.17365242908537804
0.17377395870664814
0.17572977729837858
0.17721368806739535
0.17755030372714942
0.17757397481404558
0.1775919012828171
0.17929078064524018
0.18001103481050806
0.18386489086220337
0.2240608138033563
0.22466352092455077
0.22466856568796478
0.22546456504072665
0.3153360148573653
0.37385288068031974
0.6654580550256884
0.168872664500072
0.16895151242051676
0.17161991917778585
0.1733260305060108
0.17156227040992078
0.17637919789205936
```

ArneBab:
after going down to 10 peers and then up to 27 again: 

```
0.06775750812908476
0.12303087483467923
0.12381656020970333
0.1663176153315853
0.16670290904813168
0.17040095342436268
0.17721368806739535
0.17755030372714942
0.17757397481404558
0.1775919012828171
0.1794950910742692
0.18001103481050806
0.18116350063459064
0.22546456504072665
0.413341841399664
0.7862106069713938
0.7867803200745312
0.8775555326480001
0.17001265429997814
0.1733260305060108
0.17365242908537804
0.17377395870664814
0.17423331906747197
0.17572977729837858
0.17603233842240396
0.17637919789205936
0.17929078064524018
```

PR #491, opened 2016-01-18
Random delay for opennet ack when not path folding (locally)

toad:
This should improve protection against local attackers. I would like it to be considered for 1472 if not earlier. The performance impact should be marginal: Path folding does not block returning the data, although maybe we might end up running slightly fewer requests as a result. If so increasing the limits (BANDWIDTH_LIABILITY_*) in NodeStats would be a simple solution.

### DIFF
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1735,7 +1735,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
 				// Serious race condition not possible here as we set it.
 				if(source == null) {
 				    long delay = randomDelayFinishOpennetLocal();
-				    if(logMINOR) Logger.minor(this, "Delaying opennet completion for "+TimeUtil.formatTime(delay, 2));
+				    if(logMINOR) Logger.minor(this, "Delaying opennet completion for "+TimeUtil.formatTime(delay, 2, true));
 				    node.ticker.queueTimedJob(new Runnable() {
 
                         @Override
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -1681,7 +1681,8 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
 	
 	private long randomDelayFinishOpennetLocal() {
 	    double pingTime = 
-	            Math.max(node.nodeStats.getBwlimitDelayTime(),
+	            // Noderefs are sent as real-time
+	            Math.max(node.nodeStats.getBwlimitDelayTimeRT(),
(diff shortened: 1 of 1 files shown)


toad:
Also, if the delays are too long this might impact on newbie nodes. We might want to adjust the constants, or reconsider whether to include both bwlimitDelayTime and median ping time etc. Currently my big node is showing opennet delays of around 5 seconds.

toad:
More like 2 seconds now with some tweaking, but still looks valid IMHO.

src/freenet/node/RequestSender.java
```diff
@@ -1681,7 +1699,7 @@ void ackOpennet(final PeerNode next) {
      * Add the peer.
      * @return True only if there was a fatal timeout and the caller should not unlock.
      */
-    private boolean finishOpennet(PeerNode next) {
+    private boolean finishOpennet(final PeerNode next) {
```
ArneBab: is this change related to the pull-request?

toad: Yes, you can't access non-final local variables from anonymous inner classes. We need this variable in run() below.

src/freenet/node/RequestSender.java
```diff
@@ -1716,9 +1734,18 @@ private boolean finishOpennet(PeerNode next) {
...
+				    if(logMINOR) Logger.minor(this, "Delaying opennet completion for "+TimeUtil.formatTime(delay, 2, true));
+				    node.ticker.queueTimedJob(new Runnable() {
+
+                        @Override
+                        public void run() {
+                            ackOpennet(next);
+                        }
+				        
+				    }, delay);
+				} else if(origTag.shouldStop()) {
```
toad: This is the important bit: We call ackOpennet() after a computed delay, not immediately.

src/freenet/node/RequestSender.java
ArneBab: ok, thanks for the explanation!

bertm:
I don't think this one and #490 are mutually exclusive, do you @toad?

toad:
This only happens at the originator, i.e. at maximum HTL. So it would be meaningless to have both. We could do the randomisation when we start path folding, but there wouldn't be much point given HTL gives it away anyway so it's not revealing anything useful.

toad:
That is, I don't think we need both. But I think this is less risky than #490.

ArneBab:
Not path folding at high HTL is also useful to keep the self-optimization working if we have to short paths (as we currently do), because that makes the weight of local requests far too big compared to remote requests.

#490 isn’t only for security, but also for improving path folding.

PR #492, opened 2016-01-19
Update default config

ArneBab:
This is just small stuff, but should matter for usability.

### DIFF
--- a/src/freenet/clients/http/wizardsteps/BANDWIDTH_MONTHLY.java
+++ b/src/freenet/clients/http/wizardsteps/BANDWIDTH_MONTHLY.java
@@ -27,7 +27,7 @@ public class BANDWIDTH_MONTHLY extends BandwidthManipulator implements Step {
 	 */
 	private static final Double minCap = 2*Node.getMinimumBandwidth()*secondsPerMonth/GB;
 
-	private static final long[] caps = { (long)Math.ceil(minCap), 50, 100, 150 };
+	private static final long[] caps = { (long)Math.ceil(minCap), 100, 150, 250, 500 };
 
 	public BANDWIDTH_MONTHLY(NodeClientCore core, Config config) {
 		super(core, config);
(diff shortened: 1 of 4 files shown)


src/freenet/clients/http/wizardsteps/BANDWIDTH_MONTHLY.java
```diff
@@ -27,7 +27,7 @@
 	 */
 	private static final Double minCap = 2*Node.getMinimumBandwidth()*secondsPerMonth/GB;
 
-	private static final long[] caps = { (long)Math.ceil(minCap), 50, 100, 150 };
+	private static final long[] caps = { (long)Math.ceil(minCap), 100, 150, 250, 500 };
```
toad: Is it a good idea to get rid of the 50GB limit? What is common? IIRC there is a box to specify a custom limit?

ArneBab: If i did not miscalculate, the min is now 50GiB, that’s why I removed it.

Matthew Toseland writes:

>> @@ -27,7 +27,7 @@
>>  	 */
>>  	private static final Double minCap = 2*Node.getMinimumBandwidth()*secondsPerMonth/GB;
>>  
>> -	private static final long[] caps = { (long)Math.ceil(minCap), 50, 100, 150 };
>> +	private static final long[] caps = { (long)Math.ceil(minCap), 100, 150, 250, 500 };
>
> Is it a good idea to get rid of the 50GB limit? What is common? IIRC there is a box to specify a custom limit?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/freenet/fred/pull/492/files#r50146900


-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

bertm:
ArneBab's calculations are correct. This looks good.

PR #495, opened 2016-01-30
Fix binary blob downloading through FCP (was broken in 1468)

Sadao-:
1) In the ReturnType=disk mode the DataLength field of the DataFound FCP message was always set to 0.
2) In the ReturnType=direct mode the request was stalled at the very beginning due to NPE:

java.lang.NullPointerException
    at freenet.client.async.BinaryBlobWriter.getOutputStream(BinaryBlobWriter.java:74)
    at freenet.client.async.BinaryBlobWriter.addKey(BinaryBlobWriter.java:96)
    at freenet.client.async.ClientGetter.addKeyToBinaryBlob(ClientGetter.java:764)
    at freenet.client.async.SingleFileFetcher.onSuccess(SingleFileFetcher.java:193)
    ...

The patch was tested.

### DIFF
--- a/src/freenet/clients/fcp/ClientGet.java
+++ b/src/freenet/clients/fcp/ClientGet.java
@@ -201,7 +201,10 @@ public class ClientGet extends ClientRequest implements ClientGetCallback, Clien
 			ret = new NullBucket();
 		} else {
 		    targetFile = null;
-		    ret = null; // Let the ClientGetter allocate the Bucket later on.
+			if(binaryBlob)
+				ret = core.clientContext.getBucketFactory(persistence == Persistence.FOREVER).makeBucket(-1);
+			else
+				ret = null; // Let the ClientGetter allocate the Bucket later on.
 		}
 		this.extensionCheck = extensionCheck;
 		this.initialMetadata = null;
@@ -261,7 +264,15 @@ public class ClientGet extends ClientRequest implements ClientGetCallback, Clien
 			ret = new NullBucket();
         } else {
             targetFile = null;
-            ret = null; // Let the ClientGetter allocate the Bucket later on.
+			if(binaryBlob)
+				try {
+					ret = core.clientContext.getBucketFactory(persistence == Persistence.FOREVER).makeBucket(-1);
+				} catch (IOException e) {
+					Logger.error(this, "Cannot create bucket for temp storage: "+e, e);
+					throw new MessageInvalidException(ProtocolErrorMessage.INTERNAL_ERROR, "Cannot create bucket for temporary storage (out of disk space???): "+e, identifier, global);
+				}
+			else
+				ret = null; // Let the ClientGetter allocate the Bucket later on.
 		}
 		this.extensionCheck = extensionCheck;
 		initialMetadata = message.getInitialMetadata();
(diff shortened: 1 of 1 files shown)


Thynix:
Yikes! Thanks for writing this! It sounds like a regression. Do you know when this problem was introduced? How did it manifest? What would be involved in adding a unit test for it?

Sadao-:
The problem was introduced in 16d581ee416f01ce0136cf829bae4aad2f461cf6. Probably Toad forgot about blobs when he changed the code. I discovered the problem when I tried to download a file as blob in Fuqid.

toad:
I thought I had tested that. :(
Thanks!

toad:
Having said that, this looks like the wrong place to fix the problem...

toad:
Why not create the Bucket in makeGetter()? Or does it need to be the same Bucket if we fail and restart? Hmmm...

Sadao-:
> Why not create the Bucket in makeGetter()? Or does it need to be the same Bucket if we fail and restart? Not in this case AFAICS?

I returned the creation of the Bucket to the place where is was before break. I don't think that creating the Bucket for a blob in makeGetter() would be simpler. But if you think it will, feel free to change my patch.

bertm:
This is indeed duplication of logic.

The problem is not where you try to address it, the problem is in `makeGetter` that wrongfully assumes (yet does not assert) that `binaryBlob == true` implies `ret != null`.

Thynix:
Anyone up for fixing the problem at its source in `makeGetter`?

src/freenet/clients/fcp/ClientGet.java
```diff
@@ -344,7 +355,7 @@ public void onLostConnection(ClientContext context) {
 	@Override
 	public void onSuccess(FetchResult result, ClientGetter state) {
 		Logger.minor(this, "Succeeded: "+identifier);
-		Bucket data = result.asBucket();
+		Bucket data = binaryBlob ? state.getBlobBucket() : result.asBucket();
```
Thynix: What's this change intended to do?

Thynix: This is fixing the 0-length DISK get bug?

toad:
Sorry for muddying the waters. Sadao's original patch is the correct way to solve this IMHO. We have to store the binary blob bucket, so we can't create it in makeGetter().

src/freenet/clients/fcp/ClientGet.java
toad: Not just that. It is also needed for ReturnType=direct. We want to return the binary blob, not the fetched data. The FetchResult contains the fetched data, which here will be a NullBucket, not the binary blob.

toad: So this bit is needed even if we create the Bucket in makeGetter().

Thynix:
I've merged #528 which is based on this but avoids the logic duplication by moving it to `makeGetter()`. Thanks for finding this!

src/freenet/clients/fcp/ClientGet.java
xor-freenet: Given that it wasn't clear what this code does, it would be nice if you could add the answer to it as a comment. I've checked the replacement PR, and it doesn't seem to add a comment yet either: https://github.com/freenet/fred/pull/528/files

PR #498, opened 2016-02-05
readme: start cleanup (i.e. stop calling 0.7 new)

ArneBab:


bertm:
I'm all for these changes. Do I understand correctly from your commit message that you want to clean up the readme some more?

I think we should get rid of the following:

* Full caps in section titles
* The entire "Changes from 0.5" section
* The vast majority of the "Basic security" section (Freenet will tell this in the wizard or while being used)

ArneBab:
Yes, I want to clean it up further.

bertm:
Better ☺ Looks good.

PR #499, opened 2016-02-07
BaseL10n: Add support for Hungarian translations

xor-freenet:
Fixes the new Hungarian translation of WoT build00018, which 1471 will deploy.
Thus, please include this in 1471.

The codes & name in the language itself are obtained from the links in
the documentation a few lines above.
The primary "hu" code does match what Transifex uses.


xor-freenet:
Closing in favor of version rebased against next (since this one was against branch stable-1471 which is finished already):
 https://github.com/freenet/fred/pull/530

PR #500, opened 2016-02-07
New language codes for 1472

xor-freenet:
Will be used by the currently being developed WoT build0019.


bertm:
Written as [IETF language tags](https://en.wikipedia.org/wiki/IETF_language_tag) as defined in [RFC 5646](http://tools.ietf.org/html/rfc5646), Portugal Portuguese and Brazil Portuguese would be "pt-PT" and "pt-BR", i.e. "portuguese as used in Portugal"  and "portuguese as used in Brazil" respectively. Note the use of capital letters to indicate region, not variant.

Java's `java.util.Locale` implements support for these language tags.

xor-freenet:
> Written as IETF language tags as defined in RFC 5646, Portugal Portuguese and Brazil Portuguese would be "pt-PT" and "pt-BR", i.e. "portuguese as used in Portugal" and "portuguese as used in Brazil" respectively. Note the use of capital letters to indicate region, not variant.
> 
> Java's java.util.Locale implements support for these language tags.

Thanks for your feedback:
The reasons for my choice of language codes are:

- The existing JavaDoc [says](https://github.com/xor-freenet/fred-staging/blob/c0db284aef50814316103e0ab5e4a015f750a0a9/src/freenet/l10n/BaseL10n.java#L41) the class uses ISO639-2.
- Transifex [says](http://docs.transifex.com/faq/#3-how-do-i-add-my-own-language) about its language codes that they prefer ISO639-* as well:

> These languages adhere to the ISO 639-1 standard of language names and locales. You might fallback to 639-2 or 639-3 if your language is not covered in 639-1. 
> [...]
> If your language isn't in the Languages page and your files appear in a shortened version like ‘xx_YY’, then you can help us providing the necessary information in order to add your language into Transifex.

- So ISO639-2 / -3 would be preferable, as that's what both our class and Transifex recommend. But it doesn't contain the required codes, so we must choose on our own. And then I'd choose the most robust option: To stick to the Transifex style of "xx_YY" to ensure the Transifex command line tool automatically produces filenames which fred looks for. If we used "xx-YY", then developers would have to add manual overrides to the command line tool settings, which they could forget about. If they did forget about it, it wouldn't cause any error messages, the language just wouldn't be available in the UI then. So it is quite a pitfall, and that's why I'd prefer to stick to the Transifex style instead of IETF.

Overall, I agree that this is unfortunate. Transifex should obey standards instead. Perhaps the thing to do is to contact Transifex to ask them to provide a config option for which language code standard to use for our project. I cannot offer to do that unfortunately, i'm quite busy with WoT and fundraising :(

Thynix:
We shouldn't be doing extra stuff in Fred for Transifex. That is the domain of scripts wrapped around Transifex's tools.

xor-freenet:
@Thynix 
I'm willing to comply with improving this, but I have the feeling I didn't sufficiently explain the issues yet to make you notice the core problem, so here's some more simple explanation:

> We shouldn't be doing strange things in Fred for Transifex. 

You're of course right about that. There just is a problem with this concept in practice:
fred's existing JavaDoc says we use ISO639. But in ISO639, there *is no* language code for the "Portuguese as spoken in Portugal" language which this PR wants to add. So we must chose a different standard than the previously chosen one anyway. So if we cannot obey what the JavaDoc wants as standard anyway, why not do so in a way which happens to simplify Transifex usage?
(See my previous comments for more detail about the choices of standards we have)

So I'm OK with whatever standard you chose for fred, be it IETF, Transifex, or whatever; but you need to chose one instead of just saying "we cannot merge this as is", as the current choice of language code standard in the JavaDoc does not supply the language code which I want to add :(

> That is the domain of scripts wrapped around Transifex's tools.

I had used the Transifex scripts you provided some weeks ago. IIRC, they do not automatically deal with mapping mismatching language codes, but rather require the user to manually read the source code of the script and notice that he has to tweak the configuration file of his Transifex tools. I assume this is still the case? So if there is no automation, this can easily be forgotten about, which is why I thought it might be preferable to have fred's codes match out of the box :|

Of course, if you weight this as less important than a proper choice of language code standard in fred, I am willing to comply. Just please tell me which standard you want, otherwise I don't know how to fix this :)

xor-freenet:
Here's the said part of the script, it still requires manual tweaking:
https://github.com/freenet/scripts/blob/a5381f385116af7be2f2d1e3ef53dec16da5ff93/transifex-pull#L9

Also, the list of mismatching codes there would require updating for each new code added to fred :|

Shame on Transifex for not sticking to a standardized language code list >:(

xor-freenet:
Also important to notice here: Some of the existing fred code doesn't adhere to any standard it talks about in the JavaDoc, see the comment which this PR adds as line 70:
https://github.com/freenet/fred/pull/500/commits/c0db284aef50814316103e0ab5e4a015f750a0a9#diff-1267a854c830bc7b7bfd94e44b0bfa36R71

So you may consider the whole existing stuff as something which is broken and thus due to be replaced anyway, and freely chose what you think would be optimal as a replacement for everything.
Perhaps the idea of @bertm of sticking to what Java uses might indeed be the logical choice?

xor-freenet:
@bertm:

> Written as IETF language tags as defined in RFC 5646, Portugal Portuguese and Brazil Portuguese would be "pt-PT" and "pt-BR", i.e. "portuguese as used in Portugal" and "portuguese as used in Brazil" respectively. Note the use of capital letters to indicate region, not variant.
> Java's java.util.Locale implements support for these language tags.

So I've thought a bit about this:
You're right about suggesting this - From a code quality perspective, it would certainly be most clean to use what Java provides, even if it mismatches Transifex. 
We could still make the code which loads the l10n files be graceful with regards to Transifex using "pt_PT" instead of IETF "pt-PT": It could just attempt to fall back to the non-IETF naming if a file using IETF naming is not found. This would be easy, as we can just replace "-" with "_" in generic code, we don't need to deal with individual languages.

Now unfortunately there is a problem left with the Java stuff:
I've write a tool to print and count the Locales which are returned by Locale.getAvailableLocales().
On my Ubuntu, it returns 156 locales. This certainly is incomplete: For example the ISO639-3 language list contains over 7000 entries.
If you want to check this on your system, here's the tool. Notice that the description contains a full command line to clone and execute it:
https://gist.github.com/xor-freenet/483583ab5166fea2a4b21575c2d2a6db 

As a fix, I would propose this:
The Locale class *does* support constructing Locale objects from arbitrary Strings as language code, country code, etc.. I.e. we can feed it with *complete* language code and country lists of our own. These strings are supposed to be chosen from the standards behind the Locale class, most notably ISO639 for the language code.
Luckily, I had once written a ISO639_3.java which does parse a raw ISO639 tab file obtained from the website of the standard. See ```src/freenet/l10n/ISO639_3.java```.
So I would propose we use that existing code to obtain a list of ISO639-3 language codes.
Then we only need to add a list for country codes, which should be much less complex than adding full new code for language codes, as there are much less countries than languages.
Overall, this would fix the most important problem with the existing code:
We shouldn't be manually writing an enum to list languages. We should instead import a full, unmodified official list of a language code standard (or full lists of language codes and country codes in the specific case of the IETF standard).

Thynix:
I say again: making Transifex match the scheme we choose is the realm of
the Transifex configuration under .tx. Please do not consider writing code
in Fred to accommodate its default naming scheme.

xor-freenet:
@Thynix 
> I say again: making Transifex match the scheme we choose is the realm of the Transifex configuration under .tx. Please do not consider writing code in Fred to accommodate its default naming scheme.

Sorry, my previous message wasn't really intended to argue about that, I accept your decision!
I instead mostly aimed at finding a way to replace our manually maintained language list with a full database which doesn't require any manual action to add new languages.

With regards to tx, I guess we could amend the "transifex-pull" script to automatically create mappings for the "-" vs "_" replacement. That should only be a few lines of shell scripting. It would have the code be in the correct place, and still ensure the mapping happens automatically.

Thynix:
Note that `.tx/config` configures mapping between languages and filenames.

xor-freenet:
I've filed a new PR to improve upon this: https://github.com/freenet/fred/pull/540

PR #501, opened 2016-02-07
FProxy configuration: Sort language dropdown alphabetically

xor-freenet:
The toArray(String[]) is used instead of toArray() because the one which
doesn't consume an array cannot be cast to String[], that would cause an
exception.


src/freenet/l10n/BaseL10n.java
```diff
@@ -108,12 +108,19 @@ public static LANGUAGE mapToLanguage(String whatever) {
...
+				if(allValues[i] != UNLISTED)
+					result.add(allValues[i].fullName);
 			}
-
-			return result;
+			
+			Collections.sort(result);
+			result.add(UNLISTED.fullName);
+			
+			return (String[])result.toArray(new String[result.size()]);
```
bertm: There's no need to do the additional `(String[])` cast here, `T[] toArray(T[])` is generic in `T`.

xor-freenet: Thanks! Fixed by dae54019e106e0ee4874c9576e05f8778857fc3e

PR #502, opened 2016-02-21
Make the first-time wizard configurable

Bombe:
This adds configuration items that allow to tell the first-time wizard that certain things should be enabled (or not). The configuration options “loadUPnPPlugin” and “enableAutoUpdater” are added; both default to `true` in order to not change the default behaviour.

I would like to have these new options in Fred for the docker image I am currently working on; as the docker environment doesn’t have access to a home’s router the UPnP plugin does not make any sense. And because the docker image is a 64-bit image we don’t have wrappers anyway so I don’t want the auto-updater to be enabled, either. Both options would be supplied in a skeleton configuration file in the docker image.


Thynix:
Looks good, thanks!

PR #503, opened 2016-02-21
Added missing assigment for the target location of a forwarded Opennet announce request

yadevel:
While working with Freenet, I discovered that whenever a seed node received an 
OpennetAnnounceRequest-message for a target location X, it forwards the 
request to another opennet peer node, but always for target location 0.0 
instead.  This behavior results from the fact that the constructor of the 
AnnounceSender class does not copy the given target location 
into the "target" member variable, which will be fixed by this commit.

### DIFF
--- a/src/freenet/node/AnnounceSender.java
+++ b/src/freenet/node/AnnounceSender.java
@@ -47,7 +47,7 @@ public class AnnounceSender implements PrioRunnable, ByteCounter {
 	private final int paddedLength;
 	private byte[] noderefBuf;
 	private short htl;
-	private double target;
+	private final double target;
 	private final AnnouncementCallback cb;
 	private final PeerNode onlyNode;
 	private int forwardedRefs;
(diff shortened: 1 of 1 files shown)


Thynix:
Thanks for submitting this. I've merged it as 71a788164a896d6e5b6af0dfe0f35da5a6927633 to meet the [commit message standards](https://github.com/freenet/fred/blob/next/CONTRIBUTING.md#commit-messages).

toad:
This should be deployed ASAP IMHO, since it affects bootstrapping performance for new opennet nodes.

Thynix:
I understand that we anticipate this to be an improvement, but in the years
since this regression was deployed it hasn't been critical enough that we
found it. I'm reluctant to include this in the stable build without testing
because of the possibility of revealing other dormant problems, and I am
reluctant to include it with testing because it would further delay an
already significantly delayed release.

PR #504, opened 2016-02-24
Fix synchronization on _matched

toad:
This might have been valid originally but broke because of adding asynchronous filters.

### DIFF
--- a/src/freenet/io/comm/MessageFilter.java
+++ b/src/freenet/io/comm/MessageFilter.java
@@ -43,6 +43,8 @@ public final class MessageFilter {
 
     public static final String VERSION = "$Id: MessageFilter.java,v 1.7 2005/08/25 17:28:19 amphibian Exp $";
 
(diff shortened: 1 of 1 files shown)


bertm:
If `_matched` posed a problem, `_message` is problematic as well. `getMessage` should then also be synchronized on `this`.
(and so should effectively all users of all other member variables be synchronized, although they are not as important as they appear to be of less use to set asynchronously)

toad:
Really this needs refactoring so we use another object and can make stuff final. One serious concern is that it can call callbacks inside methods that ought to have more locking!

nextgens:
Can't we use ThreadLocal for that?

hernic:
@nextgens @bertm Can you validate that this PR is mergeable as is and create another Issue for a potentiel other  target (https://github.com/hyphanet/fred/pull/504#issuecomment-197018555) and for a ThreadLocal solution (https://github.com/hyphanet/fred/pull/504#issuecomment-231556996) ?

PR #505, opened 2016-02-27
better documentation

h4ck3rm1k3:
Better documentation for the bouncy castle, change the defaults to build submodules.

### DIFF
--- a/build-clean.xml
+++ b/build-clean.xml
@@ -182,7 +182,7 @@ maintainers may prefer to use this instead of build.xml.
 	</target>
 
 	<target name="libdep-bc" depends="env" unless="lib.bouncycastle.present">
-		<fail message="Bouncycastle not available (or not recent enough); try https://www.bouncycastle.org/download/${bc.jar}"/>
+	  <fail message="Bouncycastle not available (or not recent enough); try https://www.bouncycastle.org/download/${bc.jar} and put it in the ./lib/ dir"/>
 	</target>
 
 	<target name="libdep-junit" depends="env" unless="lib.junit.present">
--- a/build.properties
+++ b/build.properties
@@ -29,12 +29,12 @@ doc.api=javadoc
 lib.dir = lib
 
 # dir for freenet library jars (aka freenet-ext or "contrib" jars)
-lib.contrib.dir = lib/freenet
+#lib.contrib.dir = lib/freenet
 ## if you use the git submodule in ./contrib (legacy-27 branch)
 ## by running `git submodule init && git submodule update`
 ## contrib.ant is needed for the old legacy layout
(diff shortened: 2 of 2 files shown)


build.properties
```diff
@@ -29,12 +29,12 @@ doc.api=javadoc
 lib.dir = lib
 
 # dir for freenet library jars (aka freenet-ext or "contrib" jars)
-lib.contrib.dir = lib/freenet
+#lib.contrib.dir = lib/freenet
```
Bombe: This looks very unrelated.

contrib
```diff
@@ -1 +1 @@
-Subproject commit ce3b7d55c64126f8fcdc573be3fc37f70feb7ab8
+Subproject commit fdaf5eb562b517bcd040e4c00179cd6cf1f89b4e
```
Bombe: This does, too.

build-clean.xml
```diff
@@ -182,7 +182,7 @@ maintainers may prefer to use this instead of build.xml.
 	</target>
 
 	<target name="libdep-bc" depends="env" unless="lib.bouncycastle.present">
-		<fail message="Bouncycastle not available (or not recent enough); try https://www.bouncycastle.org/download/${bc.jar}"/>
+	  <fail message="Bouncycastle not available (or not recent enough); try https://www.bouncycastle.org/download/${bc.jar} and put it in the ./lib/ dir"/>
```
Bombe: Don’t we have a variable for the directory in which this file is expected? If so, we should use that here.

h4ck3rm1k3: I spent 15 minutes figuring this out, it is not obvious, please let me know if there is a better way.

Bombe: Okay, it’s not super-obvious, unfortunately, but I think `${lib.dir}` would be appropriate.

h4ck3rm1k3: ${lib.dir} was not resolving for me in the ant error message, it was driving me crazy. I found no way to print out the search path being used.

h4ck3rm1k3:
So, there are some other questions I had, 
one being is there any test cases? and secondly why do I need to run the installer  to run, is there no way to run from source?

Thynix:
Yes, there are unit tests. They're run by the default build (`ant`), and can be run explicitly with `ant unit`. It's possible to run from source, but it requires that `freenet-ext.jar` and `bcprov-jdk15on-152.jar` be in the working directory. (See [here](https://wiki.freenetproject.org/Configuring_Freenet#POSIX:_Running_without_the_wrapper).) That will be running without the wrapper, which Freenet uses to upgrade itself. To run that too see a testing script that sets up a node with it [here](https://github.com/freenet/scripts/blob/master/test-autoupgrade#L93).

PR #507, opened 2016-03-01
Remove direct calls to SubConfig constructor

Bombe:
This makes it easier to mock configs and subconfigs once we
get around to unit testing.


bertm:
Looks good to me.

PR #508, opened 2016-03-01
Wait for running transfers to complete when a FNPRouteNotFound-message arrived in the meantime

yadevel:
I encountered this problem while running Freenet in an isolated local testbed, where each opennet announcement ran into a dead end quite quickly. Although latter is not a problem, it resulted in the propagation of a FNPRouteNotFound-message while the transmission of the noderef from a replying node was still running. This commit keeps running noderef-transfers from being aborted in such cases.

### DIFF
--- a/src/freenet/node/AnnounceSender.java
+++ b/src/freenet/node/AnnounceSender.java
@@ -483,16 +483,18 @@ public class AnnounceSender implements PrioRunnable, ByteCounter {
 		if(cb != null) cb.nodeFailed(next, "timed out");
 	}
 
+        private synchronized void waitForRunningTransfers() {
+           while(waitingForTransfers > 0) {
+              try {
+                 wait();
+              } catch (InterruptedException e) {
+                 // Ignore.
+              }
+           }
+        }
+
 	private void rnf(PeerNode next) {
-		synchronized(this) {
-			while(waitingForTransfers > 0) {
-				try {
-					wait();
-				} catch (InterruptedException e) {
-					// Ignore.
-				}
-			}
-		}
+		waitForRunningTransfers();
 		Message msg = DMT.createFNPRouteNotFound(uid, htl);
 		if(source != null) {
 			try {
@@ -508,15 +510,7 @@ public class AnnounceSender implements PrioRunnable, ByteCounter {
 	}
 
 	private void complete() {
-		synchronized(this) {
-			while(waitingForTransfers > 0) {
-				try {
-					wait();
-				} catch (InterruptedException e) {
-					// Ignore.
-				}
-			}
-		}
+		waitForRunningTransfers();
 		Message msg = DMT.createFNPOpennetAnnounceCompleted(uid);
 		if(source != null) {
 			try {
--- a/src/freenet/node/AnnounceSender.java
+++ b/src/freenet/node/AnnounceSender.java
@@ -484,6 +484,15 @@ public class AnnounceSender implements PrioRunnable, ByteCounter {
 	}
 
 	private void rnf(PeerNode next) {
+		synchronized(this) {
+			while(waitingForTransfers > 0) {
+				try {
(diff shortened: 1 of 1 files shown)


src/freenet/node/AnnounceSender.java
```diff
@@ -484,6 +484,15 @@ private void timedOut(PeerNode next) {
 	}
 
 	private void rnf(PeerNode next) {
+		synchronized(this) {
+			while(waitingForTransfers > 0) {
+				try {
+					wait();
+				} catch (InterruptedException e) {
+					// Ignore.
```
Bombe: I’m not a big fan of ignoring exceptions… is this acceptable here?

yadevel: Actually, I just copied this part of code from the complete()-method. Maybe it would be better to put this code into a private method that gets called at the beginning of rnf() and complete(). This method then could also handle exceptions raised during waiting.

Bombe: Deduplication would be nice, indeed. The hard part is coming up with a decent way to handle that exception. :)

yadevel:
Unfortunately, I don't think that I understand Freenet well enough to know what needs to be done if a thread got interrupted while waiting for remaining transfers to finish. If I understood correctly, an InterruptedException occurs only if another thread has called the `interrupt`-method of the AnnounceSender-thread. A typical example where this could happen would be during a shutdown of Freenet. However, this probably means that whenever this exception gets thrown, one needs to check whether it is due to a shutdown, a timeout or any other reason, as it determines whether the thread needs to terminate or if it should just continue or else. A quick grep through Freenet's code revealed that ignoring the InterruptedException seems to be common practice. Maybe this should be addressed in a separate issue report?

src/freenet/node/AnnounceSender.java
yadevel: Unfortunately, I don't think that I understand Freenet well enough to know what needs to be done if a thread got interrupted while waiting for remaining transfers to finish. If I understood correctly, an InterruptedException occurs only if another thread has called the interrupt-method of the AnnounceSender-thread. A typical example where this could happen would be during a shutdown of Freenet. However, this probably means that whenever this exception gets thrown, one needs to check whether it is due to a shutdown, a timeout or any other reason, as it determines whether the thread needs to terminate or if it should just continue or else. A quick grep through Freenet's code revealed that ignoring the InterruptedException seems to be common practice. Maybe this should be addressed in a separate issue report?

bertm: [Bug 6346](https://bugs.freenetproject.org/view.php?id=6346) is related.

yadevel:
With the latest commit, I moved the code into a separate method in order to avoid code duplication.

toad:
Looks good to me. Re InterruptedException, see https://bugs.freenetproject.org/view.php?id=6346 as bertm pointed out, but we don't use interrupt() anywhere AFAIK, and this is a daemon thread so it doesn't *have to* handle shutdown correctly (which is true nearly everywhere in fred).

Thynix:
Thanks! I've merged this as c19df92c46bc4a5639327cb7e36458df0b736b82 so that the commit messages conform to the commit message standards as documented in the [contribution guidelines](https://github.com/freenet/fred/blob/next/CONTRIBUTING.md#commit-messages). See the build [1472 release notes](https://emu.freenetproject.org/pipermail/devl/2016-March/038883.html) for an example of what happens when commit titles are too long - they're included at the end. Compare "use isEmpty instead of getSize in CachingFreenetStoreTracker and push a limited number of keys instead of one push in add" to "Add instructions for installing dependencies on CentOS" - the latter is split over multiple lines which breaks the indentation. That's not to say that commit messages shouldn't include useful explanation - just that it doesn't all belong in the title.

PR #512, opened 2016-03-08
Reduce memory usage of SplitFileInserterStorageTest

bertm:
This changes some large in-memory buffers to on-disk buffers to avoid
OOM unit test failures on Travis CI.


Thynix:
Hooray, thank you!

PR #514, opened 2016-03-15
Avoid claiming of unrejected messages by NodeDispatcher

yadevel:
As a result of [commit f3e3be4](https://github.com/freenet/fred/commit/f3e3be45698272db462ae45f784ff4049c692e92), messages received from an unroutable source may be claimed as matched although they actually weren't rejected or otherwise processed. This behavior led to the failure of noderef-transmissions from seed nodes to a new opennet node, if the first transfer message arrived before the corresponding receiver thread started. This commit ensures that unhandled messages are kept marked as unmatched.

### DIFF
--- a/src/freenet/node/NodeDispatcher.java
+++ b/src/freenet/node/NodeDispatcher.java
@@ -234,6 +234,8 @@ public class NodeDispatcher implements Dispatcher, Runnable {
 				rejectRequest(m, node.nodeStats.sskInsertCtr);
 			} else if(spec == DMT.FNPGetOfferedKey) {
 				rejectRequest(m, node.failureTable.senderCounter);
+			} else {
+				return false;
 			}
 			return true;
 		}


bertm:
This is not the correct solution: returning `false` here fails prematurely.

Consider a `FNPCHKDataRequest` message for a non-routable peer. This message, when passed on to `handleDataRequest` (line 254), would eventually send a reject (see `innerHandleDataRequest`) — the NodeDispatcher should not just ignore the message.

That said, the existing behaviour of `true` for all non-routable sources is obviously worse than your fix. I'm not sure if all subsequent message handlers handle non-routable sources well, but fall-through in case of intermediate no-match seems — in general — more sensible than returning `false`.

yadevel:
As an alternative, one could have a `return true` statement coming directly after each `rejectRequest`-call and omit the unconditional `return true` at the end, so that all messages not matching any of the tested types will be handled the same way as for routable peers. However, looking at [commit 2daf3ca](https://github.com/freenet/fred/commit/2daf3ca1012f9c99b02d34890150b92d05c0266c) and [cb4990e](https://github.com/freenet/fred/commit/cb4990e7d927b7c59ec6db880bea8d5776f89b81) it seems to me as if it is intended that all the subsequent message handlers are avoided if the source is non-routable. In fact, simply returning false if the source is non-routable has been the default for many years. Making the subsequent message handlers process messages from non-routable peers correctly would also be a solution, but this would probably require significantly more effort and also increase their complexity.

bertm:
Oh wait, I completely missed that `FNPCHKDataRequest` is in fact handled for non-routable sources before the return. That renders my point moot.
Thank you for the research though, your observations do indeed seem to confirm your theory and vouch for your changes.

At this point I'm confident that this change fixes a corner-case that was missed in f3e3be4. Looks good to me.

toad:
Yes, this is a major regression which IMHO has also broken Update Over Mandatory. The fix should be deployed ASAP.

toad:
And yes, I'm happy with the fix proposed here.

src/freenet/node/NodeDispatcher.java
```diff
@@ -234,6 +234,8 @@ public boolean handleMessage(Message m) {
 				rejectRequest(m, node.nodeStats.sskInsertCtr);
 			} else if(spec == DMT.FNPGetOfferedKey) {
 				rejectRequest(m, node.failureTable.senderCounter);
+			} else {
+				return false;
 			}
 			return true;
```
ArneBab: would it be equivalent to just change line 240 to `return false`?

toad: No. We want to consume the message if we reject it, but not if it's unrecognised.

toad: See the original commit that caused this regression in the first place (top of the conversation).

ArneBab: ah, yes, I see. Sorry.

Thynix:
Excellent! This works in my testing.

src/freenet/node/NodeDispatcher.java
Thynix: Woah, that's subtle. I missed that too.

yadevel:
I'm glad that I could help you. Thank you for the credits in your latest release note. :-) If you find the time, please also consider merging pull request #508. It addresses another problem that kept Freenet nodes in a local, isolated testbed from connecting to each other. Together with the fixes in my last two pull requests, latter now works like a charm.

ArneBab:
@yadevel Thank you again for the catch!

PR #515, opened 2016-03-16
Simple minor MessageFilter improvements

bertm:
Some *low-hanging fruit* improvements on `MessageFilter`.

### DIFF
--- a/src/freenet/io/comm/MessageFilter.java
+++ b/src/freenet/io/comm/MessageFilter.java
@@ -19,8 +19,7 @@
 
 package freenet.io.comm;
 
-import java.util.HashMap;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 
 import freenet.node.PrioRunnable;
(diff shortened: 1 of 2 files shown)


src/freenet/io/comm/MessageFilter.java
```diff
@@ -46,8 +45,8 @@
     private boolean _matched;
     private PeerContext _droppedConnection;
     private MessageType _type;
-    private final HashMap<String, Object> _fields = new HashMap<String, Object>();
-    private final List<String> _fieldList = new LinkedList<String>();
+    private final List<Object> _fields = new ArrayList<Object>();
```
Bombe: I think a `LinkedHashMap` would be more appropriate here; it avoids some of the costs of a `HashMap` but helps our code to stay cleaner.

bertm: I agree that this is not the cleanest solution, but `LinkedHashMap` is definitely the worst choice in terms of memory efficiency. Note that we actually don't care in which particular order the fields are matched.

If you feel clean code outweighs memory and computational efficiency here, I will replace the `ArrayList`s with a single `HashMap` and use the `entrySet()` to iterate over. Considering this list is traversed for every single packet we receive (with high probability), I'd rather keep the most lightweight approach.

PR #516, opened 2016-03-17
Store updated seednodes.fref in nodeDir, not userDir

bertm:
`seednodes.fred` is read from the `nodeDir` by the `Announcer`, so we should
store it there as well when we fetch the current version.

This should fix [bug 6684](https://bugs.freenetproject.org/view.php?id=6684).

### DIFF
--- a/src/freenet/node/NodeFile.java
+++ b/src/freenet/node/NodeFile.java
@@ -8,38 +8,38 @@ import java.io.File;
 public enum NodeFile {
     Seednodes(InstallDirectory.Node, "seednodes.fref"),
     InstallerWindows(InstallDirectory.Run, "freenet-latest-installer-windows.exe"),
-	InstallerNonWindows(InstallDirectory.Run, "freenet-latest-installer-nonwindows.jar"),
-	IPv4ToCountry(InstallDirectory.Run, "IpToCountry.dat");
+    InstallerNonWindows(InstallDirectory.Run, "freenet-latest-installer-nonwindows.jar"),
+    IPv4ToCountry(InstallDirectory.Run, "IpToCountry.dat");
 
     private final InstallDirectory dir;
     private final String filename;
-    
+
     /**
      * Gets the absolute file path associated with this file for the given node instance.
      */
     public File getFile(Node node) {
         return dir.getDir(node).file(filename);
     }
-    
+
     /**
      * Gets the filename associated with this file.
      */
     public String getFilename() {
         return filename;
     }
-    
+
     /**
      * Gets the base directory with this file for the given node instance.
      */
     public ProgramDirectory getProgramDirectory(Node node) {
         return dir.getDir(node);
     }
-    
+
     private NodeFile(InstallDirectory dir, String filename) {
         this.dir = dir;
         this.filename = filename;
     }
-    
+
     private enum InstallDirectory {
         // node.install.nodeDir
         Node() {
@@ -83,7 +83,7 @@ public enum NodeFile {
                 return node.pluginDir();
             }
         };
-        
+
         abstract ProgramDirectory getDir(Node node);
     }
 }
--- a/src/freenet/node/NodeFile.java
+++ b/src/freenet/node/NodeFile.java
@@ -1,0 +1,89 @@
+package freenet.node;
+
+import java.io.File;
+
+/**
+ * Mapping of files managed by the node to their respective locations.
+ */
+public enum NodeFile {
+    Seednodes(InstallDirectory.Node, "seednodes.fref"),
+    InstallerWindows(InstallDirectory.Run, "freenet-latest-installer-windows.exe"),
+	InstallerNonWindows(InstallDirectory.Run, "freenet-latest-installer-nonwindows.jar"),
+	IPv4ToCountry(InstallDirectory.Run, "IpToCountry.dat");
+
+    private final InstallDirectory dir;
+    private final String filename;
+    
+    /**
+     * Gets the absolute file path associated with this file for the given node instance.
+     */
+    public File getFile(Node node) {
+        return dir.getDir(node).file(filename);
+    }
+    
+    /**
+     * Gets the filename associated with this file.
+     */
+    public String getFilename() {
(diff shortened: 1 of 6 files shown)


src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -502,17 +509,17 @@ public void start() throws InvalidConfigValueException {
...
 		enable(wasEnabledOnStartup);
 
-		// Fetch 3 files, each to a file in the runDir.
-
+		// Fetch seednodes to the nodeDir.
 		if (updateSeednodes) {
 
 			SimplePuller seedrefsGetter = new SimplePuller(getSeednodesURI(),
-					Announcer.SEEDNODES_FILENAME);
+					Announcer.SEEDNODES_FILENAME, node.nodeDir());
```
toad: Should this be directory.dir()?

toad:
Looks good.

Thynix:
Nice find! Two things:

1. The commit body misspells `seednodes.fref` as `seednodes.fred`.
2. Would it be reasonable to have a complete path to the seed nodes file that is used both for reading and writing? This way will work but it doesn't make it more robust.

bertm:
You latter concern is addressed in 87d7fc4. The first can be taken care of when merging, or I can force-update / file another pull request once we agree on the contents of the pull request.

src/freenet/node/NodeFile.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.File;
+
+/**
+ * Mapping of files managed by the node to their respective locations.
+ */
+public enum NodeFile {
+    Seednodes(InstallDirectory.Node, "seednodes.fref"),
+    InstallerWindows(InstallDirectory.Run, "freenet-latest-installer-windows.exe"),
+	InstallerNonWindows(InstallDirectory.Run, "freenet-latest-installer-nonwindows.jar"),
```
Thynix: These lines use tabs but the rest of this file uses spaces.

src/freenet/node/NodeFile.java
```diff
@@ -0,0 +1,89 @@
...
+	IPv4ToCountry(InstallDirectory.Run, "IpToCountry.dat");
+
```
Thynix: Trailing whitespace here and elsewhere.

src/freenet/node/NodeFile.java
```diff
@@ -0,0 +1,89 @@
...
+    public ProgramDirectory getProgramDirectory(Node node) {
+        return dir.getDir(node);
+    }
+    
+    private NodeFile(InstallDirectory dir, String filename) {
+        this.dir = dir;
+        this.filename = filename;
+    }
+    
+    private enum InstallDirectory {
```
Thynix: Looks like it's redundant to mark an `enum` constructor private?

> The constructor for an enum type must be package-private or private access. It automatically creates the constants that are defined at the beginning of the enum body. You cannot invoke an enum constructor yourself. 

-- [Enum Types](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)

src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -404,7 +412,7 @@ public void onSuccess(FetchResult result, ClientGetter state) {
 			File temp;
 			FileOutputStream fos = null;
 			try {
-				temp = File.createTempFile(filename, ".tmp", node.getRunDir());
+				temp = File.createTempFile(filename, ".tmp", directory.dir());
```
Thynix: Is it an intended behavior change to use directories other than the run directory for the temporary file? What's the reasoning for that? I was going to ask if that would cause problems for packaged versions, but packaged versions are updated by the package system anyway so that's not a problem.

src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -481,12 +485,12 @@ public FreenetURI getSeednodesURI() {
 				"seednodes-" + Version.buildNumber());
 	}
 
-	public FreenetURI getInstallerWindowsURI() {
+	public FreenetURI getInstallerNonWindowsURI() {
```
Thynix: Heh. Hello bugfix.

Thynix:
Nothing else comes to mind. Thanks! Do you want me to make the final fixups and merge?

src/freenet/node/updater/NodeUpdateManager.java
bertm: It appears that the idea behind this line was that the temp files are to be
created in the same directory as the files to overwrite, to fail early when
there are permission errors.
Op 25 mrt. 2016 02:29 schreef "Steve Dougherty" <notifications@github.com>:

src/freenet/node/updater/NodeUpdateManager.java
bertm: Not a bug fix, there was no bug other than weird inconsistent naming of
methods and constants ;)
Still seemed reasonable to address that. Should have mentioned it in the
commit, though.
Op 25 mrt. 2016 02:36 schreef "Steve Dougherty" <notifications@github.com>:

bertm:
If you have the time, feel free to make the final fixups. Otherwise I'll
make them this weekend.

Thank you for reviewing!
Op 25 mrt. 2016 02:54 schreef "Steve Dougherty" <notifications@github.com>:

src/freenet/node/updater/NodeUpdateManager.java
Thynix: Ah, okay. That's better then.

Thynix:
Okay, sounds good! I don't yet know if I'll have time before the weekend.

src/freenet/node/NodeFile.java
bertm: Note that this is not the constructor, this is the enum declaration. I'll keep it private until we find another use for it (then it will probably need to move to another file as well).

bertm:
Closing in favor of identical #522 that squashes the changes and fixups and fixes the commit message.

PR #517, opened 2016-03-18
Fix peer node network traffic statistics

Ademan:
PeerNode previously failed to increment in/out traffic statistics
since startup, and did not initialize the total counters to include
the counts from prior Freenet sessions. The counters intended to
count the in/out traffic statistics since startup were initialized
to the total count and never incremented. This means that the two
sets of counters were reversed from their named purpose. This patch
fixes this and changes the total counters into totals at startup time.
The all time total traffic can be calculated as the sum of the traffic
since startup and the traffic before startup. No external code is aware
of this internal bookkeeping change, but it eliminates the need to
actively maintain both counters during runtime.

### DIFF
--- a/src/freenet/node/PeerNode.java
+++ b/src/freenet/node/PeerNode.java
@@ -339,9 +339,9 @@ public abstract class PeerNode implements USKRetrieverCallback, BasePeerNode, Pe
 	private TimeDecayingRunningAverage pRejected;
 
 	/** Bytes received at/before startup */
-	private long bytesInAtStartup;
+	private final long bytesInAtStartup;
 	/** Bytes sent at/before startup */
-	private long bytesOutAtStartup;
+	private final long bytesOutAtStartup;
 
 	/** Times had routable connection when checked */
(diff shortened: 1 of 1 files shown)


src/freenet/node/PeerNode.java
```diff
@@ -337,10 +337,12 @@
...
 	private TimeDecayingRunningAverage pRejected;
-	/** Total low-level input bytes */
-	private long totalBytesIn;
-	/** Total low-level output bytes */
-	private long totalBytesOut;
+
+	/** Bytes received at/before startup */
+	private long bytesInAtStartup;
+	/** Bytes sent at/before startup */
+	private long bytesOutAtStartup;
```
bertm: Bytes at startup sounds like a fixed value. These should be `final`.

Ademan: Good point, agreed. I probably won't update until ~tomorrow when I
hopefully will have revived my desktop.

toad:
Looks fine to me, I don't see how this would break anything.

PR #518, opened 2016-03-19
Make seednodes.fref parsing more robust

bertm:
Always log errors when we fail to parse the seednodes file. If we fail
on one entry, now read to the end of the line and proceed parsing
instead of bailing out prematurely.

### DIFF
--- a/src/freenet/node/Announcer.java
+++ b/src/freenet/node/Announcer.java
@@ -277,9 +277,15 @@ public class Announcer {
 						list.add(fs);
 				} catch (EOFException e) {
 					return list;
+				} catch (IOException e) {
+					Logger.error(Announcer.class, "Error while reading seednodes from " + file, e);
+					// Continue reading. If this entry failed, we still want the following noderefs.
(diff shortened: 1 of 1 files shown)


src/freenet/node/Announcer.java
```diff
@@ -277,9 +277,15 @@ private synchronized int connectSomeNodesInner(List<SimpleFieldSet> seeds) {
 						list.add(fs);
 				} catch (EOFException e) {
 					return list;
+				} catch (IOException e) {
+					Logger.error(Announcer.class, "Error while reading seednodes from " + file, e);
+					// Continue reading. If this entry failed, we still want the following noderefs.
+					// However, read a line to prevent us getting stuck in an endless loop.
```
Thynix: Looks good. Maybe be more specific?

```
// Continue reading. If this entry failed, we still want the following noderefs.
// Read a line to advance the parsing position and avoid an endless loop.
```

bertm: Agreed, fixed ☺

PR #520, opened 2016-03-20
Update the tray (and possibly the wrapper) on Windows

toad:
This seems to work but has one serious bug:
If the node takes too long to shut down, it may still be shutting down when the tray starts back up. If so, the tray won't restart it, and the node will stay offline until the user reboots (or logs out and in) or until they click on the rabbit.

So for nodes which take a long time to shut down - that is, nodes with plugins! - this could be a problem. On the other hand, users of such nodes are more likely to be active users, click on the rabbit regularly, and complain if it doesn't work.

Fixing this properly requires some means of *waiting until the node has actually finished shutting down*. The FreenetTray.exe does not appear to provide this AFAICS.


toad:
Also, note that this won't work unless FreenetTray.exe is in the nodeDir(). On old nodes upgraded via update.cmd, it might be in bin/. On really old nodes that haven't been upgraded it might not exist at all.

toad:
Note that it uses relative filenames so hopefully it will work on non-English installs in spite of problems with charsets and batch files.

src/freenet/node/updater/MainJarDependenciesChecker.java
```diff
@@ -1232,6 +1233,20 @@ boolean deploy() {
...
         
+        /** Temp file will actually be used after restarting, so we need it to not be deleted. 
+         * @throws IOException */
+        File moveTempFile() throws IOException {
+            // Best that these NOT be absolute, want relative filenames for windows updater script.
+            File parent = filename.getParentFile();
+            if(parent == null) parent = new File(".");
+            File newTemp = File.createTempFile(filename.getName(), ".new.tmp", parent);
+            if(File.separatorChar == '\\')
+                newTemp.delete(); // FIXME Needed for Windows, no actual symlink race here AFAICS.
```
bertm: We have `FileUtil.detectedOS.isWindows` to show whether we are on Windows, no need for `separatorChar` magic.

src/freenet/node/updater/MainJarDependenciesChecker.java
```diff
@@ -1576,6 +1601,97 @@ private boolean createRunShNoNice(File input, File output) throws IOException {
...
+            restartScript.delete();
+            FileBucket fb = new FileBucket(restartScript, false, true, false, false);
+            OutputStream os = null;
+            try {
+                os = new BufferedOutputStream(fb.getOutputStream());
+                OutputStreamWriter osw = new OutputStreamWriter(os); // Use default locale.
+                osw.write("@echo off\r\n");
+                osw.write("echo Starting wrapper update script > update-script.log\r\n");
+                
+                osw.write("ping -n 1 127.0.0.1 >nul\r\n");
```
bertm: Please document that this hack is intended to pause for one second, and has nothing to do with networks at all.
`// There's no such thing as a "sleep" command. Use "ping" to wait instead.` would do.

toad:
This would work more reliably if we wait for the wrapper to exit.
<nextgens> toad_> the standard windows way of waiting for a process to exit is to loop
<nextgens> toad_> a la https://stackoverflow.com/questions/8177695/how-to-wait-for-a-process-to-terminate-to-execute-another-process-in-batch-file

Thynix:
-exit is not documented as waiting until the node has stopped before
exiting the tray. Removing the anchor file means the node will shut down,
which is sufficient. If you want to change that behavior I could do that,
too.

Thynix:
Given the choice between changing the tray and implementing waiting for the wrapper in batch script I think changing the tray makes more sense. I'll do that.

src/freenet/node/updater/MainJarDependenciesChecker.java
Thynix: Why is this sleep here?

src/freenet/node/updater/MainJarDependenciesChecker.java
```diff
@@ -1576,6 +1601,97 @@ private boolean createRunShNoNice(File input, File output) throws IOException {
...
+        }
+        
+        private File createRestartScript() throws IOException {
+            File restartScript = new File(deployer.getNodeDir(), RESTART_SCRIPT_NAME);
+            restartScript.delete();
+            FileBucket fb = new FileBucket(restartScript, false, true, false, false);
+            OutputStream os = null;
+            try {
+                os = new BufferedOutputStream(fb.getOutputStream());
+                OutputStreamWriter osw = new OutputStreamWriter(os); // Use default locale.
```
Thynix: I'd really prefer that this script be a resource file - perhaps in two parts if we really need to generate which files to move - instead of writing lots of individual lines. Thoughts?

toad: We can get rid of a lot of the lines especially with the tray changes. Some of this is necessarily generated... But I guess we could move it if you want...

toad:
Improvements addressing bertm's issues pushed. Also it will use the new FreenetTray.exe rather than the old one if it is updating FreenetTray.exe, which should be sufficient to make waiting for shutdown reliable given operhiem1's planned changes. However, the latest changes have not been tested yet.

Thynix:
This doesn't include a 64-bit wrapper binary? I can test parts of the tray without one, but not the entire setup.

toad:
Yes, I tested it by replacing the FreenetTray.exe with an older version of the executable (in the installer history I think). Then it should update it to the latest version, which is included in the dependencies.properties. This allows updating FreenetTray.exe. It should also allow deploying the new wrapper exe's (put them in the same deployment group with the mustExist flag set to false). IIRC when I was working on this I didn't have a 64-bit wrapper binary, somebody else was working on that.

toad:
The plan was to deploy the 64-bit wrapper dll and the wrapper exe separately from the 32-bit versions, and then have a new FreenetTray that uses the appropriate wrapper exe.

Thynix:
Right. The tray changes are implemented but not yet tested and pushed. My
understanding is we do not yet have 64-bit binaries to work with. Is my
recollection correct that @unixninja92 built them?

unixninja92:
I have but it's based on code I downloaded from a random blog. I was going to recreate it and build it myself, but life has gotten really busy recently and haven't had a chance to yet.

ArneBab:
@Thynix is this still up to date or can it be closed?

src/freenet/node/updater/MainJarDependenciesChecker.java
ArneBab: Maybe to ensure that the file is actually accessible in the OS?

ArneBab:
We now have a 64-bit wrapper (thanks to work by @Thynix in 2018 that got it built). The changes here look like they could be important. Can these be finished so they are ready to merge?

hernic:
@unixninja92 @Thynix any comment ?

PR #521, opened 2016-03-22
build: add IntelliJ IDEA configuration

Thynix:
IntelliJ IDEA does not currently support completely disabling wildcard
imports, as suggested by the project code style. Setting a large number
of classes seems the next best thing.

See https://youtrack.jetbrains.com/issue/IDEA-124296


Bombe:
Just for the record: I am opposed to putting local IDE configuration under source control. In my opinion this simply does not belong here, and I will seriously hate it if somebody else keeps fucking up _my_ _local_ IDE configuration.

Thynix:
This does exclude some configuration Jetbrains documents as always local,
but good point. Would committing just the code style configuration be
viable?

nextgens:
Where are we at with the gradle integration? iirc it can produce/manage the IDE/code-style configs

xor-freenet:
Conceptual ACK: We don't have enough developers, so we need to make things easy for them to get more.
If an IDE overwrites its own configuration with a project specific one, that is a bug of the IDE, not our fault. Sorry Bombe. If you don't want the config, you can remove the config files and add them to the ".gitignore" file so you don't get them again.

("Conceptual" ACK = I'm not qualified to review an IntelliJ configuration, so I didn't read the actual code)

Bombe:
> If you don't want the config, you can remove the config files and add them to the ".gitignore" file so you don't get them again.

Yeah, that’s not how Git works…

xor-freenet:
> Yeah, that’s not how Git works…

You're right, I'm sorry.
Here's the proper way to get rid of the config files in your local copy of the repository (from the manpage of gitignore):

```
The purpose of gitignore files is to ensure that certain files not tracked by Git remain untracked.

To ignore uncommitted changes in a file that is already tracked, use git update-index --assume-unchanged.

To stop tracking a file that is currently tracked, use git rm --cached.
```

nextgens:
The gradle branch has been merged; please convert it to 
https://docs.gradle.org/current/userguide/idea_plugin.html
We'll get rid of the existing IDE config stuff in the repository

PR #522, opened 2016-03-27
Store updated seednodes.fref in nodeDir, not userDir

bertm:
seednodes.fref is read from the nodeDir by the Announcer, so we should
store it there as well when we fetch the current version.

This should fix bug 6684.

To prevent future confusion, this adds a central registry for node-
managed files and their locations. This initially contains only the
locations of the files updated by NodeUpdateManager, but may later
be reused for other purposes.


Thynix:
Thanks!

PR #523, opened 2016-03-31
Remove load limiting high level token buckets

toad:
This removes a bunch of old load management complexity that is barely used in practice.


toad:
I should probably get rid of the first commit? The problem was that the buckets start empty, this was causing problems in simulations because we can't send/accept requests immediately. It might be a small issue in reality for real nodes too. Obviously removing the whole mechanism obsoletes it.

Other reasons for this: Bandwidth claimed varies a lot, output bandwidth liability is nearly always dominant in practice, and it doesn't help much when things aren't working well because of no stats yet (e.g. very new nodes). So there's no good reason for it.

PR #528, opened 2016-04-10
Fix fcp blob downloading

Thynix:
This is based on #495 and creates the bucket in makeGetter() instead of at two of its callers. It works in my testing - I get hangs when fetching blobs on 1472 where it only writes what seems to be a header to disk. With these changes applied I get a much larger blob back.

    00000000  6d 58 24 9f 72 d6 7e d9  00 00 00 00 00 00 00 02  |mX$.r.~.........|
    00000010  00 00                                             |..|
    00000012


toad:
Looks good to me.

Thynix:
Thanks for reviewing!

PR #532, opened 2016-05-03
Replace our DSA implementation by BouncyCastle's

nextgens:
This hasn't been tested thoroughly yet but should address the concerns of Eleriseth. Seems to work for me.

As a side bonus, it gets rid of NativeBigIntegers... yes, it'll be slower than what we currently have but makes deploying the next freenet-ext way simpler.


src/freenet/keys/SSKBlock.java
```diff
@@ -137,22 +138,32 @@ public SSKBlock(byte[] data, byte[] headers, NodeSSK nodeKey, boolean dontVerify
...
-			NativeBigInteger r = new NativeBigInteger(1, bufR);
-			NativeBigInteger s = new NativeBigInteger(1, bufS);
-			if(!(DSA.verify(pubKey, new DSASignature(r, s), new NativeBigInteger(1, overallHash), false) ||
-					(DSA.verify(pubKey, new DSASignature(r, s), new NativeBigInteger(1, overallHash), true)))) {
+			BigInteger r = new BigInteger(1, bufR);
+			BigInteger s = new BigInteger(1, bufS);
+			DSASigner dsa = new DSASigner();
+			dsa.init(false, new DSAPublicKeyParameters(pubKey.getY(), Global.getDSAgroupBigAParameters()));
+
+			// We probably don't need to try both here... but that's what the legacy code was doing.
```
nextgens: Q is invariant in our case (we always use the same group)... and from what I can tell we always truncate... whereas I think that we don't have to... do we want to keep the same behaviour? The madness has started in 567bf11a954edc31f3b6d4348782792fe7d5bae5 ; Do we care about the content that was inserted before? If we don't we might be able to safe half the cpu cycles here.

For the record, we don't need "m < q"!
It's the opposite: "N must be less than or equal to the output length of the hash H." according to https://en.wikipedia.org/wiki/Digital_Signature_Algorithm

Bombe:
I am generally in favor of removing code but can’t really comment on the crypto side of this. :)

PR #534, opened 2016-05-15
Add password confirmation prompt for disk crypto.

nathanielnebel:
This pull request aims to resolve issue #0006745 (https://bugs.freenetproject.org/view.php?id=6745) which requires the user to confirm their password when performing an upgrade (change password or setting a master password). Apologies for not signing this commit against my GPG key.

### DIFF
--- a/src/freenet/node/Version.java
+++ b/src/freenet/node/Version.java
@@ -50,7 +50,7 @@ public class Version {
 	public static final String protocolVersion = "1.0";
 
 	/** The build number of the current revision */
-	private static final int buildNumber = 1471;
+	private static final int buildNumber = 1472;
 
 	/** Oldest build of fred we will talk to *before* _cal */
 	private static final int oldLastGoodBuild = 1465;
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -79,7 +79,7 @@ public class NodeUpdateManager {
 	 * The last build on the old key with Java 6 support. Older nodes can
 	 * update to this point via old UOM.
 	 */
-	public final static int TRANSITION_VERSION = 1471;
+	public final static int TRANSITION_VERSION = 1472;
 
 	/** The URI for post-TRANSITION_VERSION builds' freenet.jar. */
 	public final static String UPDATE_URI = "USK@O~UmMwTeDcyDIW-NsobFBoEicdQcogw7yrLO2H-sJ5Y,JVU4L7m9mNppkd21UNOCzRHKuiTucd6Ldw8vylBOe5o,AQACAAE/jar/"
@@ -223,7 +223,8 @@ public class NodeUpdateManager {
 
 		// Set default update URI for new nodes depending on JVM version.
 		updaterConfig
-				.register("URI", JVMVersion.isTooOld() ? LEGACY_UPDATE_URI : UPDATE_URI, 3, true, true,
+				.register("URI", JVMVersion.isTooOld() ? transitionMainJarURIAsUSK.toString() : UPDATE_URI,
+				          3, true, true,
 						"NodeUpdateManager.updateURI",
 						"NodeUpdateManager.updateURILong",
 						new UpdateURICallback());


src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -115,7 +115,9 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
 				if(newPhysicalLevel == oldPhysicalLevel && newPhysicalLevel == PHYSICAL_THREAT_LEVEL.HIGH) {
 					String password = request.getPartAsStringFailsafe("masterPassword", MAX_PASSWORD_LENGTH);
 					String oldPassword = request.getPartAsStringFailsafe("oldPassword", MAX_PASSWORD_LENGTH);
-					if(password != null && oldPassword != null && password.length() > 0 && oldPassword.length() > 0) {
+					String confirmPassword = request.getPartAsStringFailsafe("confirmMasterPassword", MAX_PASSWORD_LENGTH);
+					boolean passwordsMatch = password.equals(confirmPassword);
```
Thynix: This uses `password` before the check for `null` in the next line. The use should come after the `null` check; having `password.equals(confirmPassword)` inline seems a fine option to me.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -568,10 +572,26 @@ private void drawSecurityLevelsPage(HTMLNode contentNode, ToadletContext ctx) {
 				if(physicalLevel == level) {
 					addPasswordChangeForm(inner);
 				} else {
-					// Add password form
-					p = inner.addChild("p");
-					p.addChild("label", "for", "passwordBox", l10nSec("setPassword"));
-					p.addChild("input", new String[] { "id", "type", "name" }, new String[] { "passwordBox", "password", "masterPassword" });
+					p = inner.addChild("p", l10nSec("setPassword"));
+
```
Thynix: You've added trailing whitespace to this line; please try to avoid that.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -568,10 +572,26 @@ private void drawSecurityLevelsPage(HTMLNode contentNode, ToadletContext ctx) {
...
+					HTMLNode table = inner.addChild("table", "border", "0");
+		            HTMLNode row = table.addChild("tr");
+		            HTMLNode cell = row.addChild("td");
+		            cell.addChild("label", "for", "passwordBox", l10nSec("passwordLabel"));
+		            cell = row.addChild("td");
+		            cell.addChild("input",
+		                    new String[] { "id", "type", "name", "size" },
+		                    new String[] { "passwordBox", "password", "masterPassword", "100" });
+		            row = table.addChild("tr");
+
```
Thynix: This one too.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -698,10 +726,33 @@ public static void generatePasswordFormPage(boolean wasWrong, HTMLNode formNode,
...
+        
+		    row = table.addChild("tr");
+		    cell = row.addChild("td");
+		    cell.addChild("label", "for", "confirmPasswordBox", l10nSec("confirmPasswordLabel"));
+		    cell = row.addChild("td");
+		    cell.addChild("input",
+		            new String[] { "id", "type", "name", "size" },
+		            new String[] { "confirmPasswordBox", "password", "confirmMasterPassword", "100"});
+		}
+		if(!forUpgrade){
```
Thynix: This would be less surprising as an `else` to the `if`.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -698,10 +726,33 @@ public static void generatePasswordFormPage(boolean wasWrong, HTMLNode formNode,
 			content.addChild("#", l10nSec("enterPassword"));
 		}
 
-		formNode.addChild("input",
-		        new String[] { "type", "name", "size" },
-		        new String[] { "password", "masterPassword", "100" });
-
+		//Creates a table for password prompt and the confirmation box.
+		if(forUpgrade){
```
Thynix: The [style guide](https://google.github.io/styleguide/javaguide.html) (the link to which from the [contribution guidelines](https://github.com/freenet/fred/blob/next/CONTRIBUTING.md) was broken) suggests a space between the `if` and the open parenthesis.

Thynix:
Ooh, this is a good feature. Thanks for opening this!

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -144,7 +146,9 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
 					if(newPhysicalLevel == PHYSICAL_THREAT_LEVEL.HIGH && node.securityLevels.getPhysicalThreatLevel() != newPhysicalLevel) {
 						// Check for password
 						String password = request.getPartAsStringFailsafe("masterPassword", MAX_PASSWORD_LENGTH);
-						if(password != null && password.length() > 0) {
+						String confirmPassword = request.getPartAsStringFailsafe("confirmMasterPassword", MAX_PASSWORD_LENGTH);
+						boolean passwordsMatch = password.equals(confirmPassword);
+						if(password != null && confirmPassword != null && password.length() > 0 && confirmPassword.length() > 0 && passwordsMatch) {
```
Thynix: `!password.isEmpty()` seems more semantic than a length check; same sort of thing for the `password.length() == 0`.

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1783,14 +1783,18 @@ SecurityLevels.enterPasswordTitle=Enter the master password to access cached con
 SecurityLevels.passwordWrong=You entered the wrong password. Please enter the correct password to enable access to the client-cache and improve browsing performance.
 SecurityLevels.enterPassword=Please enter the master password to enable access to downloads and uploads and to improve performance by enabling the recently browsed freesites cache.
 SecurityLevels.setPasswordLabel=Set the password
+SecurityLevels.confirmPasswordLabel=Confirm password
+SecurityLevels.passwordLabel=Password
 SecurityLevels.setPassword=Please set the master password, which will be used to protect cached content. The stronger the password the stronger the protection: lots of letters and numbers and punctuation is best.
 SecurityLevels.passwordNotZeroLength=The password must be at least one character long i.e. it must not be "".
+SecurityLevels.passwordsDoNotMatch=The passwords must match and they must be at least one character long i.e. it must not be "".
```
Thynix: Why include the first message in the second?

nathanielnebel:
Thank you for the feedback! I will make the suggested changes and push it to my repo so it is included in the pull request.

src/freenet/l10n/freenet.l10n.en.properties
nathanielnebel: Because the second message is used in cases where password confirmation is occurring and the passwords can be zero length or not match, hence extending the first message that is already used.

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: While not an excuse that guideline does not appear to be respected at all here, I will fix all occurrences of this.

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: Makes sense, don't know how I overlooked that.

(note to self: fixed)

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: Fixed

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: fixed

nathanielnebel:
I have made the suggested changes to the code

src/freenet/clients/http/SecurityLevelsToadlet.java
Bombe: I’m pretty sure that `getPartAsStringFailsafe` always returns a non-`null` value…

nathanielnebel: According to the documentation provided it does return an empty string if the "part" does not exist, it did probably return null sometime in the past but that no longer holds true.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -83,6 +84,10 @@ public void getStep(HTTPRequest request, PageHelper helper) {
 				HTMLNode p = div.addChild("p");
 				p.addChild("label", "for", "passwordBox", WizardL10n.l10nSec("setPasswordLabel")+":");
 				p.addChild("input", new String[] { "id", "type", "name" }, new String[] { "passwordBox", "password", "masterPassword" });
+
```
Thynix: Trailing whitespace here.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -698,19 +725,42 @@ public static void generatePasswordFormPage(boolean wasWrong, HTMLNode formNode,
...
+		    cell = row.addChild("td");
+		    cell.addChild("input",
+		            new String[] { "id", "type", "name", "size" },
+		            new String[] { "confirmPasswordBox", "password", "confirmMasterPassword", "100"});
+		}
+		else {
+		    formNode.addChild("label", "for", "passwordBox", l10nSec("passwordLabel"));
+		    formNode.addChild("input",
+		            new String[] { "id", "type", "name", "size" },
+		            new String[] { "passwordBox", "password", "masterPassword", "100" });
```
Thynix: When it was a form with a single field it was less of a deal to duplicate it, but now that it's a table with multiple we should avoid duplicating it. Can you please move generating this to a function - probably one that takes `formNode`?

Thynix:
For future reference, unless the pull request is entirely formatting changes, please only change formatting in lines you're already otherwise changing. Doing otherwise makes review more difficult.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -202,7 +217,10 @@ public String postStep(HTTPRequest request) throws IOException {
 			if (passwordIsBlank) {
 				// Must set the password to something non-blank.
 				return promptPassword(newThreatLevel, PASSWORD_PROMPT.SET_BLANK);
-			} else {
+			} else if(passwordsDoNotMatch){
```
Thynix: Whitespace problems here.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -202,7 +217,10 @@ public String postStep(HTTPRequest request) throws IOException {
...
+			    //Must Confirm the password before setting it
+			    return promptPassword(newThreatLevel, PASSWORD_PROMPT.SET_NO_MATCH);
+			}else {
```
Thynix: Whitespace problems here too.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -182,8 +195,10 @@ private boolean errorHandler(HTTPRequest request, PageHelper helper) {
 	public String postStep(HTTPRequest request) throws IOException {
 		final String errorCorrupt = FirstTimeWizardToadlet.WIZARD_STEP.SECURITY_PHYSICAL+"&error=corrupt";
 		String pass = request.getPartAsStringFailsafe("masterPassword", SecurityLevelsToadlet.MAX_PASSWORD_LENGTH);
-		final boolean passwordIsBlank = pass != null && pass.length() == 0;
-
+		String confirmPass = request.getPartAsStringFailsafe("confirmMasterPassword", SecurityLevelsToadlet.MAX_PASSWORD_LENGTH);
+		final boolean passwordIsBlank = pass != null && pass.length() == 0 && confirmPass != null && confirmPass.length() == 0;
+		final boolean passwordsDoNotMatch = !pass.equals(confirmPass);
```
Thynix: The `pass != null` check in the line above implies `pass` can be null, but it can't be. The existing code acted like it can, but it can't. (We could clean that up in existing code in another pull request.) [`getPartAsStringFailsafe`](https://github.com/freenet/fred/blob/c9bcbb88dc28034e4dafda583958f8612b89c0ac/src/freenet/clients/http/HTTPRequestImpl.java#L683) prevents it - even in the case of an encoding problem in [`getPartAsLimitedString`](https://github.com/freenet/fred/blob/c9bcbb88dc28034e4dafda583958f8612b89c0ac/src/freenet/clients/http/HTTPRequestImpl.java#L691) where the behavior of the String constructor [is undefined](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String%28byte[],%20java.lang.String%29), my understanding of this part of [the Java specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.9.4) is that new instance creation cannot return null, so there is no need to check return values of `getPartAsStringFailsafe` against null - at most it will return an empty string.

I find booleans are often easier to think about when they're the positive case - with `passwordsDoNotMatch` it does have immediate use in the thing you're writing right now, but in the future if we want to do something if they match we could find ourselves writing `!passwordsDoNotMatch`. `!passwordsMatch` and `passwordsMatch` are easier to use and understand.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -182,8 +195,10 @@ private boolean errorHandler(HTTPRequest request, PageHelper helper) {
 	public String postStep(HTTPRequest request) throws IOException {
 		final String errorCorrupt = FirstTimeWizardToadlet.WIZARD_STEP.SECURITY_PHYSICAL+"&error=corrupt";
 		String pass = request.getPartAsStringFailsafe("masterPassword", SecurityLevelsToadlet.MAX_PASSWORD_LENGTH);
-		final boolean passwordIsBlank = pass != null && pass.length() == 0;
-
+		String confirmPass = request.getPartAsStringFailsafe("confirmMasterPassword", SecurityLevelsToadlet.MAX_PASSWORD_LENGTH);
+		final boolean passwordIsBlank = pass != null && pass.length() == 0 && confirmPass != null && confirmPass.length() == 0;
```
Thynix: The `pass != null` check in the line above implies `pass` can be null, but it can't be. The existing code acted like it can, but it can't. (We could clean that up in existing code in another pull request.) [`getPartAsStringFailsafe`](https://github.com/freenet/fred/blob/c9bcbb88dc28034e4dafda583958f8612b89c0ac/src/freenet/clients/http/HTTPRequestImpl.java#L683) prevents it - even in the case of an encoding problem in [`getPartAsLimitedString`](https://github.com/freenet/fred/blob/c9bcbb88dc28034e4dafda583958f8612b89c0ac/src/freenet/clients/http/HTTPRequestImpl.java#L691) where the behavior of the String constructor [is undefined](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String%28byte[],%20java.lang.String%29), my understanding of this part of [the Java specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.9.4) is that new instance creation cannot return null, so there is no need to check return values of `getPartAsStringFailsafe` against null - at most it will return an empty string.

src/freenet/l10n/freenet.l10n.en.properties
Thynix: Oh, okay. That works.

src/freenet/clients/http/wizardsteps/SECURITY_PHYSICAL.java
```diff
@@ -152,6 +163,8 @@ private boolean errorHandler(HTTPRequest request, PageHelper helper) {
 
 			if (type == PASSWORD_PROMPT.SET_BLANK || type == PASSWORD_PROMPT.DECRYPT_BLANK) {
 				content.addChild("p", WizardL10n.l10nSec("passwordNotZeroLength"));
+			} else if(type == PASSWORD_PROMPT.SET_NO_MATCH){
```
Thynix: Whitespace problems here.

Thynix:
Once these things are resolved I think this will be ready to merge.

PR #537, opened 2016-05-22
Set rel='noreferrer noopener' where appropriate

nextgens:


src/freenet/clients/http/PproxyToadlet.java
```diff
@@ -533,7 +533,7 @@ private void showPluginList(ToadletContext ctx, PluginManager pm, HTMLNode conte
 					pluginRow.addChild("td");
 				} else {
 					if (pi.isPproxyPlugin()) {
-						HTMLNode visitForm = pluginRow.addChild("td").addChild("form", new String[] { "method", "action", "target" }, new String[] { "get", pi.getPluginClassName(), "_blank" });
+						HTMLNode visitForm = pluginRow.addChild("td").addChild("form", new String[] { "method", "action", "target", "rel" }, new String[] { "get", pi.getPluginClassName(), "_blank", "noreferrer noopener" });
```
ArneBab: Do I understand it correctly that this leaves out the referrer when accessing a plugin page from a freesite?

If yes: Why do you consider that as problematic?

nextgens: It's the second parameter that matters... noreferrer is just there to deal with older browsers

ArneBab: Florent Daigniere writes:

>> @@ -533,7 +533,7 @@ private void showPluginList(ToadletContext ctx, PluginManager pm, HTMLNode conte
>>  					pluginRow.addChild("td");
>>  				} else {
>>  					if (pi.isPproxyPlugin()) {
>> -						HTMLNode visitForm = pluginRow.addChild("td").addChild("form", new String[] { "method", "action", "target" }, new String[] { "get", pi.getPluginClassName(), "_blank" });
>> +						HTMLNode visitForm = pluginRow.addChild("td").addChild("form", new String[] { "method", "action", "target", "rel" }, new String[] { "get", pi.getPluginClassName(), "_blank", "noreferrer noopener" });
>
> It's the second parameter that matters... noreferrer is just there to deal with older browsers

I looked up noopener now, and you’re right.

Thank you!

PR #538, opened 2016-05-22
Truecrypt is now Veracrypt

nextgens:


toad:
I thought the decision was not to explicitly recommend Truecrypt any more? I believe the wizard links to the Wikipedia list of disk encryption software - so if we're recommending TC in the security levels page, this is a bug?

nextgens:
Feel free to send a better PR :)

ArneBab:
I think this bugfix is good. No change in behaviour, just a fix for a no longer valid string.

nextgens:
#444 is doing the same

PR #539, opened 2016-05-22
FIX: on too low bandwidth, use min bandwidth

ArneBab:
I think this should have been the behaviour all along to allow us to upgrade the network to new internet connection requirements without continuously killing old nodes.

I cannot just re-use the checking function, because that must throw an exception for the config settings.

### DIFF
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -1582,6 +1582,10 @@ public class Node implements TimeSkewDetectorCallback {
 		});
 
 		int obwLimit = nodeConfig.getInt("outputBandwidthLimit");
+		if (obwLimit < minimumBandwidth) {
+			obwLimit = minimumBandwidth; // upgrade slow nodes automatically
+		}
+
 		outputBandwidthLimit = obwLimit;
 		try {
 			checkOutputBandwidthLimit(outputBandwidthLimit);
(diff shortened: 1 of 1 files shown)


Thynix:
What changed the bandwidth limit?

src/freenet/node/Node.java
```diff
@@ -1582,6 +1582,10 @@ public void set(Integer obwLimit) throws InvalidConfigValueException {
 		});
 
 		int obwLimit = nodeConfig.getInt("outputBandwidthLimit");
+		if (obwLimit < minimumBandwidth) {
+			obwLimit = minimumBandwidth; // upgrade slow nodes automatically
```
Thynix: Thoughts on logging about clamping the values?

ArneBab: that would be good, yes.

ArneBab:
@Thynix I fear that *I* changed the bandwidth limit, because the existing limit did not actually suffice to keep the connections (I tried). I did not expect Freenet to just not start when the value is lower than the minimum value (seems no one expected that we might want to raise that) — the bug report took me completely by surprise…

src/freenet/node/Node.java
ArneBab: it logs now

ArneBab:
Stronger reason: Look at the connection number stats. The peak at 11 connections goes much too far to 5 (and below). More than I expect from the randomization. That means too many people don’t manage to keep the minimum of 10 connections.

Thynix:
What commit changed the bandwidth limit?

ArneBab:
This one: https://github.com/ArneBab/fred-staging-1/commit/565e7fcaea564ed5b50c3403e0a207cc2cd01d8e

Thynix:
Oh. Whoops. I misremembered the minimum as like 12 KiB/s already.

ArneBab:
thank you!

PR #541, opened 2016-05-26
Deprecate functions of class Fields in favor of Java API

xor-freenet:
Please review carefully:
I'm not qualified to perfectly judge whether the replacements behave sufficiently similar.
This especially applies to commits b2fdf98cb8322fc8578c0d0ee3d94ce23ce5238e and cacf24e30ceb5d0b45073d3bd50eb899b1d6f05d.

### DIFF
--- a/src/freenet/support/Fields.java
+++ b/src/freenet/support/Fields.java
@@ -425,9 +425,7 @@ public abstract class Fields {
 			return compareBytes(o1, o2);
 		}
 	}
-
-	// could add stuff like IntegerComparator, LongComparator etc.
-	// if we need it
+	
 	public static int hashCode(byte[] b) {
 		return hashCode(b, 0, b.length);
 	}
--- a/src/freenet/support/Fields.java
+++ b/src/freenet/support/Fields.java
@@ -391,6 +391,8 @@ public abstract class Fields {
 		return 0;
 	}
 
+	/** @deprecated Use {@link Arrays#equals(byte[], byte[])} */
+	@Deprecated
 	public static boolean byteArrayEqual(byte[] a, byte[] b) {
 		if(a.length != b.length)
 			return false;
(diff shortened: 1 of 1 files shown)


src/freenet/support/Fields.java
```diff
@@ -924,21 +934,33 @@ public static int compareObjectID(Object o1, Object o2) {
...
 		return 0;
 	}
 
-	/** Avoid issues with NaN's. */
+	/**
+	 * Avoid issues with NaN's.
+	 * @deprecated
+	 *     Use {@link Double#compare(double, double)}. Notice: It may behave differently with
+	 *     regards to special values (e.g. -0.0). Check its source code if you need precisely
+	 *     the same behavior. */
```
bertm: We really can't use this kind of comments. This should either document what is different exactly. If the differences are too great, this method serves a good purpose of preventing duplication of logic and should not be deprecated.

That said, this method is blatantly inconsistent with `Double.compare` on negative integers and `NaN`.

xor-freenet: Thanks! You're absolutely right, this is dissatisfying.

I have to say sorry - In the PR, I should have pointed at the the commit message to spare you from having to write this comment, it already hinted at why I did this:
I'm not a floating point expert. So I thought before I make false claims in the JavaDoc, I better leave it vague; and instead show myself to be willing to improve it from review results in case a floating point expert shows up! :)

I aimed at the following options for resolving this:
- I intentionally committed the deprecation of different types of code (hash codes, integer stuff, float stuff) in separate commits. This is to allow you to cherry-pick the good ones and leave stuff such as this behind if you don't think it's worth the hassle.
- If someone already is a float expert and can briefly give us a hint at the behavioral differences, I'm willing to write them into a longer JavaDoc and commit it.
- I can offer to try fulfilling my duty of figuring this out myself! May take some weeks though, lots of nice stuff is waiting to be finished (WoT release!), I don't want to get stuck too much on this :)

nextgens:
I like the idea here... but as @bertm said, it's just not ready. What do we do with it?

PR #543, opened 2016-05-27
Add password confirmation to disk crypto.

nathanielnebel:
This pull request aims to resolve issue #0006745 (https://bugs.freenetproject.org/view.php?id=6745) which requires the user to confirm their password when performing an upgrade (change password or setting a master password). Apologies for not signing this commit against my GPG key. Sorry for opening a new pull request for this issue, I tried to squash all of the changes into one commit and it messed up my repo and made it so I was unable to reopen the original pull request.


Thynix:
Unfortunately I will be largely unavailable for probably the next month or so for medical reasons, so I won't be able to review this as early as I'd like.

@ArneBab @nextgens Are you up for reviewing this?

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -168,32 +170,32 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
...
 							return;
 						}
 					}
-					if((newPhysicalLevel == PHYSICAL_THREAT_LEVEL.LOW || newPhysicalLevel == PHYSICAL_THREAT_LEVEL.NORMAL) &&
+					if ((newPhysicalLevel == PHYSICAL_THREAT_LEVEL.LOW || newPhysicalLevel == PHYSICAL_THREAT_LEVEL.NORMAL) &&
 							oldPhysicalLevel == PHYSICAL_THREAT_LEVEL.HIGH) {
 						// Check for password
 						String password = request.getPartAsStringFailsafe("masterPassword", SecurityLevelsToadlet.MAX_PASSWORD_LENGTH);
-						if(password != null && password.length() > 0) {
+						if (password != null && password.length() > 0) {
```
bertm: `password` cannot be `null` here as per the specification of `HttpRequest.getPartAsStringFailsafe`

bertm:
Reviewing this is hugely frustrating due to all the unrelated changes to whitespace.

Also, the whitespace changes belong in their own commit as they are completely unrelated to the commit message of 51e540b. As it is now, `git blame` would show "Add password confirmation to disk crypto." as justification of unrelated changes all over the place.

nathanielnebel:
@bertm Sorry about that, Thynix already told me to avoid doing that in the future.

ArneBab:
@bertm couldn’t you diff -w the commits? (ignore all whitespace)

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: This was one of the style changes that I should have put into another commit, I have removed the unneeded null check though.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -110,43 +110,45 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
...
 							return;
 						} catch (AlreadySetPasswordException e) {
 							sendChangePasswordForm(ctx, false, true, newPhysicalLevel.name());
-							if(changedAnything)
+							if (changedAnything)
 								core.storeConfig();
 							return;
 						}
-					} else if(password != null || oldPassword != null) {
+					} else if (password != null || oldPassword != null) {
```
nathanielnebel: Unneeded nullcheck, remove this.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -250,7 +252,7 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
 									l10nSec("passwordForDecryptTitle"), contentNode, "password-prompt", false).
 									addChild("div", "class", "infobox-content");
 
-							if(password != null && password.length() == 0) {
+							if (password != null && password.length() == 0) {
```
nathanielnebel: Unneeded nullcheck, please remove.

nathanielnebel:
If you guys see any more errors that I or you have not made a line-note for please let me know and I will fix it.

bertm:
FYI: with all changes unstaged, `git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero` adds the non-whitespace changes to the index but keeps the whitespace-only changes unstaged.

I strongly suggest you to use a separate branch for each pull request by the way, currently you're working in next and that will eventually lead to a mess when you try to pull other changes.

src/freenet/clients/http/SecurityLevelsToadlet.java
```diff
@@ -686,6 +693,25 @@ public static void generatePasswordFormPage(boolean wasWrong, ToadletContainer c
 		HTMLNode form = ctx.addFormChild(content, postTo, "masterPasswordForm");
 		generatePasswordFormPage(wasWrong, form, content, forDowngrade, forUpgrade, physicalSecurityLevel, redirect);
 	}
+	
+	public static void generatePasswordConfirmationForm(HTMLNode formNode) {
```
bertm: Why is this method public and static?

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1783,14 +1783,18 @@ SecurityLevels.enterPasswordTitle=Enter the master password to access cached con
 SecurityLevels.passwordWrong=You entered the wrong password. Please enter the correct password to enable access to the client-cache and improve browsing performance.
 SecurityLevels.enterPassword=Please enter the master password to enable access to downloads and uploads and to improve performance by enabling the recently browsed freesites cache.
 SecurityLevels.setPasswordLabel=Set the password
+SecurityLevels.confirmPasswordLabel=Confirm password
+SecurityLevels.passwordLabel=Password
 SecurityLevels.setPassword=Please set the master password, which will be used to protect cached content. The stronger the password the stronger the protection: lots of letters and numbers and punctuation is best.
 SecurityLevels.passwordNotZeroLength=The password must be at least one character long i.e. it must not be "".
+SecurityLevels.passwordsDoNotMatch=The passwords must match and they must be at least one character long i.e. it must not be "".
```
bertm: The non-matching password error message does not have to include the non-empty password requirement: at least one of the passwords entered (either password or confirmation) was non-empty after all!

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: I made it static because the method doesn't need to be specific to that instance of the class, public on the other hand is a good question. I considered making it private since it is only used in that class but I saw that methods that provide similar functionality were also made public. I will make it private though as that does make way more sense.

src/freenet/l10n/freenet.l10n.en.properties
bertm: While in general it is a great idea to tell the user what would be expected of them (such as the rules passwords have to adhere to), in this case we want to point out a user error that does not need further explanation. We'd better tell the user what went wrong instead, something like "Password doesn't match confirmation"

nathanielnebel: I talked to Thynix about this and said that the message passwordsDoNotMatch was specific to cases where there was both prompting and confirmation and Xor said that more specific error messages were preferred. The message on line 1789 is specific to cases where it is only prompting for the password and the only possible error is zero length passwords.

src/freenet/clients/http/SecurityLevelsToadlet.java
bertm: There's no need to use this in a static context, it is only used (and to be used) by instances of `SecurityLevelsToadlet`. It's generally a good idea to reduce the exposure of any piece of code wherever possible.

src/freenet/l10n/freenet.l10n.en.properties
bertm: What I'm trying to say is the following: by the time we display this message to the user, we know that he has entered a non-empty password into at least one of the inputs (two empty passwords are equal after all), so he already demonstrated to be aware of the latter requirement.

I'm fine with a plain "The passwords must match" if you prefer to address the user this way. @xor-freenet any comments on this?

bertm:
What I'd like to see is the following:

* Have the functional and whitespace-only changes as separate commits.
* Use `.isEmpty()` instead of comparing `.length()` with zero.
* The indentation now uses mixed spaces and tabs. The surrounding code uses tabs AFAICS, please follow that convention.
* Address the other code quality comments from above.

The functional changes look fine. That said, I haven't tested it yet and should take another look when it's still before midnight sometime soon.

src/freenet/clients/http/SecurityLevelsToadlet.java
nathanielnebel: @bertm A lot of the methods in this class that draw components are static and therefor can only call other static methods. I have changed it to private but I was wondering if you would like me to change all of the other methods so they are not static?

src/freenet/l10n/freenet.l10n.en.properties
nathanielnebel: @bertm Just tested it with the new error message, and it worked as you just said. No need to include the password requirements.

src/freenet/clients/http/SecurityLevelsToadlet.java
bertm: There's no need for that, it's a style issue (that does not affect readability), not a defect.

Don't fix it if it ain't broken, it will be addressed once that code sees other changes.

PR #544, opened 2016-05-28
Eclipse config: Fix compiler Java version to match Ant builder

xor-freenet:
= Set to Java 7.

The date stamp was removed by Eclipse automatically. I think that makes sense, the file was modified more recently now.


bertm:
Looks good to me

PR #545, opened 2016-05-28
l10n: avoid suggesting tracing is impossible

Thynix:


bertm:
Looks good to me.

ArneBab:
looks good to me

PR #548, opened 2016-05-30
Add confirmation prompt to disk crypto.

nathanielnebel:
This pull request aims to resolve issue #0006745 (https://bugs.freenetproject.org/view.php?id=6745) which requires the user to confirm their password when performing an upgrade (change password or setting a master password). Apologies for making this any messier than it needed to be with the pull requests and all, I thought I could reopen the one I made last night before I closed it because I noticed an issue with one section that I failed to notice when testing the changes. Thank you for taking the time to look over this and I hope after all of the valuable feedback that it is ready to be merged.


nextgens:
This looks good. Thanks!

PR #549, opened 2016-05-31
Hotfix for Frostbite

bertm:
This adds additional sanity checks and makes the KeyListener code more resilient to malicious splitfiles.


bertm:
An improved version of 0eb5960 can be found in c15b24d, but the changes therein are more involved and need review.

PR #551, opened 2016-06-01
fproxy-page: sanitize the content of the key input field

ArneBab:
With this change, the fetch key field can be used with an anchor as well as with links which use the default host and port. That allows telling users to just put a Freenet link into that fetch field (which also avoids attacks from putting a key into the url field of the browser).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 next ():
 
+- fproxy: fetch key form: preserve anchor and kill host:port
+- fix monthly bandwidth selection bug
+- fix breakage from increased min bandwidth 
+- prevent denial-of-service via misconstructed manifests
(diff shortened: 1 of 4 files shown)


bertm:
9bf078b should not be part of this pull request.

ArneBab:
@bertm including that wasn’t intentional.

ArneBab:
Can you estimate the effort for switching to a POST request? I likely won’t have much time to work on this the next weeks.

ArneBab:
I think you’re right ☹

Bombe would say that I have to differentiate between easy solutions and too easy solutions which break things.

nextgens:
I've implemented it differently in #551 please review :)

PR #553, opened 2016-06-05
Ant builder: Analyze test coverage and code complexity using Cobertura

xor-freenet:
(Please consider including the subject and description of this PR as body of the merge commit message.)

Cobertura is a tool which injects its own code during the unit tests to analyze which code of the fred core is called during test runs. This yields HTML output which contains things such as:
- a view of the source code with the lines being color-coded as: red = untested, green = tested. 
- a percentage of how many lines of code are covered by the tests.
- a percentage of how many code paths (if() etc) are covered by the tests.
- bonus static analysis: the "average McCabe's cyclomatic code complexity for all methods." = average number of code paths. This can be used to find giant functions which need to be split up.
- the above statistics both for the whole repository and also for each package and class.

This branch amends the Ant builder to optionally use Cobertura and provide the HTML output at `./test-coverage/html`

As dependency you only need to install the standard Ubuntu "cobertura" package. Alternatively, put a cobertura.jar into `./lib`
It can then be executed by:
`ant clean && ant -Dtest.coverage=true`
Incremental compilation, i.e. leaving out the `ant clean`, should also work as long as you don't delete classes.
You can also compute test coverage as caused by only a specific single unit test class:
`ant -Dtest.coverage=true -Dtest.class=freenet.packageName.ClassName`

Full HTML output, accessible with Freenet:
http://127.0.0.1:8888/CHK@ukoL~NGt6t0p9VRJ7RATO0vxlGVr0mBvL0qB164A4Y8,DbjfwdzSP7hDN11-s20k79tRURobCRtyZyVGppVH1ZY,AAMC--8/test-coverage-fred_testing-build-1474-pre1-122-g5baa142/index.html

Clearnet-accesible screenshot of the overall statistics for the whole repository: http://i.imgur.com/xTuZwBv.png

For review purposes, consult the Cobertura Ant documentation at https://github.com/cobertura/cobertura/wiki/Ant-Task-Reference

### DIFF
--- a/build-clean.xml
+++ b/build-clean.xml
@@ -222,6 +222,8 @@ maintainers may prefer to use this instead of build.xml.
 		<mkdir dir="${main.dst}"/>
 		<mkdir dir="${test.make}"/>
 		<mkdir dir="${test.dst}"/>
+		<mkdir dir="${test.coverage.make}"/>
+		<mkdir dir="${test.coverage.dst}"/>
 	</target>
 
 	<target name="dep" depends="ensure-ext, ensure-bc, ensure-gjs"/>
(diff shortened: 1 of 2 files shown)


nextgens:
This is deprecated by #590 

If you want code coverage analysis, do it with gradle.

PR #554, opened 2016-06-05
Fix bug #6488: visit freesite looses the anchor

nextgens:
The better version of #551


src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -265,6 +265,18 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
             writeHTMLReply(ctx, 200, "OK", pageNode.generate());
             request.freeParts();
             bucket.free();
+            return;
+        } else if (request.isPartSet("key")) {
+            if(!ctx.checkFormPassword(request)) return;
+            String key;
+            try {
+              key = URLDecoder.decode(new FreenetURI(request.getPartAsStringFailsafe("key", 256)).toURI("/").toString(), false);
```
bertm: FreenetURI imposes a limit of 32767 (= Short.MAX_VALUE) to the key's length. Any reason to limit it to less than that?

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -265,6 +265,18 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
...
+            } catch (Exception e) {
+              sendErrorPage(ctx, "Invalid URI", "This URI is invalid!", e);
```
bertm: This need l10n

nextgens: I know that we've translated error messages in the past... but I'm not sure it's a good idea; do you insist on this?

src/freenet/clients/http/WelcomeToadlet.java
nextgens: I've copied/doubled what was above... but fair enough, we don't need a limit here

src/freenet/clients/http/WelcomeToadlet.java
bertm: I don't care much for the more obscure (internal) error messages, but this one can easily be triggered by the user mistyping stuff. This is user interface stuff after all.

I won't object to having this merged without being l10n'ed, but if you find the time, please do so.

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -265,6 +265,18 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
             writeHTMLReply(ctx, 200, "OK", pageNode.generate());
             request.freeParts();
             bucket.free();
+            return;
+        } else if (request.isPartSet("key")) {
+            if(!ctx.checkFormPassword(request)) return;
+            String key;
+            try {
+              key = URLDecoder.decode(new FreenetURI(request.getPartAsStringFailsafe("key", Short.MAX_VALUE)).toURI("/").toString(), false);
```
ArneBab: could you split this line? (it’s longer than the visible width in Github)

PR #558, opened 2016-06-13
Warn users that java1.7 is EOL

nextgens:
Odds are we won't require java8 before it's in debian stable...
and it won't be before stretch is released (freeze is scheduled for 02/2017)


nextgens:
14:00 <@Bombe> nextgens, we should have started nagging them about that a long 
               time ago. :)

PR #559, opened 2016-06-13
adjust bw limit: min peers 4 and min bandwidth 5KiB/s (more risky option)

ArneBab:
This is one of two options I see for reducing the required peer count per bandwidth.

I see it as important, because current versions of Freenet have difficulty keeping more than 70 connections, so users with high bandwidth have flapping connections, which will stop convergence of the network.

This is the more risky option: reducing the minimum bandwidth back down to 5KiB and allowing the peers to go down to 4 — one long connection and 3 short ones. With this change, very slow nodes should contribute to routing roughly as much (or as little) as they did with 10 connections before the link length fix.

On the other hand, with this change, slow nodes should easily be able to keep all their connections.

### DIFF
--- a/src/freenet/clients/http/wizardsteps/BANDWIDTH_MONTHLY.java
+++ b/src/freenet/clients/http/wizardsteps/BANDWIDTH_MONTHLY.java
@@ -27,7 +27,7 @@ public class BANDWIDTH_MONTHLY extends BandwidthManipulator implements Step {
 	 */
 	private static final Double minCap = 2*Node.getMinimumBandwidth()*secondsPerMonth/GB;
 
-	private static final long[] caps = { (long)Math.ceil(minCap), 100, 150, 250, 500 };
+	private static final long[] caps = { (long)Math.ceil(minCap), 50, 100, 150, 250, 500 };
 
 	public BANDWIDTH_MONTHLY(NodeClientCore core, Config config) {
 		super(core, config);
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -782,11 +782,11 @@ public class Node implements TimeSkewDetectorCallback {
 	private boolean enableRoutedPing;
 
 	/**
-	 * Minimum bandwidth limit in bytes considered usable: 5 KiB. If there is an attempt to set a limit below this -
+	 * Minimum bandwidth limit in bytes considered usable: 4 KiB. If there is an attempt to set a limit below this -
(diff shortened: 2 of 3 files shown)


bertm:
> I see it as important, because current versions of Freenet have difficulty keeping more than 70 connections, so users with high bandwidth have flapping connections, which will stop convergence of the network.

Are you confident that this observation is not just a side-effect forced disconnects due to Frostbite and the recent mandatory update?

ArneBab:
I’m pretty confident about that, yes, because the fraction of peers with 80 connections increased. I might still be wrong, though.

This is the old distribution: `http://127.0.0.1:8888/SSK@pxtehd-TmfJwyNUAW2Clk4pwv7Nshyg21NNfXcqzFv4,LTjcTWqvsq3ju6pMGe9Cqb3scvQgECG81hRdgj5WO4s,AQACAAE/statistics-1226/plot_peer_count.png`

This is the current one:

xor-freenet:
@ArneBab 
> This is the old distribution: http://127.0.0.1:8888/SSK@pxtehd-TmfJwyNUAW2Clk4pwv7Nshyg21NNfXcqzFv4,LTjcTWqvsq3ju6pMGe9Cqb3scvQgECG81hRdgj5WO4s,AQACAAE/statistics-1226/plot_peer_count.png

Can you please edit your comment to include an actual copy of the image? Github allows that and it will be more convenient for reviewers & cannot fall out of the network.

> This is the current one:

Did you intentionally post an empty graph or was this a mistake?

ArneBab:

> @ArneBab 
>> This is the old distribution: http://127.0.0.1:8888/SSK@pxtehd-TmfJwyNUAW2Clk4pwv7Nshyg21NNfXcqzFv4,LTjcTWqvsq3ju6pMGe9Cqb3scvQgECG81hRdgj5WO4s,AQACAAE/statistics-1226/plot_peer_count.png
>
> Can you please edit your comment to include an actual copy of the image? Github allows that and it will be more convenient for reviewers & cannot fall out of the network.
>
>> This is the current one:
>
> Did you intentionally post an empty graph or was this a mistake?

The graph is from Steve’s site and he changed the stats collection to
not include that.

I can replace both, yes.
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

ArneBab:
This should actually be less risky than I thought, because existing users should all have at least 6 connections (with the old minimum of 10KiB/s upload).

PR #560, opened 2016-06-13
adjust bandwidth: min peers 7 and min bandwidth 7KiB/s (safer option)

ArneBab:
This is one of two options I see for reducing the required peer count per bandwidth.

I see it as important, because current versions of Freenet have difficulty keeping more than 70 connections, so users with high bandwidth have flapping connections, which will stop convergence of the network.

This is the less risky option: reducing the minimum bandwidth back down to 7KiB and allowing the peers to go down to 7 — 2 long connections and 5 short ones. With this change, very slow nodes should still contribute to routing — for long, medium (<0.01), and for short distances. 

They might have problems to keep these connections (as do 10KiB/s nodes with 10 connections).


ArneBab:
Closed because min peers 4 is now tried and tested and works well.

PR #564, opened 2016-06-21
Reduce algorithmic complexity of FOAF routing

bertm:
This design works under the assumption that location doubles can be compared by equality; something that is not necessarily true for doubles in general. It can be modified to work without this assumption, at the expense of some additional computational complexity.

This drops the algorithmic complexity of finding the next peer to route to from O(n² m) to O(n log n + nm), where *n* is the number of peers (assuming our peers have the same number of peers as we have), and *m* is the number of peers that are excluded from routing towards.

Despite its simplicity, this needs careful review and testing. ~~Oh, and did I mention that it still needs a unit test at least for the algorithmic parts?~~ Breaking routing means breaking Freenet at its core.
*In vivo* this seems to work fine, but breakage will probably only be perceivable if the majority of nodes breaks.

For review, not for merge yet.

### DIFF
--- a/src/freenet/node/PeerLocation.java
+++ b/src/freenet/node/PeerLocation.java
@@ -1,8 +1,6 @@
 package freenet.node;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.Arrays;
 import java.util.Set;
 
 import freenet.support.Logger;
@@ -11,8 +9,9 @@ public class PeerLocation {
 	
 	/** Current location in the keyspace, or -1 if it is unknown */
 	private double currentLocation;
-	/** Current sorted unmodifiable list of locations of our peer's peers */
-	private List<Double> currentPeersLocation;
+	/** Current sorted array of locations of our peer's peers. Must not be modified,
+	 may only be replaced entirely. */
+	private double[] currentPeersLocation;
 	/** Time the location was set */
 	private long locSetTime;
 
@@ -39,8 +38,9 @@ public class PeerLocation {
 		return currentLocation;
 	}
 
-	synchronized List<Double> getPeerLocations() {
-		return currentPeersLocation;
+	/** Returns an array copy of locations of our peer's peers. */
+	public synchronized double[] getPeersLocationArray() {
+		return Arrays.copyOf(currentPeersLocation, currentPeersLocation.length);
 	}
 
 	public synchronized long getLocationSetTime() {
@@ -53,7 +53,7 @@ public class PeerLocation {
 
 	public synchronized int getDegree() {
 		if (currentPeersLocation == null) return 0;
-		return currentPeersLocation.size();
+		return currentPeersLocation.length;
 	}
 
 	boolean updateLocation(double newLoc, double[] newLocs) {
@@ -63,17 +63,18 @@ public class PeerLocation {
 			return false;
 		}
 
-		ArrayList<Double> newPeersLocation = new ArrayList<Double>(newLocs.length);
-		for (double loc : newLocs) {
+		final double[] newPeersLocation = new double[newLocs.length];
+		for (int i = 0; i < newLocs.length; i++) {
+			final double loc = newLocs[i];
 			if (!Location.isValid(loc)) {
 				Logger.error(this, "Invalid location update for " + this + " (" + loc + ")", new Exception("error"));
 				// Ignore it
 				return false;
 			}
-			newPeersLocation.add(loc);
+			newPeersLocation[i] = loc;
 		}
 		
-		Collections.sort(newPeersLocation);
+		Arrays.sort(newPeersLocation);
 		boolean anythingChanged = false;
 
 		synchronized (this) {
@@ -81,18 +82,18 @@ public class PeerLocation {
 				anythingChanged = true;
 			}
 			if (!anythingChanged) {
-				anythingChanged = currentPeersLocation.size() != newPeersLocation.size();
+				anythingChanged = currentPeersLocation.length != newPeersLocation.length;
 			}
 			if (!anythingChanged) {
-				for (int i = 0; i < currentPeersLocation.size(); i++) {
-					if (!Location.equals(currentPeersLocation.get(i), newPeersLocation.get(i))) {
+				for (int i = 0; i < currentPeersLocation.length; i++) {
+					if (!Location.equals(currentPeersLocation[i], newPeersLocation[i])) {
 						anythingChanged = true;
 						break;
 					}
 				}
 			}
 			currentLocation = newLoc;
-			currentPeersLocation = Collections.unmodifiableList(newPeersLocation);
+			currentPeersLocation = newPeersLocation;
 			locSetTime = System.currentTimeMillis();
 		}
 		return anythingChanged;
@@ -112,16 +113,16 @@ public class PeerLocation {
 	 * or -1 if none.
 	 * This is a binary search of logarithmic complexity.
 	 */
-	private static int findFirstGreater(List<Double> elems, double x) {
+	private static int findFirstGreater(final double[] elems, final double x) {
 		int low = 0;
-		int high = elems.size();
+		int high = elems.length;
 		int mid;
-		if (elems.get(high - 1) <= x) {
+		if (elems[high - 1] <= x) {
 			return -1;
 		}
 		while (low != high) {
 			mid = low + (high - low) / 2;
-			if (elems.get(mid) > x) {
+			if (elems[mid] > x) {
 				high = mid;
 			} else {
 				low = mid + 1;
@@ -134,9 +135,9 @@ public class PeerLocation {
 	 * Finds the position of the closest location in the sorted list of locations.
 	 * This is a binary search of logarithmic complexity.
 	 */
-	static int findClosestLocation(List<Double> locs, double l) {
-		assert(locs.size() > 0);
-		if (locs.size() == 1) {
+	static int findClosestLocation(final double[] locs, final double l) {
+		assert(locs.length > 0);
+		if (locs.length == 1) {
 			return 0;
 		}
 		final int firstGreater = findFirstGreater(locs, l);
@@ -145,13 +146,13 @@ public class PeerLocation {
 		if (firstGreater == -1 || firstGreater == 0) {
 			// All locations are greater, or all locations are smaller.
 			// Closest location must be either smallest or greatest location.
-			left = locs.size() - 1;
+			left = locs.length - 1;
 			right = 0;
 		} else {
 			left = firstGreater - 1;
 			right = firstGreater;
 		}
-		if (Location.distance(l, locs.get(left)) <= Location.distance(l, locs.get(right))) {
+		if (Location.distance(l, locs[left]) <= Location.distance(l, locs[right])) {
 			return left;
 		}
 		return right;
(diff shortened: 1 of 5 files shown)


bertm:
Now that I have added unit tests (and fixed the bugs discovered with those) I feel more confident that my algorithms behave correctly. More *in vivo* testing will follow.

nextgens:
First the context is: https://emu.freenetproject.org/pipermail/devl/2015-October/038254.html
Some unfinished work was done at https://github.com/nextgens/fred/tree/optimize-peermanager-closerpeer

I love it; I've got a few comments though... will review in-line

src/freenet/node/PeerLocation.java
```diff
@@ -56,43 +60,43 @@ public synchronized int getDegree() {
...
-		for(double currentLoc : newLocs) {
-			if(!Location.isValid(currentLoc)) {
-				Logger.error(this, "Invalid location update for " + this + " ("+currentLoc+')', new Exception("error"));
+		final double[] newPeersLocation = new double[newLocs.length];
+		for (int i = 0; i < newLocs.length; i++) {
+			final double loc = newLocs[i];
+			if (!Location.isValid(loc)) {
+				Logger.error(this, "Invalid location update for " + this + " (" + loc + ")", new Exception("error"));
 				// Ignore it
 				return false;
```
nextgens: Is it really what we should be doing here? Ignore the whole update rather than just skip what  we're not happy with (continue)?

src/freenet/node/PeerLocation.java
```diff
@@ -1,14 +1,16 @@
...
 import freenet.support.Logger;
 
 public class PeerLocation {
 	
 	/** Current location in the keyspace, or -1 if it is unknown */
 	private double currentLocation;
-	/** Current locations of our peer's peers */
+	/** Current sorted array of locations of our peer's peers. Must not be modified,
+	 may only be replaced entirely. */
 	private double[] currentPeersLocation;
```
nextgens: Would it be better if currentPeersLocation included currentLocation?
Would that allow us to optimize further and walk it only once per peer?

src/freenet/node/PeerManager.java
```diff
@@ -1038,21 +1047,9 @@ public PeerNode closerPeer(PeerNode pn, Set<PeerNode> routedTo, double target, b
...
-						ignoreLoc = true;
-					else {
-						for(PeerNode cmpPN : routedTo)
-							if(Math.abs(l - cmpPN.getLocation()) < Double.MIN_VALUE * 2) {
-								ignoreLoc = true;
-								break;
-							}
-					}
-					if(ignoreLoc) continue;
+			if (p.shallWeRouteAccordingToOurPeersLocation()) {
```
nextgens: shallWeRouteAccordingToOurPeersLocation currently doesn't take the HTL into consideration; should it?

test/freenet/node/PeerLocationTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import junit.framework.TestCase;
+
+public class PeerLocationTest extends TestCase {
+
+    private static final double[][] PEER_LOCATIONS = new double[][] {
```
nextgens: I think that it's important to compare what the old logic was doing to the new one... maybe run it only when extensive tests are turned on... 

Have a look to smallestDistanceLegacy in
https://github.com/nextgens/fred/commit/f3f77b4e804581d6fe0204bc5c63283b740c2a28

test/freenet/node/PeerLocationTest.java
```diff
@@ -0,0 +1,110 @@
...
+        // Must be sorted!
+        { 0.1 },
+        { 0.0 },
+        { 0.9 },
+        { 0.1, 0.3 },
+        { 0.1, 0.3, 0.4 },
+        { 0.0, 0.1, 0.2, 0.3, 0.3, 0.35, 0.5, 0.5, 0.99 },
+        { 0.1, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21, 0.22, 0.23, 0.24 }
+    };
```
nextgens: I'd add a specific test to try out the behaviour around 0 and 1; In my branch that's what I could never get right.

src/freenet/node/PeerLocation.java
nextgens: Just to clarify, what I'm suggesting is to factor https://github.com/bertm/fred-staging/blob/ca8ff8a171d187b0e4678b15950fae8f9482d44c/src/freenet/node/PeerManager.java#L1047 inside the same for loop

nextgens:
Do you think I should make room for it in 1475 or is it a 1476 thing?

src/freenet/node/PeerManager.java
nextgens: Specifically, what happens if we're at HTL=1; do we want to preemptively disable FOAF in order to avoid landing the request to a node that's far from the location but has a closer peer?

nextgens:
@toad @Thynix your input would be welcome on the above

src/freenet/node/PeerLocation.java
ArneBab: this is not a change in behaviour, so I think it should not be touched in the complexity improvement. changing `return false` here would entangle changes which should purely improve performance with changes which cause behaviour change — which are two changes which I think should not be released together.

src/freenet/node/PeerLocation.java
bertm: It would be an optimization indeed, moving another single check into the logarithmic complexity.

There is code (PeerManager.java#L1057) keeping track of whether our peer's location or a FOAF location was chosen. Including `currentLocation` here would break that.

src/freenet/node/PeerManager.java
bertm: That's an interesting thought. I think you just identified a bug, the insert chain would end up at a slightly wrong location (or with low probability, a real bad location).

test/freenet/node/PeerLocationTest.java
bertm: Is there anything specific you'd like to see tested?
From my point of view, the choice of peer locations and target locations should already capture all cases.

src/freenet/node/PeerLocation.java
bertm: I must note that there is a slight change in behaviour: we may not always return the exact same locations as the previous code did, as we traverse them in different order. The tests however ensure that we always return locations that have the exact same distance as with the previous code (namely the minimum distance).

bertm:
I think this should be a 1476 thing, unless we can agree that the HTL1 FOAF fix is worth having in 1475.

src/freenet/node/PeerLocation.java
```diff
@@ -107,4 +111,101 @@ synchronized double setLocation(double newLoc) {
...
+			} else {
+				final double loc = locs[right];
+				if (!exclude.contains(loc)) {
+					return loc;
+				}
+				right = (right == locs.length - 1) ? 0 : (right + 1);
+				rightDist = Location.distance(l, locs[right]);
+			}
+		}
+		final double loc = locs[left];
```
ArneBab: this is worst case performance, so it might be useful to add some logging here.

The average runtime of this algorithm should not be *O(log (n) + m)* but rather something slightly above *O(log(n) + m/(n-m))*, since the exclude set is likely to be distributed randomly. It would be O(log(n) + m/(n-m)) for a uniform distribution of exclude in the location list.

nextgens: I'd suggest another change/optimization: persist exclude in a sorted Set across all peers (after all, if we've already routed/tried those locations we don't want to go there because of it again, even if we see it from another peer).

src/freenet/node/PeerManager.java
nextgens: I'll prepare another PR with that behaviour change; I think it should be part of 1475

test/freenet/node/PeerLocationTest.java
nextgens: 0.9,0.99,1,0,0.1,0.11
What comes out with both implementations? :)

ArneBab: there might be some worst case performance in that case. Proposal for a test:

```
{0.00001, 0.00003, 0.0001, 0.0002, 0.4, 0.5, 0.999, 0.9999, 0.99995}
```

test/freenet/node/PeerLocationTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private static final double[] TARGET_LOCATIONS = new double[] {
+        0.0, 0.01, 0.05, 0.09, 0.1, 0.11, 0.29, 0.3, 0.31,
+        0.35, 0.4, 0.45, 0.5, 0.51, 0.9, 0.91, 0.9999
```
ArneBab: maybe add a location very close to 1 here, i.e. 0.999999999.

ArneBab:
This looks good.

test/freenet/node/PeerLocationTest.java
bertm: Performance should remain logarithmic regardless of choice of values. `{ 0.9,0.99,1,0,0.1,0.11 }` is not sorted and thus invalid input. I'll add both of your suggestions nonetheless (after sorting).

src/freenet/node/PeerLocation.java
ArneBab: (I add a note at the relevant line)

src/freenet/node/PeerLocation.java
```diff
@@ -107,4 +111,101 @@ synchronized double setLocation(double newLoc) {
...
+			// All locations are greater, or all locations are smaller.
+			// Closest location must be either smallest or greatest location.
+			left = locs.length - 1;
+			right = 0;
+		} else {
+			left = firstGreater - 1;
+			right = firstGreater;
+		}
+		if (Location.distance(l, locs[left]) <= Location.distance(l, locs[right])) {
+			return left;
```
ArneBab: could this lead to inhomogenous load at the edges (close to 0.0 or 1.0), because we’re always approaching from the middle? This could be avoided by choosing low or high at random if the distance is equal, but taking a random choice is quite expensive.

test/freenet/node/PeerLocationTest.java
bertm: Addressed in 25926d7.

test/freenet/node/PeerLocationTest.java
bertm: Addressed in 25926d7.

src/freenet/node/PeerLocation.java
bertm: > It should be noted, that the complexity estimate uses the information that exclude is likely to be uniformly distributed.

Not really (except that I ignored duplicate values in my analysis). `findFirstGreater` is plain old binary search, so O(log n). `findClosestLocation` invokes it once, plus O(1) additional work, so O(log n). `getClosestPeerLocation` invokes the previous method once, and then iterates over at most *m* unique locations (with O(1) work per location, assuming HashSet) before it finds one that is not excluded. That makes O(log(n) + m).

We cannot make assumptions on the distribution of values in `exclude`.

src/freenet/node/PeerLocation.java
bertm: We're always approaching it from the *left* in the circular key space: we always take the solution that is, say, counterclockwise from the target (or is it actually clockwise? doesn't really matter). Whether or not we are wrapping around from 1 to 0 does not matter here. I named the variables `left` and `right` instead of `low` and `high` to avoid this exact confusion ☺

As such, this only matters when we route to a location exactly inbetween two FOAFs of a single peer, which happens practically never. The previous approach to resolving this choice was "whichever FOAF appeared earlier", the new approach is "whichever FOAF is counterclockwise to the target".

src/freenet/node/PeerLocation.java
bertm: With regards to the logging: it's worst case, but not actually bad. Consider a short-link peer with only a few FOAFs: then it is quite likely that we hit this case.

src/freenet/node/PeerLocation.java
bertm: In cases where the target is between two sequential elements of the FOAF list, yes. When the target is between the last and the first (wrapping around), the resulting location may be different (but the distance equal). My comment to your other node explains this more in-depth.

src/freenet/node/PeerLocation.java
bertm: Actually the old code could lead to inhomogenous load at the edges (in the rare event of equal distance peers), and this code fixes it as a side effect ☺

src/freenet/node/PeerLocation.java
ArneBab: My text was actually wrong ☹ — this should not note “uniformly distributed” but “neighboring”. That’s a leftover from rewriting the text a few times…

src/freenet/node/PeerLocation.java
ArneBab: I tried to estimate the probability of a request being exactly in between two FOAFs, but I did not reach the point where I can give a conclusive probability. My gut-feeling is that there is a small but significant probability to get that, since each request travelling through the network is checked against a large number of nodes.

I stopped the calculation, because your code should not be affected by that anyway.

nextgens:
This has been merged in eaf376509e1667cf654fb8ed563ec20782588149

PR #566, opened 2016-06-22
Don't use FOAF if HTL<2

nextgens:
This is less invasive than #564 and probably a good idea.

We don't want to land a request far from the target just because it has a peer that's closer to our current location!

Please ACK it; I'm not sure whether the callers setting HTL at silly values should be changed.

### DIFF
--- a/src/freenet/node/BaseSender.java
+++ b/src/freenet/node/BaseSender.java
@@ -1,7 +1,5 @@
 package freenet.node;
 
-import static java.util.concurrent.TimeUnit.MINUTES;
-
 import java.util.HashMap;
 import java.util.HashSet;
 
@@ -21,6 +19,8 @@ import freenet.support.LogThresholdCallback;
 import freenet.support.Logger;
 import freenet.support.Logger.LogLevel;
 import freenet.support.TimeUtil;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
 
 /** Base class for request and insert senders.
  * Mostly concerned with what happens *before and up to* we get the Accepted.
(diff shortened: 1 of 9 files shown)


src/freenet/node/BaseSender.java
```diff
@@ -169,7 +169,7 @@ protected void innerRouteRequestsOld(PeerNode next, UIDTag origTag) {
 			 * Don't use sendAsync().
 			 */
         	next.sendSync(req, this, realTimeFlag);
-        	next.reportRoutedTo(key.toNormalizedDouble(), source == null, realTimeFlag, source, nodesRoutedTo);
+                next.reportRoutedTo(key.toNormalizedDouble(), source == null, realTimeFlag, source, nodesRoutedTo, htl);
```
bertm: Whitespace

src/freenet/node/BaseSender.java
```diff
@@ -502,7 +502,7 @@ protected void innerRouteRequestsNew(PeerNode next, UIDTag origTag) {
     			 * Don't use sendAsync().
     			 */
     			if(logMINOR) Logger.minor(this, "Sending "+req+" to "+next);
-    			next.reportRoutedTo(key.toNormalizedDouble(), source == null, realTimeFlag, source, nodesRoutedTo);
+                        next.reportRoutedTo(key.toNormalizedDouble(), source == null, realTimeFlag, source, nodesRoutedTo, htl);
```
bertm: Whitespace

bertm:
This breaks some leading whitespace, but apart from that it looks good to me.

PR #568, opened 2016-06-26
Gradle script

nextgens:
This works for me locally... but I can't get it to work on the CI environment; any idea?


build.gradle
```diff
@@ -0,0 +1,179 @@
...
+                       "Specification-Vendor": "freenetproject.org",
+                       "Implementation-Title": "Freenet",
+                       "Implementation-Version": "0.7.5 ${gitrev}",
+                       "Implementation-Vendor": "freenetproject.org",
+                   ], "common")
+    }
+}
+jar.dependsOn buildInfo
+
+task copyResourcesToClasses2 {
```
Bombe: We could get rid of all this manual copying if we just used the correct paths…

nextgens: yeah, but that means changing the ant scripts... which I am not keen on doing atm. I'd like us to have both build systems for a few releases

Bombe: I hope you do plan to abolish the ant shit (and all its horrendous consequences) at some point in the future…

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -3,28 +3,24 @@
  * http://www.gnu.org/ for further details of the GPL. */
 package freenet.client.filter;
 
+import junit.framework.TestCase;
```
Bombe: Why are these files in this pull request? Looks like they don’t belong here.

nextgens: The tests don't pass without those changes... I've removed the dependency on l10n from the test

nextgens:
@bertm do you see a smarter solution than not running PooledFileRandomAccessBufferTest here?

Is there anything else that the gradle script should do before it's merged?

PR #569, opened 2016-06-29
SessionManager: remove dead sanity check

bertm:
`previousTime` was never written, so this check was useless and would never fail.

These kind of checks belong in a unit test, not in production code.


nextgens:
That SessionManager is a monstruosity inherited from WoT :<

PR #571, opened 2016-06-29
Reduce custom code

bertm:
This reduces the amount (or usage) of custom code included in `freenet.support` for no obvious reason.

This includes the following changes:
* Rewrite of BitArray to use BitSet internally; it provides the same basic functionality.
* Replace some occurrences of our own DoublyLinkedList where LinkedList could have been used. I'd love to get rid of it some day, but that would require a rewrite of LRUMap and LRUQueue.
* Deprecate SortedIntSet. It is still in use by the Library and Spider plugin, so we can't remove it yet. I've rewritten it to use ArrayList internally, so the code is now more or less readable at least.

And the following removals:
* Remove unused and slightly obscure LimitedRangeIntByteArrayMap and its element class; I cannot imagine any plugin using this.
* Remove the Timer class that is not used in fred, nor in a wide collection of plugins.
* Remove LimitedEnumeration: it only has a single user (MultiValueTable), move it in there (and adapt it to its sole use).


src/freenet/support/MultiValueTable.java
```diff
@@ -73,7 +75,7 @@ public boolean containsElement(K key, V value) {
 			v = table.get(key);
 		}
         return (v == null ?
-                new LimitedEnumeration<V>(null) :
+                new EmptyEnumeration<V>() :
```
Bombe: I would have loved to see `Enumeration` leave our code altogether…

src/freenet/support/SortedIntSet.java
```diff
@@ -1,64 +1,70 @@
...
 /**
  * Sorted array of int's.
+ *
+ * This implementation is synchronized on all operations.
+ * This class does not properly respect the Set interface, and the value -1 is treated as a special
+ * number for some operations.
+ *
+ * @deprecated Use ArrayList<Integer>, TreeSet<Integer> or int[] depending on performance needs
  */
+@Deprecated
```
Bombe: \o/

src/freenet/support/SortedIntSet.java
```diff
@@ -1,64 +1,70 @@
...
 	/**
 	 * Get the smallest item on this set
 	 * 
-	 * @return the smallest item
+	 * @return the smallest item, or -1 if the set is empty
 	 */
 	public synchronized int getFirst() {
-		if(length == 0) return -1;
-		return data[0];
+		if (data.isEmpty()) {
```
Bombe: `return data.isEmpty() ? -1 : data.get(0);`?

src/freenet/support/SortedIntSet.java
```diff
@@ -1,64 +1,70 @@
...
 	/**
 	 * Get the largest item on this set
 	 * 
-	 * @return the largest item
+	 * @return the largest item, or -1 if the set is empty
 	 */
 	public synchronized int getLast() {
-		if(length == 0) return -1;
-		return data[length-1];
+		if (data.isEmpty()) {
```
Bombe: Here I would also prefer a one-liner with a ternary operator.

src/freenet/support/SortedIntSet.java
```diff
@@ -93,44 +99,22 @@ public synchronized boolean contains(int num) {
...
 	 */
 	public synchronized boolean remove(int item) {
-		boolean ret = false;
 		int x = binarySearch(item);
 		if(x >= 0) {
-			if(x < length-1)
-				System.arraycopy(data, x+1, data, x, length-x-1);
-			data[--length] = Integer.MAX_VALUE;
-			ret = true;
+			return data.remove(x) != null;
```
Bombe: How did you arrive at that `!= null` expression? I can’t really spot anything similar in the original code…

src/freenet/support/SortedIntSet.java
```diff
@@ -93,44 +99,22 @@ public synchronized boolean contains(int num) {
...
-	 */
-	private synchronized boolean verify() { // TODO: Move to a unit test.
-		int lastItem = -1;
-		for(int i=0;i<length;i++) {
-			int item = data[i];
-			if(i>0) {
-				if(item <= lastItem)
-					throw new IllegalStateException("Verify failed!");
+	private void assertSorted(int[] input) {
+		if (input.length < 2) {
```
Bombe: `<= 1` seems to be clearer than `< 2`…

src/freenet/support/SortedIntSet.java
```diff
@@ -221,12 +191,11 @@ public synchronized void clear() {
 	 * @return sorted array of all items
 	 */
 	public synchronized int[] toArrayRaw() {
```
Bombe: What purpose does this method now have?

src/freenet/support/SortedIntSet.java
bertm: `ArrayList.remove(int index)` returns the value that was removed, or `null` if nothing was removed (note that the returned value is an `Integer` object or `null`). The original method is the equivalent of `ArrayList.remove(Object value)`, which returns whether the object was removed, as a boolean value.

That all said, considering removal of an element from the `ArrayList` is of linear complexity already… we might just use the latter method directly (with explicit cast to `Integer`).

src/freenet/support/SortedIntSet.java
bertm: It is used in the Spider and Library plugin. If I were to remove it, I could as well have removed the entire class immediately ;)

src/freenet/support/SortedIntSet.java
Bombe: Okay, yes, now I see. The amount of `if`s in the original code was slightly confusing. :)

src/freenet/support/SortedIntSet.java
Bombe: Well, those plugins are under our control, I guess… use the other call in those plugins, deprecate it here, remove it once the new plugin versions are released. :)

src/freenet/support/MultiValueTable.java
bertm: I considered doing that, but it would sure cause scope creep for this pull request.

I am planning on rewriting `MultiValueTable` to use `HashMap` and `ArrayList` (or plain arrays) instead of `Hashtable` and `Vector` — this class is a great example of how not to do synchronization after all. That'll be a separate pull request.

src/freenet/support/SortedIntSet.java
bertm: Note that this entire class is deprecated as of this pull request, no need to deprecate every single method as well ☺

src/freenet/support/MultiValueTable.java
Bombe: When we get to updating dependencies I would recommend simply using one of Guava’s excellent (and well-tested) `Multimap` implementations.

src/freenet/support/SortedIntSet.java
Bombe: Point taken. :)

nextgens:
lgtm; do we wait for the ternary operator changes or do I just merge it?

bertm:
I'll address those tonight.
Op 2 jul. 2016 10:28 schreef "Florent Daigniere" <notifications@github.com>:

bertm:
Added a unit test just to be sure I did not break `SortedIntSet`. Should be ready for merge now.

PR #572, opened 2016-06-29
Fix compiler warnings

bertm:
This brings down the number of compilation warnings to a manageable number of 3 for the settings in our ant build on my machine, making a full recompile (and finding that one real error) less of a pain in the ass.

### DIFF
--- a/src/freenet/io/xfer/BulkTransmitter.java
+++ b/src/freenet/io/xfer/BulkTransmitter.java
@@ -247,7 +247,7 @@ public class BulkTransmitter {
 		long lastSentPacket = System.currentTimeMillis();
 outer:	while(true) {
 			int max = Math.min(Integer.MAX_VALUE, prb.blocks);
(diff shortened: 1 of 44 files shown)


src/freenet/io/xfer/BulkTransmitter.java
```diff
@@ -247,7 +247,7 @@ public boolean send() throws DisconnectedException {
 		long lastSentPacket = System.currentTimeMillis();
 outer:	while(true) {
 			int max = Math.min(Integer.MAX_VALUE, prb.blocks);
-			max = Math.min(max, (int)Math.min(Integer.MAX_VALUE, peer.getThrottleWindowSize()));
+			max = Math.min(max, Math.min(Integer.MAX_VALUE, peer.getThrottleWindowSize()));
```
Bombe: If `peer.getThrottleWindowSize()` returns an `int`, the whole `Math.min` expression can go.

bertm:
Don't merge this yet. It breaks stuff, despite not triggering unit test failures.

PR #573, opened 2016-07-04
Remove writeFilter

bertm:
`writeFilter` is never used and most implementations are obviously broken.


Bombe:
Looks good! Off with all the code!

PR #574, opened 2016-07-04
Fix GIF filter

bertm:
This makes the GIF filter actually filter something.

As usual, really invalid files are rejected, but if we can make anything valid out of it (i.e. at least one render block ("frame") succeeds), we will output all that we could validate as a valid (but not necessarily meaningful) GIF file. Unknown extensions and extensions that we cannot validate (such as those that allow free-form text or other arbitrary data) are stripped. The extensions required for looping GIFs are supported.


src/freenet/client/filter/GIFFilter.java
```diff
@@ -30,22 +35,452 @@
...
+		protected static final int EXTENSION_INTRODUCER = 0x21;
+
+		protected GIFValidator(InputStream input, OutputStream output) {
+			this.input = input;
+			this.output = output;
+		}
+
+		/** Checks whether the parsed screen descriptor is valid. */
+		protected boolean validateScreenDescriptor() {
+			if (screenBackgroundColor >= screenColors) {
```
Bombe: Please simplify this.

src/freenet/client/filter/GIFFilter.java
```diff
@@ -30,22 +35,452 @@
...
+		}
+
+		/** Filters the next extension blocks, and skips it when it is unsupported or invalid. */
+		protected void filterExtensionBlock() throws IOException {
+			skipExtensionBlock();
+		}
+
+		/** Signals that image data is found. If the image data is valid, it will be written
+		  * immediately after this method returns. */
+		protected void foundImageData(boolean valid) throws IOException {
```
Bombe: Is this method with a `boolean` argument necessary? Would it be clearer to use two different methods?

src/freenet/client/filter/GIFFilter.java
```diff
@@ -30,22 +35,452 @@
...
+			}
+			final int lzwCodeSize = readByte();
+			if (imageLeft + imageWidth > screenWidth || imageTop + imageHeight > screenHeight ||
+					lzwCodeSize < 2 || lzwCodeSize >= 12 || !validateImageFlags(imageFlags)) {
+				foundImageData(false);
+				skipSubBlocks();
+				return false;
+			} else {
+				foundImageData(true);
+				writeByte(0x2c);
```
Bombe: What’s the function of this `0x2c` here? A constant with a descriptive name is preferrable.

src/freenet/client/filter/GIFFilter.java
```diff
@@ -30,22 +35,452 @@
...
+			int val = input.read();
+			if (val == -1) {
+				throw new EOFException();
+			}
+			return val;
+		}
+
+		/** Write an unsigned byte to the output. */
+		protected final void writeByte(int val) throws IOException {
+			assert(val >= 0x00 && val <= 0xFF);
```
Bombe: I’m not a big fan of `assert`s in production code…

src/freenet/client/filter/GIFFilter.java
bertm: Its purpose is primarily to signal that image data (a render block) has been found. Some extension blocks (such as the graphic control) may only appear once before a render block, and only apply to the render block they precede. See `GIF89aFilter.foundImageData` for details: there's stuff it has to do regardless of the image data being valid (i.e. invalidating a cached graphic control extension block, because it should not apply to the *next* render block even if this one is invalid), and stuff that depends on the data being valid.

Given the simplicity of  `GIF89aFilter.foundImageData`, the sole method that overrides this method, I don't think the additional code for having two different methods is worth it, nor do I think that it would substantially improve readability.

Bombe:
I haven’t delved into the specifics of the GIF format but haven’t found any glaring holes either.

src/freenet/client/filter/GIFFilter.java
Bombe: That does sound a bit like two methods would be appropriate here: one that is always called, and one that is only called for valid data… then again, I’m just nitpicking here. :)

bertm: …and then I'd have to add additional documentation that explains in which order the methods are called, and when; and probably add more state tracking in `GIF89aValidator`: I'd rather not go there ;)

If this method and its usage were more complex than it is right now, or if it were exposed to other classes or even be public, I'd be all for it. But as it is now, I'd rather keep it this way.

Bombe: Okay, okay, I’ll back down, single method is probably fine for this code. :)

src/freenet/client/filter/GIFFilter.java
bertm: The assertions here are mostly to make the intention of the code clear: to human readers and to static code analysis tools. If only Java would have unsigned bytes and shorts… neither this or excessive casting (and `& 0xFF` tricks) would be required :)

Would you rather have it throw `IllegalArgumentException` or `AssertionError`, or just assume we are doing the right thing all along?

Bombe: I would simply discard all unwanted bits, i.e. `& 0xff` on `writeByte` and `& 0xffff` on `writeShort`. Well, changing the type of the argument to `byte` or `short` would be cool, too, but that then requires you to cast a lot when calling the function… well, I guess sometimes you just can’t win. :)

Bombe:
Oh, and did I mention that I would love to see a couple of unit tests for this? :)

bertm:
Here you go, added some unit tests for your pleasure. Oh, and I fixed quite an important bug because of them ;)

test/freenet/client/filter/GIFFilterTest.java
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * @throws AssertionError on failure
+     */
+    private static Bucket filterGIF(Bucket input) {
+        ContentDataFilter filter = new GIFFilter();
+        Bucket output = new ArrayBucket();
+
+        InputStream inStream;
+        OutputStream outStream;
+        try {
```
Bombe: We’re on Java 7 now, we can use

    try (InputStream inStream = input.getInputStream();
        OutputStream outStream = output.getOutputStream()) {
        … do stuff …
    }

Also, in tests there is usually no need to catch exceptions, just let them bubble up so that the test method fails.

Bombe:
Are the images that seem to be book covers available for us to use, license-wise?

bertm:
I pulled most of them from the test cases contained in the Firefox 46 source (which I happened to have around). If we consider them *artistic works*, using them for testing would likely qualify as *fair use*. If we consider them *source code* (which we probably shouldn't, as they are not human-readable), they are released under the Mozilla Public License version 2.0 which is compatible with the GPL we release under.

These are the files not covered by the above:

* `11khz-48kbps-cbr-stereo.mp3`, which is my original work (copied from the MP3 filter unit test).
* `basn0g01.png`, which we already shipped as PNG test case
* `*.filtered.gif` and `*.truncated.gif`, which are hand-crafted derivatives of their originals
* `share-the-safety-like.gif`, which is pulled from [this freesite][share-safety] (after being reported on FMS as a file that did not get filtered well) and must be considered *fair use*

To address your concerns over the book covers: those are mere thumbnails of book covers; using them for unit testing should be covered by *fair use* (and if Mozilla thinks it is okay to use them, it should probably be okay for us as well).

[share-safety]: http://127.0.0.1:8888/SSK@7wRUwQiJ7ZctR6pBPhSt4DffpYQ~JICymgzz48tbc0k,RcE2YadyJ3JkNmGHt-t2TeWQlKLaXTdxttDRdpsTWtM,AQACAAE/NRA%5F2016%5Fcampaign.%5FShare%5Fthe%5Fsafety.-11/common_files/explore-fb.gif

Bombe:
Okay, I just wanted to make sure that this has been given consideration which it obviously has been. :)

test/freenet/client/filter/GIFFilterTest.java
nextgens: I agree, it should significantly reduce the boilerplate... get that done and I'll merge it :)

nextgens: done in 41b38dd

bertm:
Thanks!
Op 9 jul. 2016 23:21 schreef "Florent Daigniere" <notifications@github.com>:

PR #575, opened 2016-07-05
Use JNA to reduce our dependency on freenet-ext

nextgens:
This works for me. I wonder if we should just drop support of preallocation for non-posix systems


src/freenet/support/io/PooledFileRandomAccessBuffer.java
```diff
@@ -99,19 +99,25 @@ public PooledFileRandomAccessBuffer(File file, boolean readOnly, long forceLengt
...
-                if(seedRandom != null)
-                    mt = new MersenneTwister(seedRandom.nextLong());
-                byte[] buf = new byte[4096];
-                for(long l = 0; l < forceLength; l+=4096) {
-                    if(mt != null)
-                        mt.nextBytes(buf);
-                    int maxWrite = (int)Math.min(4096, forceLength - l);
-                    raf.write(buf, 0, maxWrite);
+                if (Fallocate.isSupported()) {
+                    Fallocate.forDescriptor(raf.getFD(), forceLength).execute();
```
bertm: This is not equivalent: the original code write random data, not zeroes.

bertm: Never mind, it probably filled with random data to prevent compression, not for "security".

src/freenet/support/io/PooledFileRandomAccessBuffer.java
```diff
@@ -99,19 +99,25 @@ public PooledFileRandomAccessBuffer(File file, boolean readOnly, long forceLengt
...
+                } else {
+                    Logger.normal(this, "fallocate() not supported; using legacy method");
+                    raf.seek(0);
+                    MersenneTwister mt = null;
+                    if(seedRandom != null)
+                        mt = new MersenneTwister(seedRandom.nextLong());
+                    byte[] buf = new byte[4096];
+                    for(long l = 0; l < forceLength; l+=4096) {
```
bertm: Can we move the legacy fallback method into `Fallocate` please?

nextgens: Done in 3f5291a

src/freenet/support/io/Fallocate.java
```diff
@@ -111,4 +118,28 @@ private static int getDescriptor(FileDescriptor descriptor) {
...
+      if (offset % (1024 * 1024 * 1024L) == 0) {
+        mt = new MersenneTwister();
+        if (starting) {
+          WrapperManager.signalStarting( (int) MINUTES.toMillis(5));
+          if (x++ % 32 == 0)
+            System.err.println(
+                "Preallocating space : "
+                + offset
+                + "/"
+                + newLength);
```
bertm: Can we reduce the spamminess of this method, or should I file a separate pull request once this has landed?

nextgens: toad wanted it to be visible in wrapper.log since that's the only thing users can be expected to see; I can reduce the periodicity if you want... what would be reasonable?

src/freenet/support/io/Fallocate.java
```diff
@@ -111,4 +118,28 @@ private static int getDescriptor(FileDescriptor descriptor) {
...
+    ByteBuffer bb = ByteBuffer.wrap(b);
+    int x = 0;
+    while (offset < newLength) {
+      bb.rewind();
+      mt.nextBytes(b);
+      offset += fc.write(bb, offset);
+      if (offset % (1024 * 1024 * 1024L) == 0) {
+        mt = new MersenneTwister();
+        if (starting) {
+          WrapperManager.signalStarting( (int) MINUTES.toMillis(5));
```
nextgens: I don't remember tbh; I'll need to check

src/freenet/support/io/Fallocate.java
bertm: Alright, that sort of makes sense. Once per 32 GiB as it is now is not too
bad, changing that to % 32 = 1 or similar would help to avoid displaying it
for small resizes.
Op 8 jul. 2016 00:58 schreef "Florent Daigniere" <notifications@github.com>:

In src/freenet/support/io/Fallocate.java
<https://github.com/freenet/fred/pull/575#discussion_r70002452>:

> +    ByteBuffer bb = ByteBuffer.wrap(b);
> +    int x = 0;
> +    while (offset < newLength) {
> +      bb.rewind();
> +      mt.nextBytes(b);
> +      offset += fc.write(bb, offset);
> +      if (offset % (1024 * 1024 * 1024L) == 0) {
> +        mt = new MersenneTwister();
> +        if (starting) {
> +          WrapperManager.signalStarting( (int) MINUTES.toMillis(5));
> +          if (x++ % 32 == 0)
> +            System.err.println(
> +                "Preallocating space : "
> +                + offset
> +                + "/"
> +                + newLength);

toad wanted it to be visible in wrapper.log since that's the only thing
users can be expected to see; I can reduce the periodicity if you want...
what would be reasonable?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/freenet/fred/pull/575/files/548f081d4ed3f62f2d1b83eaebcc9dffc1d79f23..3f5291a7bd2ab96d77417dd12fbd08f50ea5529c#r70002452>,
or mute the thread
<https://github.com/notifications/unsubscribe/AATu91YLWJ_5SGOfriajGXzn6F9aj-E3ks5qTYSqgaJpZM4JFRFm>
.

nextgens: Done in 9225580

src/freenet/support/io/Fallocate.java
nextgens: We can't; NodeStarter.start() needs to return (WrapperListener.start()) ... otherwise the wrapper kicks us... so setting starting to true (and pinging WrapperManager.signalStarting periodically) is mandatory until that's happened.

PR #576, opened 2016-07-15
Faster simulations (summer project)

toad:
This will likely need significant refactoring. The main goals are to enable fast unit tests involving multiple actual nodes and to demo load management simulations; to this end I implemented various bypass layers etc, see the PDF, especially the intro. I tried to break stuff out into other branches, pull requests for which have been filed.

Documentation:
http://www.toselandcs.co.uk/flogmirror/mjt92-diss-final.pdf
http://www.toselandcs.co.uk/flogmirror/mjt92-part2-source.tar.gz

I will not be able to engage in detailed conversation on this in the near future because of my employer's intellectual property policy. So feel free to summarily close this, unless somebody wants to take ownership of it. I will document it on my blog/flog too.


xor-freenet:
It's totally awesome you did so much scientific work on this and even ship it with a 60-page paper!
It'd be a shame if this bitrotted to death; so now that you said on IRC you're not blocked by your work contract anymore let's try to figure out how to avoid that :)

You say: 

> This will likely need significant refactoring.

I applaud your carefulness, but unfortunately very vague, possibly overly strict self-criticism on very complex PRs can very likely be a reason for them to bitrot - there's a wall of code for reviewers to tackle, but no idea where to start with the required fixes.
So can you clarify upon your statement please?
What I'd be interested in:
- Is there any very serious breakage as in "has bugs which will affect the users", or did this work properly in your tests? If yes what?
- Is "need significant refactoring" to be read as "this MUST be refactored before it is merged" or rather "it'd be nice to refactor it before, but it may work without that as well".
- Now that your contract isn't blocking you anymore, could you maybe even do the fixes yourself and ship this in a "I'd merge that as is" state? Just asking, don't feel pushed :)

I hope we can find a way to resolve this - you're our most renowned contributor so it should in theory be very easy to merge such large contributions from you because you're trustworthy enough so we don't have to review it with big scrutiny. It should mostly be a matter of you clarifying what you think is critical to be fixed IMHO :) So I'd be very happy if you could do that!

toad:
Hmmm, I'll need to think about that. It was some time back and I don't recall exactly what its state is.

The main problem is it touches various core code though.

It shouldn't break anything really IIRC.

ArneBab:
Hi @toad, do you have resources for fixing the conflicts? 

Also could you list the pull-requests in an order you’d suggest for reading them? I’d like to have a stab at reviewing your work on this the next few months.

PR #577, opened 2016-07-16
Fix a potential problem with fallocate()

nextgens:
It turns out that some old libc will rewrite calls from
 fallocate() to posix_fallocate() ... and this might take
 a while depending on the FS.

We need to ensure that regardless of the situation,
 the wrapper doesn't kill the node.

### DIFF
--- a/src/freenet/support/WrapperKeepalive.java
+++ b/src/freenet/support/WrapperKeepalive.java
@@ -1,0 +1,27 @@
+package freenet.support;
+
+import org.tanukisoftware.wrapper.WrapperManager;
+
+import java.io.IOException;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+public class WrapperKeepalive extends Thread implements AutoCloseable {
+  private volatile boolean shutdown = false;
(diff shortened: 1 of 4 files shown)


src/freenet/support/WrapperKeepalive.java
```diff
@@ -14,7 +15,7 @@
   public void run() {
     while (!shutdown) {
       try {
-        WrapperManager.signalStarting(INTERVAL);
+        WrapperManager.signalStarting(INTERVAL + (int)SECONDS.toMillis(5));
```
Thynix: 5 seconds is to account for delay in the thread wakeup? What's the reasoning for that value? Is there a more reliable way to schedule this to be regularly called that would avoid the need for a fudge factor like this?

nextgens: The magic value comes out of thin air... and sounded reasonable at the time; arguably it could be smaller.

I don't know of any better way to schedule it regularly... we could pass a reference to the node's Ticker... but it's pretty deep in the stack... and I've figured that simplicity was better here (after all the thread dies when we leave the try{} block!)

PR #578, opened 2016-07-22
Add JNAGMP magic to speedup DSA

nextgens:
I am not convinced that their binaries are more trustworthy than ours... but they're more recent and support constant-time operations...

Alternatively we could go back to using the NativeBigInteger we already ship (still with JNA)


nextgens:
Oh, and OFC they only ship:
 linux-x86-64/libgmp.so
 sunos-x86-64/libgmp.so
 darwin/libgmp.dylib

so it might be completely unsuitable.

PR #579, opened 2016-07-23
Improve upon a few low-hanging fruits

nextgens:
The profiler has made these obvious...

### DIFF
--- a/src/freenet/node/NewPacketFormat.java
+++ b/src/freenet/node/NewPacketFormat.java
@@ -10,6 +10,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 import freenet.crypt.BlockCipher;
 import freenet.crypt.HMAC;
@@ -60,7 +61,7 @@ public class NewPacketFormat implements PacketFormat {
 
 	/** The actual buffer of outgoing messages that have not yet been acked.
 	 * LOCKING: Protected by sendBufferLock. */
-	private final ArrayList<HashMap<Integer, MessageWrapper>> startedByPrio;
+	private final ArrayList<ConcurrentSkipListMap<Integer, MessageWrapper>> startedByPrio;
 	/** The next message ID for outgoing messages.
 	 * LOCKING: Protected by (this). */
 	private int nextMessageID;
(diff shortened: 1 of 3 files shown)


src/freenet/node/NewPacketFormat.java
```diff
@@ -10,6 +10,7 @@
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.ConcurrentSkipListMap;
```
Bombe: Is there a reason you chose `ConcurrentSkipListMap` here? From what I see, we’re mostly iterating over it so this doesn’t really gain us anything…

nextgens: It's improving upon the HashMap... but yes, we could use a Deque or something else. Should we?

IMO the whole class should be re-factored... it doesn't make much sense as is... We could maintain reference to packets sorted by deadline instead.

Bombe: Improving upon what? `ConcurrentSkipListMap` iteration may be faster but at increased memory cost. Maybe a `LinkedHashMap` would suffice?

(Also, I’m a fan of interfaces in type declarations so I’d prefer having `startedByPrio` declared as `List<Map<Integer, MessageWrapper>>`…)

src/freenet/node/NewPacketFormat.java
```diff
@@ -104,9 +104,9 @@ public void shouldUpdate(){
 	public NewPacketFormat(BasePeerNode pn, int ourInitialMsgID, int theirInitialMsgID) {
 		this.pn = pn;
 
-		startedByPrio = new ArrayList<ConcurrentSkipListMap<Integer, MessageWrapper>>(DMT.NUM_PRIORITIES);
+		startedByPrio = new ArrayList(DMT.NUM_PRIORITIES);
```
Bombe: This (and the next one) are now completely missing the type parameters?

src/freenet/node/NewPacketFormat.java
```diff
@@ -1147,8 +1147,8 @@ private double averageRTT() {
 	static class SentPacket {
 		final SessionKey sessionKey;
 		NewPacketFormat npf;
-		List<MessageWrapper> messages = new ArrayList<MessageWrapper>();
-		List<int[]> ranges = new ArrayList<int[]>();
+		List<MessageWrapper> messages = new ArrayList();
```
Bombe: Use `ArrayList<>()` instead of a raw type, please. :)

src/freenet/node/NewPacketFormat.java
```diff
@@ -973,9 +965,9 @@ public void checkForLostPackets() {
 		// LOCKING: No packet may be sent while connected = false.
 		// So we guarantee that no more packets are sent by setting this here.
 		synchronized(sendBufferLock) {
-			for(HashMap<Integer, MessageWrapper> queue : startedByPrio) {
+			for(Map<Integer, MessageWrapper> queue : startedByPrio) {
 				if(items == null)
```
Bombe: Is this `if` seriously only here because someone didn’t want to initialize the `items` where they were declared?

nextgens: Took care of it in 732b42e

PR #580, opened 2016-07-25
Implement the osx/linux glue for ProcessPriority

nextgens:
This works for me on linux but needs wider testing.


nextgens:
Just to be clear: we don't use the logger there because at that stage it's not loaded!

steveatinfincia:
This works fine here, tested on a new install using OS X 10.11.6 and a darknet peer

PR #581, opened 2016-07-29
Update SSL code

bertm:
This updates the SSL code so that it works on Java 8+ (as well as on Java 7), and uses more up-to-date crypto algorithms and protocols. In addition, this includes a few cleanups and error message improvements.

### DIFF
--- a/src/freenet/crypt/SSL.java
+++ b/src/freenet/crypt/SSL.java
@@ -47,6 +47,15 @@ import java.net.ServerSocket;
 
 public class SSL {
 
+	private static final int RSA_STRENGTH = 2048;
+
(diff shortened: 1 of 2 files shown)


bertm:
(not that I really care about this functionality, but is was easier to fix than to remove)

nextgens:
Hmfff. Don't create 2048 bit RSA keys ; Try to be consistent with a security-level:
https://www.keylength.com/en/3/

Make a p256 key if need be... and switch to the ECDSA version of the cipher

bertm:
Thanks for the suggestion, fixed.

nextgens:
I haven't tested it but lgtm

PR #582, opened 2016-08-03
Display the default value as intended

nextgens:


bertm:
Nice catch! Looks good to me.

PR #583, opened 2016-08-06
bug6562: mark temporary files as hidden

nextgens:
This should do the trick; I haven't tested it on windows

### DIFF
--- a/src/freenet/client/async/SplitFileFetcher.java
+++ b/src/freenet/client/async/SplitFileFetcher.java
@@ -5,6 +5,9 @@ import java.io.DataOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.file.FileSystemException;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
 import java.util.List;
 
 import freenet.client.ClientMetadata;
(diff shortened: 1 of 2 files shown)


bertm:
Are you sure changing these filenames does not break existing persistent downloads?

Apart from possible persistence issues, the code looks sane :)

nextgens:
I haven't tested it, but considering that it's calls to File.createTempFile() that I have changed, I think it will be fine.

Thynix:
Are temporary files moved to the destination filename? I haven't tested it
(and won't have time to until next week) but that seems to mean on Windows
that those files will have to be set to not hidden if desired.

nextgens:
I think that we copy from an encrypted bucket to a new file in ClientGetter... but yes, it deserves to be tested :)

bertm:
Will test in a few days.

Op 20 aug. 2016 18:37 schreef "Florent Daigniere" <notifications@github.com

bertm:
Tested working on Linux when switching from `next` to your branch with pending downloads.

ArneBab:
@bertm does that mean you consider this ready to be merged?

ArneBab:
did anyone test this under Windows?

ArneBab:
This makes Freenet fail under Windows 8.1: 

> Under Windows you can set the hidden flag on any file (e.g. text.txt). And if it is set, but the writeable flag is not set then you can edit the file but never save it until the hidden flag is cleared. And the error message is not very clear in what was wrong. 

https://bugs.eclipse.org/bugs/show_bug.cgi?id=194216

I’ll revert this change.

nextgens:
That's fine; I've explicitly said it hadn't been tested... you've merged it without testing it :|

nextgens:
That's fine.

PR #584, opened 2016-08-10
Class IfNull: Deprecate in favor of Java 7 providing the same

xor-freenet:
Fred does not use these, they were added for plugins.
So let's deprecate them for a while before we remove them.


bertm:
Looks good to me

nextgens:
I'd get rid of the whole package

PR #586, opened 2016-08-17
Fix Eclipse classpath to match the Gradle one

xor-freenet:
Notice:
This adds some dependencies even though they're only listed at `dependencyVerification` sections of Gradle, not  `dependencies`. This is because they're dependencies of other dependencies:
- `hamcrest-core-1.3.jar`  is a compile-time dependency of `hamcrest-library-1.3.jar` 
- `objenesis-1.0.jar` is a runtime dependency of `mockito-core-1.9.5.jar`

_Please consider using this PR message as message of the merge commit._


nextgens:
This is deprecated by #590

PR #588, opened 2016-08-19
Persist "Bookmark Updated" notifications across restarts

xor-freenet:
Previously, if you received a notification about an updated USK bookmark, and then restarted the node, the notification would be lost: The node stored the updated edition number to disk, but not the notification. So it wouldn't download the edition again, and thus not create another notification.
This is fixed by this PR.

_Please use this PR message as message of the merge commit._

### DIFF
--- a/src/freenet/clients/http/BookmarkEditorToadlet.java
+++ b/src/freenet/clients/http/BookmarkEditorToadlet.java
@@ -401,7 +401,10 @@ public class BookmarkEditorToadlet extends Toadlet {
 						if (!isValidName(name)) {
               addNameError(pageMaker, content);
 						} else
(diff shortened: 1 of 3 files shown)


nextgens:
I don't think it should be merged until  flush-on-shutdown is implemented. The current behaviour is consistent / reproductible... the behaviour with this in depends on timing and will undoubtedly lead to frustrated users.

xor-freenet:
Thanks for the review!
I've fixed this - adding shutdown code to fred is easier than I had anticipated :)

ArneBab:
@nextgens does the additional commit address your review?

xor-freenet:
@ArneBab 
> @nextgens does the additional commit (7d1f549) address your review?

Yes that was the intention of that commit, see the comment before it (which is why I said that your release notes saying "semi"-persistent weren't correct).

ArneBab:
I need a reply from @nextgens on that.

nextgens:
I haven't tested it but yes; lgtm

PR #589, opened 2016-08-20
Handle EOPNOTSUP

nextgens:
This should deal with https://gist.githubusercontent.com/xor-freenet/18027234f1849531f17e84f62056086c/raw/62a39f65c53555bda0d6f37788d80cde9a8ff447/fred-failure.txt as mentioned in #585

### DIFF
--- a/src/freenet/support/io/Fallocate.java
+++ b/src/freenet/support/io/Fallocate.java
@@ -27,20 +27,21 @@ public final class Fallocate {
   private final int fd;
   private int mode;
   private long offset;
-  private final long length;
+  private final long final_filesize;
   private final FileChannel channel;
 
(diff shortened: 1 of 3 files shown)


src/freenet/support/io/Fallocate.java
```diff
@@ -66,19 +58,24 @@ private void requireLinux(String feature) {
   }
 
   private void throwUnsupported(String feature) {
-    throw new UnsupportedOperationException(feature +
-                                            " is not supported on this operating system");
+    throw new UnsupportedOperationException(feature + " is not supported on this operating system");
```
Thynix: Maybe s/operating/file/ ?

xor-freenet:
This fixes the unit test failures indeed. Downloads and uploads also work.
Thank you!

PR #590, opened 2016-08-20
Remove ant buildfile

nextgens:
This is meant as an alternative to both #587 and #586


xor-freenet:
You say this is an alternative to #586, i.e. to a PR which fixes Eclipse+Ant to work with the Gradle-induced changes.
However the list of files changed by this PR contains no Eclipse related files, which means that Eclipse would be broken by this PR removing Ant: Eclipse currently relies on Ant for building fred.
Did you maybe just forget to commit the Eclipse config files?

(Please don't get all my PRs wrong: I'd be eager to try Gradle, I accept it's better than Ant only from looking at the much nicer readability of the scripts. But I cannot yet use it as long as it doesn't support what I was doing with Ant; so I'm trying to keep Ant working meanwhile.)

ArneBab:
@nextgens you are breaking any kind of sane review process by self-merging disputed changes without ack by any other developer.

PR #596, opened 2016-09-18
Ant builder: Force user to notice deprecation in favor of Gradle

xor-freenet:
Intended as a replacement for https://github.com/freenet/fred/pull/590
Based on https://github.com/freenet/fred/pull/587 to avoid merge conflicts.

The Ant builder supports various things which the Gradle one cannot do yet, for example:
- Building the JavaDoc
- Computing test coverage (https://github.com/freenet/fred/pull/553)
- Static code analysis using PMD (https://pmd.github.io/)
- Static code analysis using findbugs (http://findbugs.sourceforge.net/)
- Finding duplicate code with CPD (http://pmd.sourceforge.net/pmd-4.3.0/cpd.html)
- The Eclipse IDE (used by toad_ and xor). Notice that albeit #590 claims that it replaces #586 (= fix Eclipse to work) it doesn't actually seem to contain any Eclipse-related code so Eclipse would still be broken with Gradle.

I would thus really like to keep the Ant builder for some time.
To justify that, this PR will make it fail with a deprecation warning until the user manually disables the warning by a command line parameter / changing the override file.

As a compensation for keeping it in the codebase I will continue maintaining it.


nextgens:
This is deprecated by #590

PR #597, opened 2016-09-18
Ant builder: Fix JavaDoc API links to use correct version and SSL

xor-freenet:
Based on https://github.com/freenet/fred/pull/587 so it actually works.


nextgens:
This is deprecated by #590

PR #598, opened 2016-09-20
Make feed detection more similar to Firefox-2

bertm:
Make feed detection more similar to Firefox-2 to avoid some false positives.

Pull request forwarded from FMS post by Eleriseth as reposted by FuckSárközi. Whitespace/indentation is mine, for it appeared stripped in the repost.


Bombe:
Apart from making this a whole lot more unreadable by emulating C, what problem is being solved here?

bertm:
I'm not particularly happy with the code style either.

What this is trying to solve is false positives by piece of code that determines whether Firefox will content-type sniff non-RSS content as RSS content (and show a warning if this is the case). Most often it were PDF files that caused problems; they contain RDF XML tags at times, but usually only after other XML-like (but otherwise invalid) tags. Firefox avoids false positives in these situations by checking for top-level tags, not just any tag.

Bombe:
Okay, can this be closed in favor of #599? Because that looks a lot more polished. :)

PR #599, opened 2016-09-20
Improve RSS/feed sniffing

bertm:
This more resembles the sniffing done by Firefox and should lead to fewer
false positives.

This pull request is intended as a more readable alternative to #598 and includes unit tests.

### DIFF
--- a/src/freenet/clients/http/FProxyToadlet.java
+++ b/src/freenet/clients/http/FProxyToadlet.java
@@ -164,7 +164,7 @@ public final class FProxyToadlet extends Toadlet implements RequestClient {
 			if(mimeType.compareTo("application/xhtml+xml")==0){
 				mimeType="text/html";
 			}
-			if(horribleEvilHack(data) && !(mimeType.startsWith("application/rss+xml"))) {
+			if(isSniffedAsFeed(data) && !(mimeType.startsWith("application/rss+xml"))) {
 				PageNode page = context.getPageMaker().getPageNode(l10n("dangerousRSSTitle"), context);
 				HTMLNode pageNode = page.outer;
 				HTMLNode contentNode = page.content;
@@ -399,7 +399,7 @@ public final class FProxyToadlet extends Toadlet implements RequestClient {
 	 * This is a horrible evil hack; we shouldn't be doing blacklisting, we should be doing whitelisting.
 	 * REDFLAG Expect future security issues!
 	 * @throws IOException */
-	private static boolean horribleEvilHack(Bucket data) throws IOException {
+	private static boolean isSniffedAsFeed(Bucket data) throws IOException {
 		DataInputStream is = null;
 		try {
 			int sz = (int) Math.min(data.size(), 512);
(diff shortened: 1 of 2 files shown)


src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -409,46 +409,88 @@ private static boolean horribleEvilHack(Bucket data) throws IOException {
...
+		}
+		return -1;
+	}
+
+	/**
+	 * Checks whether the given data starts with the characters in the key value, when starting
+	 * at fromIndex in the data array.
+	 */
+	private static boolean startsWithString(byte[] data, String key, int fromIndex) {
+		if (data.length - fromIndex < key.length()) {
```
Bombe: Hmm, this wouldn’t find an “a” in a buffer of “<a” when starting from the a itself. I think you need a `+1` in there somewhere. :)

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -409,46 +409,88 @@ private static boolean horribleEvilHack(Bucket data) throws IOException {
...
+			return false;
+		}
+		for (int i = 0; i < key.length(); i++) {
+			if (data[i + fromIndex] != key.charAt(i)) {
```
Bombe: Two nitpickings:

* I’d rather say `fromIndex + i` because it reads better.
* `key.charAt` returns a `char`. For US-ASCII this doesn’t matter, though…

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -409,46 +409,88 @@ private static boolean horribleEvilHack(Bucket data) throws IOException {
...
-			if(b == find.charAt(bufProgress)) {
-				bufProgress++;
-				if(bufProgress == find.length()) return true;
-				offset++;
-			} else {
-				if(bufProgress == 0)
-					offset++; // Try the next byte.
-				else
-					bufProgress = 0; // Reset to the first char of the keyword.
+	/**
```
Bombe: Oh, and can you move this to its own file? `FproxyToadlet` doesn’t really need to grow… :)

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -409,46 +409,88 @@ private static boolean horribleEvilHack(Bucket data) throws IOException {
...
+	 * Look for any of the following strings as top-level XML tags:
+	 * <rss
```
Bombe: This should be “&amp;lt;rss” too.

test/freenet/clients/http/RssSniffTest.java
```diff
@@ -0,0 +1,39 @@
...
+        "<?xml version=\"1.0\" encoding=\"utf-8\"?<rss xmlns:atom=\"http://www.w3.org/2005/Atom\" version=\"2.0\" xmlns:dc=\"http://purl.org/dc/elements/1.1/\">",
+        "<?xml?><!-- <rss> --><? <rss><bogus><rss>", // <bogus> is top-level
+        "<!DOCTYPE html><bogus><rss", // <bogus> is top-level
+        "<!rdf:RDF", // comment
+        "<bogus><rdf:RDFbogus", // <bogus> is top-level
+        "<!--<rdf:RDF>bogus-->", // comment
+        "<!DOCTYPE html>bogus ... <!-- more bogus --><?xml version><pdf><rss" // <pdf> is top-level
+    };
+
+    public void testRssSniffingPositive() {
```
Bombe: I’d also rather have more test methods here, each with a descriptive name, such es `rdfTagIsRecognized`, `bogusRdfTagIsRecognized`, `rdfTagInBogusTagIsNotRecognized`.

test/freenet/clients/http/RssSniffTest.java
```diff
@@ -0,0 +1,39 @@
+package freenet.clients.http;
+
+import junit.framework.TestCase;
+
+/**
+ * Test cases for the RSS feed sniffing logic in FProxyToadlet.
+ */
+public class RssSniffTest extends TestCase {
```
Bombe: And I think we already have a dependency on JUnit 4, so just drop the `extends TestCase` and annotate the test methods with `@Test`, please? :)

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -399,7 +399,7 @@ public static String l10n(String msg) {
 	 * This is a horrible evil hack; we shouldn't be doing blacklisting, we should be doing whitelisting.
 	 * REDFLAG Expect future security issues!
 	 * @throws IOException */
-	private static boolean horribleEvilHack(Bucket data) throws IOException {
+	private static boolean isSniffedAsFeed(Bucket data) throws IOException {
```
Bombe: Oh, and can you move this to its own file? FproxyToadlet doesn’t really need to grow… :)

ArneBab:
@bertm do you plan to finish addressing the algorithm changes requested by Bombe?

ArneBab:
@Bombe would you be ok with merging as is, or can you simply do the changes?

Bombe:
I can do the changes myself but I’ll need some days. :)

ArneBab:
David Roden <notifications@github.com> writes:

> I can do the changes myself but I’ll need some days. :)

Cool! Thank you!

src/freenet/clients/http/FProxyToadlet.java
Bombe: Turns out this is fine.

Bombe:
Superseded by #659.

PR #600, opened 2016-09-23
Remove unused parameters

Bombe:
This removes a lot of unused method parameters.

Due to plugins being exposed to _everything_, this might theoretically break an unknown number of plugins.


xor-freenet:
- Was this generated automatically, e.g. with an IDE? If you tell us how, then it is easier to validate.

- Have you checked how many merge conflicts this generates with https://github.com/freenet/fred/pull/576 ? Both PRs are very large and thus are very likely to collide. However toad's work is so complex it even ships with a whole scientific paper of 62 pages from our most experienced contributor; while this PR here is a rather simple thing - albeit I do of course recognize it may have been lots of tedious work and it's honorable you did it! :|   Thus should we maybe defer merging this until #576 is merged, to ensure the more expensive #576 doesn't bitrot to death from potential merge conflicts?

PR #601, opened 2016-09-28
fix: Readme described incorrectly how to skip tests

ArneBab:
thanks to muixirt!


nextgens:
This is deprecated by #590

ArneBab:
I did not see any consensus about deleting the ant build scripts.

PR #603, opened 2016-10-23
TimeUtilTest: Test setTimeToZero()

xor-freenet:
Testing this because WoT assumes it to be correct so it can use it as foundation for unit tests of further Date manipulation functions.


nextgens:
It's great that WoT makes assumptions on what the JDK doesn't guarantee...

None of this concerns fred. Either you should upstream your brilliant idea or keep it in the code base that relies on those assumptions.

PR #604, opened 2016-11-15
Fix ContentFilterTest to work with both Gradle and Ant

xor-freenet:
Notice: This is based on https://github.com/freenet/fred/pull/587 to ensure Ant works in the first place.

Commit 813c2c94c95e3cde72c7c2ba5a12bbb3ea2e3e9b fixed ContentFilterTest to succeed with Gradle but broke it for Ant.

This PR reverts that part of 813c2c94c95e3cde72c7c2ba5a12bbb3ea2e3e9b and uses code which the same commit added to other tests to fix ContentFilterTest to work with *both* Gradle and Ant.

_Please consider using the subject and body of this PR as the message of the merge commit._


nextgens:
https://docs.gradle.org/current/userguide/eclipse_plugin.html

send a patch.

PR #605, opened 2016-12-13
n2n: Add remaining letter counter to textfield

ArneBab:
This should resolve the most pressing issue of sending a too large N2N message and losing it. It is not a deep solution but resolves the main issue.

### DIFF
--- a/src/freenet/clients/http/N2NTMToadlet.java
+++ b/src/freenet/clients/http/N2NTMToadlet.java
@@ -76,7 +76,7 @@ public class N2NTMToadlet extends Toadlet {
 			}
 			HashMap<String, String> peers = new HashMap<String, String>();
 			peers.put(input_hashcode_string, peernode_name);
-			createN2NTMSendForm(pageNode, ctx.isAdvancedModeEnabled(), contentNode, ctx, peers);
+			createN2NTMSendForm(pageNode, ctx.isAdvancedModeEnabled(), contentNode, ctx, peers, node.isFProxyJavascriptEnabled());
 			this.writeHTMLReply(ctx, 200, "OK", pageNode.generate());
 			return;
 		}
@@ -261,10 +261,17 @@ public class N2NTMToadlet extends Toadlet {
 		ctx.sendReplyHeaders(302, "Found", headers, null, 0);
 	}
 
-	public void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
-			HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers)
+	public static void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
+										   HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers)
 			throws ToadletContextClosedException, IOException {
-		final boolean fProxyJavascriptEnabled = node.isFProxyJavascriptEnabled();
+		// if we do not know whether javascript is enabled, assume that it is not (provides the old API).
+		createN2NTMSendForm(pageNode, advancedMode, contentNode, ctx, peers, false);
+	}
+	
+	public static void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
+										   HTMLNode contentNode, ToadletContext ctx,
+										   HashMap<String, String> peers, boolean fProxyJavascriptEnabled)
+			throws ToadletContextClosedException, IOException {
 		if (fProxyJavascriptEnabled) {
 			contentNode.addChild("script", new String[] {"type", "src"}, new String[] {"text/javascript",  "/static/js/countletters.js"});
 		}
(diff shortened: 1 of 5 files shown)


src/freenet/clients/http/N2NTMToadlet.java
```diff
@@ -285,6 +289,14 @@ public static void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
 		}
 		messageForm.addChild("textarea", new String[] { "id", "name", "rows",
 				"cols" }, new String[] { "n2ntmtext", "message", "8", "74" });
+		if (fProxyJavascriptEnabled) {
+			HTMLNode letterCountInfo = messageForm.addChild("div", new String[] {"class"}, new String[] {"n2ntmcountouter"});
+			letterCountInfo.addChild("#", NodeL10n.getBase().getString("N2NTMToadlet.remainingLetters"));
+			letterCountInfo.addChild("span", new String[] {"id"}, new String[] {"n2ntmcount"});
+			letterCountInfo.addChild("script", new String[] {"type"}, new String[] {"text/javascript"})
```
Bombe: This should be done in `countletters.js`, on the document’s “onload” event (or jQuery’s “ready” if we have jQuery available).

ArneBab: Do you mean refactoring `countletters()` to accept the id of the text-field we want to enhance with the letter count as well as the L10n string to use, instead of giving it the id of the span to fill with the number?

`countletters(n2ntmtext, n2ntmcount)` -> `countletters(n2ntextfield, n2nremainingtextlabel)`?

ArneBab: I’m calling `countletters` to pass the element IDs to the js code, so we can re-use this in different places or for multiple text-fields on one page.

Bombe: Add classes to some of the elements and bind the listeners to all instances of elements with the right classes. In order to use it multiple times on a single page, add one “top-level” class to some container that contains both the input field and the counter display, then add classes to both the the input field and the counter display itself. Searching for the all elements with the top-level class and then searching for the elements with the classes below that gives you an easy way to restrict your listeners and actions so that you don’t have to mess around with single IDs.

ArneBab: That sounds like a nice separation of concerns. I’ll try to go with it when I get the free creative time to work on this code again. Thank you!

ArneBab:
it’s DONE! ☺

src/freenet/clients/http/N2NTMToadlet.java
ArneBab: I did that, now, using the DOMContentLoaded event to run the code.

nextgens:
So, +80 lines of code for "helping" users who use fproxy with javascript (despite instructions)...

Why don't you fix the root cause of the problem instead? There is no fundamental reason why N2N messages are stripped at that length.

ArneBab:
Because I don’t see the clear path to fix the root cause — and no one who knew the transport layer cared enough during the past several years despite this being one of the few features which actually encourage people to use darknet — so I’m now following the path I see.

And finally plain lines of code are not a useful metric, neither for accomplishment, nor for problems. Most of the change keeps to javascript, so it does not increase the maintenance load for fproxy itself by much.

nextgens:
N2N messages allow for the transfer of files nowadays... and while I am not sure how much they have evolved since I originally implemented them, I am pretty sure that the arbitrary limitation I introduced at the time has no reason to be that low any-more. Have you tried the obvious, adding a few zeros to the two lines at L145 and
https://github.com/freenet/fred/blob/next/src/freenet/clients/http/N2NTMToadlet.java#L147 ?

LOC changes are a useful metric that is used widely in software development, whether you like it or not. It's used to estimate cost, cost of review, cost of storage, "quality", ...

ArneBab:
Seriously, after having spent the hours to restructure and debug the JS after Bombe’s comment, I’m not going to shift and re-do the approach now. If you want it done another way, you do it and file a pull-request. The right time to request a changed approach was between 15 Dec 2016 and yesterday —  between when Bombe requested restructuring of the JS and when I did it.  Sisyphos would be proud of the way this happens here right now.

nextgens:
Heh, don't blame me for (you) not having made any progress on this in the last 5 months! I am looking at it again because you claim that it is "done"...

I am the second one telling you that the 80 lines of code you have suggested for inclusion are just not good enough.
1) it doesn't address the root cause of the problem (and you are aware of it according to the initial message on this thread)
2) it has already introduced regressions (bfbb967)
3) it is ugly code-style wise: one-letter variables, doesn't refactor the constants, ...

ArneBab:
I’m not blaming you for the delay of a few months. That you only wrote after that delay was over and that you ask for killing that work again, that’s is what I’m not happy with. And no, Bombe did not say that what you comment on is bad. He wrote that what I now fixed was a problematic approach.

And your LOC metric is still bullshit. It was bullshit when people talked about lots of LOC as metric for good work, and it is still bullshit when people talk about the number of LOC as metric for bad quality.

Go and file an alternative pull-request in which you fix the root cause and ensure that nothing else breaks. I consider this one a working fix for the UI problems due to the limitation. I’m now taking a break from this, because this discussion is massively frustrating.

PR #593, opened 2016-09-04
Mark updated bookmarks with bold link; add button for dismissing updates

xor-freenet:
Based on https://github.com/freenet/fred/pull/588 to avoid merge conflicts.

Screenshot:
bookmark-notifications

PR #607, opened 2017-03-03
merge master into next

ArneBab:
This should avoid forking the history: it contains the changes released as 1476. It includes https://github.com/freenet/fred/pull/593 and https://github.com/freenet/fred/pull/588 (so if this is merged, these two can be closed).

Tests done: It builds (with gradlew) and runs for me as a working node.


xor-freenet:
You should instead file a pull request from freenet:master to freenet:next, and merge that - not from ArneBab:next to freenet:next.

(Or actually, you can just merge master into next yourself given that you are the RM and have access to do it :)

ArneBab:
I merged into next myself, but I want this merge to be reviewed by someone else besides me. Therefore I merged in my own repo and filed a pull-review to include the merge in the official repo.

ArneBab:
@nextgens is there anything left which you see as blocker for merge into next?

nextgens:
no

src/freenet/clients/http/UserAlertsToadlet.java
```diff
@@ -47,7 +50,16 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx) t
...
-		headers.put("Location", ".");
+		
+		
+		String redirect;
+		try {
+			redirect = request.getPartAsStringThrowing("redirectToAfterDisable", 1024);
+		} catch (SizeLimitExceededException | NoSuchElementException e) {
+			redirect = ".";
+		}
+		headers.put("Location", redirect);
```
nextgens: This right here is called an open-redirect vulnerability.
https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet

ArneBab: @xor-freenet could you have a stab at fixing this? It is from https://github.com/freenet/fred/pull/607/commits/e2a63110e403f71a7596b049495eb27e72a60f82

ArneBab:
@nextgens damn, that’s from https://github.com/freenet/fred/pull/607/

src/freenet/clients/http/UserAlertsToadlet.java
xor-freenet: I'm sorry I would not have expected that browsers would redirect the user to a different domain and NOT display the different domain as the location. Why on earth do they do that?

Will fix it tomorrow.

ArneBab: @xor-freenet web security is hairy.

nextgens:
@xor-freenet the java idiom is "expectedValue.equals(actualValue)" whether you like it or not; our code style guideline should be very clear about it.

There are very good reasons for it to be that way:
1) it's less likely to throw a NPE
2) if you meant '==' and wrote '=', you're more likely to notice (since the expected value is likely a constant or final variable)
3) in many languages it's faster (I wouldn't bet on it still being the case in modern java)

Bombe:
Excuse me, but `expectedValue.equals(actualValue)` is either not an idiom in Java at all or I have never come across it in 20 years – which is probably the same. Yes, I have seen people doing it in order to skip one `null` check but apart from that I believe neither order is preferred. Whatever makes your code clearer.

nextgens:
Most of fred's code uses that idiom; it's consistent with the rest of the codebase.

We used to have checkstyle running on our CI tools... that does have a check for it.
http://checkstyle.sourceforge.net/config_coding.html#EqualsAvoidNull

test/freenet/client/filter/GIFFilterTest.java
```diff
@@ -7,6 +7,7 @@
 import java.io.OutputStream;
 import java.util.Arrays;
 
+import freenet.support.HexUtil;
```
Bombe: This line on its own doesn’t make much sense.

Bombe:
The tests still don’t run on macOS (Sierra).

PR #609, opened 2017-03-08
Misc SSL related fixes for next release

nextgens:
- Update BC to 1.56
- Re-enable RSA-based ciphers
- Fix a potential click-jacking vulnerability with some browsers


nextgens:
Oh, and ofc dependencies.properties needs to be updated... but I'll let the RM do that (and adapt the release/installer scripts)

PR #612, opened 2017-03-20
add amazon root certificate for the new site, for next

ArneBab:
This is required to make plugin-updating over HTTPS work with the new site,
and we need that as contingency in case connectivity plugins break.


src/freenet/pluginmanager/PluginDownLoaderOfficialHTTPS.java
```diff
@@ -28,7 +28,13 @@
 
 public class PluginDownLoaderOfficialHTTPS extends PluginDownLoaderURL {
 	private static final String certurlNew = "freenet/clients/http/staticfiles/globalsign.pem";
-	private static final String[] certURLs = new String[] { certurlNew };
+	// amazon root certificate, so we can rekey the S3-based site more easily
+	private static final String certurlAmazon1 = "freenet/clients/http/staticfiles/AmazonRootCA1.pem";
+	private static final String certurlAmazon2 = "freenet/clients/http/staticfiles/AmazonRootCA1.pem";
```
nextgens: Four copies of the same cert isn't what we want...

nextgens:
The URLs need to change (and while you're at it, you might want to upgrade/get rid of the sha1 stuff)

nextgens:
Oh, and there's also Node.fixCertsFiles() that needs updating

ArneBab:
fixCertsFiles() seems to use what’s in the official plugin downloader

ArneBab:
Closing because it’s out of date.

PR #615, opened 2017-04-16
Build.gradle Add repos, publish and upload

D-sha:
This should fix local building if freenet-ext.jar is present in lib/freenet.
Add ability to publish locally and to mvn.freenetproject.org

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,13 +13,41 @@ configurations {
     compile
 }
 
+group = 'org.freenetproject'
+
 apply plugin: 'witness'
 apply plugin: 'java'
+apply plugin: 'maven-publish'
+apply plugin: 'maven'
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            from components.java
+        }
+    }
+    repositories {
+        maven {
+            "http://mvn.freenetproject.org"
+        }
+    }
+}
 
 repositories {
+	flatDir {
+		dirs 'lib/freenet'
+		}
     jcenter()
     mavenLocal() // TODO: use lib/ instead?
     maven { url 'https://mvn.freenetproject.org' }
+}
+
+uploadArchives {
+    repositories {
+        mavenDeployer {
+            repository(url: "http://mvn.freenetproject.org")
+        }
+    }
 }
 
 sourceSets {
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ publishing {
     }
     repositories {
         maven {
-            "http://mvn.freenetproject.org"
+            "https://mvn.freenetproject.org"
         }
     }
 }
@@ -45,7 +45,7 @@ repositories {
 uploadArchives {
     repositories {
         mavenDeployer {
-            repository(url: "http://mvn.freenetproject.org")
+            repository(url: "https://mvn.freenetproject.org")
         }
     }
 }


build.gradle
```diff
@@ -13,10 +13,30 @@ configurations {
...
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            from components.java
+        }
+    }
+    repositories {
+        maven {
+            "http://mvn.freenetproject.org"
```
Thynix: Not `https`?

Thynix:
@D-sha Ping?

D-sha:
@Thynix Sorry to stuff you around! 
Is this better?

Thynix:
I'm confused how this publishing would function. Is there magic here I'm not anticipating, or does this need an AWS S3 plugin of some description to upload artifacts?

Thynix:
Okay, I'm starting to get a better handle on this I think.

I don't know if anyone mentioned this to you, but the larger goal is for Gradle scripts to replace [this repository of shell scripts](https://github.com/freenet/scripts). That's far too large a goal for a single PR, so I'd suggest narrowing its scope to just uploading artifacts to the mvn.freenetproject.org S3 bucket. Also now that we have freenet-ext in mvn.freenetproject.org I don't think the `flatDir` is needed any more. Do you think it'd make sense to open a separate PR to focus on just uploading to S3? As of version 2.4 Gradle supports `s3://` URIs: [see the release notes](https://docs.gradle.org/2.4/release-notes.html#support-for-amazon-web-services-s3-backed-repositories).

D-sha:
Ok yes I'm starting to see the big picture now. Wow that's a lot of scripts!, a release must be a real headache? 
I totally agree with your other points.
How is freenet-ext uploaded? and what is the s3 url?
I will work on this on the weekend and see if i can manage to produce something usable....
Thanks!

Thynix:
Yep, releases sure are a pain. When I did them they took about 12 hours between starting the process and the final insert completing. Part of what rewriting them in Gradle can allow is continuing the decoupling of installer and Fred releases. They're already decoupled for Mac OS.

I think freenet-ext was a manual thing currently; I don't anticipate it changing because we're trying to stop using it. There's what should be a Maven repository at mvn.freenetproject.org; does that work as an S3 URL? freenet-ext v29 is [here](https://mvn.freenetproject.org/org/freenetproject/freenet-ext/29/freenet-ext-29.jar).

Thynix:
Based on [this documentation](https://docs.gradle.org/3.5/userguide/artifact_management.html#sec:publishing_artifacts) I have this thing that looks like a starting point; I'm not able to test it currently because I don't have write permission to the bucket.

```
uploadArchives {
    repositories {
        mavenDeployer {
            repository(url: "s3://mvn.freenetproject.org/")
        }
    }
}
```

Thynix:
I've pushed b4c20c3368a11753fabb28dd73900651b0ebd87c and 05dac3cc917108174bfda3101274650e864fa5ed; to my [gradle-publish branch](https://github.com/Thynix/fred-staging/tree/gradle-publish). With it I was able to publish [fred-1.jar](https://mvn.freenetproject.org/org/freenetproject/fred/1/fred-1.jar)!

The version remains a problem; the release scripts did it thusly, but I figure we can use `git describe`?
https://github.com/freenet/scripts/blob/master/freenet-scripts-common#L32

D-sha:
Cool that worked! but i did have some problems:
* is the repo called fred-staging? does this build to fred-staging*.jar?
* task :jar fail's at jarWithVersion because build/libs/freenet-without-version-1.jar has a different name (version number)

I'm not sure if this local / related to my setup...  Thoughts?
Apart from that I get a response from amazon so it that's great!

and the version this is where i get stuck... are you are talking about renaming after the fact?
would it be possible to start using a freenet-0.7.5-build1479.jar file and linking to it (ln -s freenet.jar freenet-x.x.x-buildxxxx.jar or windows equivalent) in the program folder? I realise this is increasing complexity and possibly making things more fragile, but this would be closer to best practices? 
and adjust the code when possible?

Thynix:
It builds to freenet.jar. I don't understand your second point.

No, I'm talking about somehow setting version when the build runs to... I guess to the output of `git describe --always HEAD`. I hope it doesn't have to be a number.

PR #616, opened 2017-05-04
Ignore Opennet peers from specified countries

Bombe:
From FMS, board freenet:

> Due to some of the recent LEA activities involving Freenet, I've created 
a little stop-gap patch that will prevent your node from connecting to 
Opennet peers in specified countries. In the Opennet options page you 
can add as many excluded country codes as you need and your node won't 
connect to Opennet peers from there. You can see the list of country 
codes in the Freenet source file IPConverter.java.
> 
> Some caveats:
> You should delete openpeers before the first time you startup with this 
patch or your node will automatically reconnect to old peers.
Doesn't work with IPv6. Mostly because Freenet only has an IPv4 country 
database.
There is a forceAddPeer in OpennetManager. I'm not sure when or if that 
gets called.

I have included SomeDude’s original patch as the first commit of this branch; however, it does not build with Java 7 (because it uses `String.join` which is Java 8+). The additional commit removes this requirement, uses a `Set` instead of a `String[]`, and cleans up some formatting and whitespace usage.

I have not tested any of this and I lack the knowledge of the internals involved in the opennet peer connection decision making process.


Bombe:
I am currently running this patch and no flag on my strangers page belongs to one of the countries I have excluded so it appears to be working. I will keep watching this.

Bombe:
Half a day later still no opennet peer from the excluded countries so this looks like it’s working.

nextgens:
I haven't looked at the patch in details... but do we update IpToCountry.dat ? Where does it come from? https://dev.maxmind.com/geoip/geoip2/geolite2/ ? When was it last updated?

Blocking peers based on a database that is years (if not decades) old is not okay!

I would expect to have at least relevant changes in dependencies.properties ... and if possible the gradle magic to pull a newer version (according to the link it's updated monthly).

Bombe:
Well, we discussed some of it on IRC and I’m still of the opinion that despite the size of the binary (we should switch to the current de-facto standard DB which clocks in at 1.3 MB gzipped, 2.7 MB unpacked) we should keep it in Git and update it periodically so it’s always fresh in new releases. However, I see that only as a task separate from this PR.

src/freenet/node/OpennetManager.java
```diff
@@ -546,6 +589,39 @@ public boolean wantPeer(OpennetPeerNode nodeToAddNow, boolean addAtLRU, boolean
...
+					if (address == null) {
+						continue;
+					}
+					InetAddress inetAddress = address.getAddress(false);
+
+					IPConverter ipConverter = IPConverter.getInstance(NodeFile.IPv4ToCountry.getFile(node));
+					try {
+						Country country = ipConverter.locateIP(inetAddress.getAddress());
+						if (country != null) {
+							if (excludedCountries.contains(country.toString().toUpperCase())) {
```
xor-freenet: This is missing thread synchronization against concurrent modification of excludedCountries, i.e. the user changing the configuration meanwhile.

Quote from https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html:
> Note that this implementation is not synchronized.

src/freenet/node/OpennetManager.java
```diff
@@ -278,6 +286,41 @@ public OpennetManager(Node node, NodeCryptoConfig opennetConfig, long startupTim
...
+				}
+				return joined.toString();
+			}
+
+			@Override
+			public void set(String newValue) /* throws InvalidConfigValueException, NodeNeedRestartException*/ {
+				excludedCountries.clear();
+				if ((newValue != null) && !newValue.equals("")) {
+					for (String excludedCountry : newValue.split(",")) {
+						excludedCountries.add(excludedCountry.toUpperCase());
```
xor-freenet: Please use String.trim() on excludedCountry.

src/freenet/node/OpennetManager.java
xor-freenet: Additionally it is inefficient to always to the toUpperCase() before each contains() query of the HashSet.
Instead the initialization of the HashSet should make the set contain the same case as which country.toString() returns, then case conversion only has to happen once at init, not at every query.

src/freenet/node/OpennetManager.java
```diff
@@ -546,6 +589,39 @@ public boolean wantPeer(OpennetPeerNode nodeToAddNow, boolean addAtLRU, boolean
...
+						continue;
+					}
+					InetAddress inetAddress = address.getAddress(false);
+
+					IPConverter ipConverter = IPConverter.getInstance(NodeFile.IPv4ToCountry.getFile(node));
+					try {
+						Country country = ipConverter.locateIP(inetAddress.getAddress());
+						if (country != null) {
+							if (excludedCountries.contains(country.toString().toUpperCase())) {
+								Logger.error(this, "Not connecting to Opennet peer " + inetAddress.toString() + " in country " + country.toString());
```
xor-freenet: error() should be used to indicate bugs in the code, not for things which the user told us to do.
I'd say:
- Use normal() if the frequency of this occurring is above "every few seconds"
- If it occurs more frequently than that, use "if(logMINOR) Logger.minor(...)"

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1157,6 +1157,8 @@ Node.oneConnectionPerIP=Limit to one connection per address? (friends) (NOT RECO
 Node.oneConnectionPerIPLong=Automatically drop friend connections when other friends have the same (external) IP address.
 Node.opennetEnabled=Enable insecure mode (automatically connect to untrusted nodes)?
 Node.opennetEnabledLong=Enable insecure mode (aka opennet)? If this is enabled, your Freenet node will automatically exchange node references with other untrusted nodes (Strangers as opposed to Friends). But this means that the fact that you are running a node is no longer private, and many attacks are much easier. If you know enough people running Freenet, you should stick to trusted (Friends) connections to them, and turn this off.
+Node.opennetExcludedCountriesShort=Ignore Opennet peers from these countries
+Node.opennetExcludedCountriesLong=A comma separated list of country codes that will prevent your node from connecting to Opennet nodes originating in these countries.
```
xor-freenet: There is a giant zoo of country code standards available.
This should at least mention which standard the codes are chosen from. Ideally a link to the list of codes should be included.
Notice that ```src/freenet/l10n/ISO639_3.java``` contains an API for obtaining a list of ISO639-3 codes in case this is based on ISO639-3.

xor-freenet: There are probably multiple standards for country codes.
This should at least mention which standard the codes are chosen from. Ideally a link to the list of codes should be included.

src/freenet/node/OpennetManager.java
xor-freenet: Further, please validate whether the user's input country codes are valid before storing them in the configuration.

src/freenet/node/OpennetManager.java
```diff
@@ -278,6 +286,41 @@ public OpennetManager(Node node, NodeCryptoConfig opennetConfig, long startupTim
...
+					for (String excludedCountry : newValue.split(",")) {
+						excludedCountries.add(excludedCountry.toUpperCase());
+					}
+				}
+			}
+		});
+		String initialExcludedCountries = sConfig.getString("excludedCountries");
+		excludedCountries.clear();
+		if (initialExcludedCountries != null && !initialExcludedCountries.equals("")) {
+			Collections.addAll(excludedCountries, initialExcludedCountries.split(","));
```
xor-freenet: This should also use String.trim() on each country in case the user edits the ini file manually.

src/freenet/node/OpennetManager.java
```diff
@@ -546,6 +589,39 @@ public boolean wantPeer(OpennetPeerNode nodeToAddNow, boolean addAtLRU, boolean
 				return false;
 			}
 		}
+		// Don't connect to peers in excluded countries
+		if (nodeToAddNow != null) {
```
xor-freenet: I don't know how the geo-lookup code works but I'd blind guess it may not be very cheap given that the 32 bit IPv4 address space is already exhausted.
Thus as an optimization I'd add ```&& excludedCountries.size() > 0``` to this (and take into consideration that this needs to be thread-safe as well as any other accesses to excludedCountries, so I'd probably wrap the whole if() in a locked section)

xor-freenet:
Here's a comment on [line 638 ](https://github.com/freenet/fred/pull/616/files#diff-d76b8606fca32bfb80b9bf768cefda41R638) - I cannot use GitHub's feature for attaching the comment directly to it because it is too far away from what's modified by the PR:

```if(crypto.allowConnection(nodeToAddNow, addr))```

As you can see on that line, the conceptually very similar code to disallow more than 1 peer per IP address does put the "is a peer with this IP allowed?" decision into a function in a different class.
Why doesn't this PR put it into that same place?

Thynix:
I don't think that conceptually this is a thing I'd want to merge. I'd much rather have opennet peers not get FOAF information, and so not have peer count. This doesn't prevent rate-based attacks, but it makes it more difficult, and nothing prevents LE nodes from being overseas VPSs. Thoughts?

Thynix:
As long as it's not enabled by default I'd be okay with merging it if people really want it, but I don't think it's worth much beyond a false sense of security. Blocking the US will not fix opennet's security problems.

nextgens:
I do believe that in any sane legal system, it's not that simple for LEO to operate overseas... Limits to jurisdiction exist for a reason...

What's not sensible is to encourage users to break their connectivity with a patch that uses a geoip DB that is decades old... @Bombe is working on that and I do think we should assist where possible (if only because that means modernizing the decade old codebase)

nextgens:
Performance wise, there is an argument to do exactly the opposite. Back in the days it was called p4p

Maybe we should provide the option and pick a default based on the security level the user picks

https://en.wikipedia.org/wiki/Proactive_network_provider_participation_for_P2P

src/freenet/node/OpennetManager.java
xor-freenet: Suggested thread safety fix:
Make the excludedCountries reference volatile, and do writes to the HashSet by replacing the whole set with a new one, i.e. making the reference point to a new HashSet object.

That should be the most efficient concurrency tool for our situation of very many reads vs very few writes: The only low level effect on the reads should be that the excludedCountries variable has to be fetched from main memory every time instead of being cached, locks won't need to be acquired. The cost of creating a new HashSet upon every write can be ignored for that benefit of lock-freeness as users will rarely change the config.

xor-freenet:
@Thynix wrote:
> I don't think that conceptually this is a thing I'd want to merge. I'd much
> rather have opennet peers not get FOAF information, and so not have peer
> count. This doesn't prevent rate-based attacks, but it makes it more
> difficult, and nothing prevents LE nodes from being overseas VPSs. Banning
> countries is too blunt. Thoughts?

Law enforcement isn't our users' only threat model. Once the issues are fixed please do merge it - if you like something to be done about your concerns of whether this is efficient then possibly with the tradeoff of only displaying the option in advanced mode (if that isn't already the case?).

I had sent you an encrypted email with another threat model. If anyone wants to know about it please ask me personally.

Thynix:
Alright. You all can focus on what goes in the next build (and merging master and next) then. I'll focus on getting a build script for AWS together.

nextgens:
@xor-freenet 
> I had sent you an encrypted email with another threat model. If anyone wants to know about it please ask me personally.

Seriously? contribute in a useful manner or abstain from doing it at all...

Thynix:
@xor-freenet Could you please discuss the different threat model you mention publicly?

xor-freenet:
@Thynix wrote:
> @xor-freenet Could you please discuss the different threat model you mention publicly?

Yes. According to various news sources, e.g. [this](https://www.chicagobusinesslitigationlawyerblog.com/2017/02/lawyers-targeted-porn-users-allegedly-extortionate-lawsuits-now-face-criminal-charges.html) one, the following happened:
> Lawyers Who Targeted Porn Users With Allegedly Extortionate Lawsuits Now Face Criminal Charges
> [...] Two attorneys were allegedly using the courts to extort money from their opponents, rather than their clients. [...]
> [...] The two attorneys allegedly created sham companies in 2010, which they used to acquire the rights to certain pornographic films, some of which they made themselves. [...]
> [...] [They] allegedly uploaded their copyrighted porn to file-sharing websites, knowing people would download their porn. Steele and Hansmeier would then file a copyright lawsuit against the individual on behalf of their “clients” – which were, in fact, the companies they themselves owned.
> [...] [They] allegedly offered to settle the lawsuit for about $4,000. The alternative was to face public exposure and fines that could potentially get as high as $150,000.
> [...] [They]Allegedly went home with approximately $6 million as a result of the allegedly false and misleading lawsuits they filed against unsuspecting porn users or purported porn users all over the country. 

**TL;DR:** Copyright trolls bought copyright to porn (and actually even created their own), uploaded it to filesharing apps on their own, and the demanded money from the downloaders, threatening to sue them otherwise, hoping they comply as the victims are usually ashamed about the porn.

Further, In Germany, where I live, it has even gotten as far as people being accused of porn copyright infringement even though they didn't do it at all, see [this](https://www.wbs-law.de/eng/warning-letter-filesharing/huge-wave-fake-file-sharing-warning-letters-hits-germany-54150/) article of a German lawyer. In that case it was only random email spam to spread viruses, not actual legal targeting of filesharing users, but I think it is only a matter of time until lawyers start to harvest IPs of random filesharers and deliver completely false accusations to them.

Freenet may be very interesting for those trolls:
In addition to threatening to sue for copyright infringement they could also invent accusations of having observed that the users spread even more illegal content, and threaten to report them to law enforcement for that if they don't pay.

Now why is this relevant to this Opennet PR?
As opposed to law enforcement who is legally bound to prosecute **everyone** they consider as criminal, copyright trolls don't care **who** their victims are.
Small technical barriers such as blocking the country where their machines are may very well be sufficient to deter them from choosing a certain victim. They'll just move on to the next potential victim.

ArneBab:
What’s the state of this? This is kinda urgent for quite many of our users, given that they are being targeted by partly clueless, under-resourced attackers with a high probability of hitting the wrong target, who are prone to unleash the full power of the law on innocents.

And I hope that we can do releases from next within at most two weeks.

xor-freenet:
> What’s the state of this? 

Without re-reading my review results I can tell you at least one of them is blocking - this isn't thread-safe.

So someone needs to fix at least that, and I would ask whoever does it to at least have a look at the other ones as well and if you don't intend to fix them then please provide an explanation why.

Once it is fixed I would also like to see this merged, I personally consider it as really useful.

desyncr:
Bumping this. I have some reservations thou: What about people on ignored countries? Will they have problems to connect to opennet? Will this create separate networks?

Regarding geoip db: It should update automatically, not sure if it's possible with maxmind, thou.

ArneBab:
Is someone interested in picking this up?

ArneBab:
Will someone take this up again? @Bombe are you interested in finishing this? The main points I see:

- What about attackers who choose IPs without country result? Could there be a safe fallback? Alternative: Provide a way to specify "block nodes without country result".
- Needs an example in the description for a blocking config (i.e. blocking the US and north korea and "no country result").

With these two taken care of, I’d be up for merging this.

ArneBab:
@desyncr would you be willing to take this up?

PR #617, opened 2017-05-06
Allow caching of CHK and SSK keys

Bombe:
This branch adds a configurable switch to FProxy which allows it to send the data of CHK and SSK keys with a header that will enable the browser to cache the data for 30 days.


src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -139,7 +140,7 @@ public void handleMethodPOST(URI uri, HTTPRequest req, ToadletContext ctx) throw
 		}
 	}
 
-	public static void handleDownload(ToadletContext context, Bucket data, BucketFactory bucketFactory, String mimeType, String requestedMimeType, String forceString, boolean forceDownload, String basePath, FreenetURI key, String extras, String referrer, boolean downloadLink, ToadletContext ctx, NodeClientCore core, boolean dontFreeData, String maybeCharset) throws ToadletContextClosedException, IOException {
+	private void handleDownload(ToadletContext context, Bucket data, BucketFactory bucketFactory, String mimeType, String requestedMimeType, String forceString, boolean forceDownload, String basePath, FreenetURI key, String extras, String referrer, boolean downloadLink, ToadletContext ctx, NodeClientCore core, boolean dontFreeData, String maybeCharset) throws ToadletContextClosedException, IOException {
```
nextgens: I bet that this will break some plugins

nextgens:
I like it; I would make the default depend on the security level though

src/freenet/clients/http/FProxyToadlet.java
Bombe: A `grep` over my source code directory (which contains quite a number of our plugins) does not reveal any hits for `handleDownload`. However, yes, with the way we currently expose *everything* to plugins changing *anything* might break a plugin.

Bombe:
I don’t think our `Config` does support a default value that depends on a different section because we don’t have a value for “not set” — unless that is what the `forceWrite` parameter is all about?

nextgens:
forceWrite is to ensure that it is written to the config file (even if the value set is the default).

The right way to do it is to modify the getter; If the value is default AND the security level low enough, then you might return something that isn't the "stored" value.

Bombe:
That still gives me no way to recognize if a value is the default or whether it has been intentionally set to the same value as the default.

nextgens:
You're right.

I wouldn't care in this instance (if the security level is low enough)

Bombe:
Well, the default is not to cache (i.e. the current behaviour) and there is a warning on the option’s description. I believe that should suffice. :)

PR #618, opened 2017-05-30
README: focus on developers and update for Gradle

Thynix:
I changed the readme to focus on developers wishing to build the software instead of users.

GitHub's render of it is [here](https://github.com/freenet/fred/blob/readme/README.md).


README.md
```diff
@@ -1,109 +1,66 @@
...
+recommended by the Gradle project. If you trust the version we've committed you can build
+immediately:
+
+#### POSIX-based:
+
+    $ ./gradlew jar
+
+#### Windows:
+
+    > gradlew.bat jar
```
nextgens: Are you sure that the above doesn't work on windows too? iirc ./ will be ignored and .bat inferred

README.md
```diff
@@ -1,109 +1,66 @@
...
+
+* Arch: `sudo pacman -S gradle`
+* CentOS: `sudo yum install gradle`
+* Debian-based / Ubuntu: `sudo apt-get install gradle`
+* Windows: official installation directions are [here](https://gradle.org/install#manually)
+
+then:
+
+    $ gradle jar
+
```
nextgens: I would get rid of the whole "System Gradle" section; Odds are the distro will never have the version we want... and gradle's syntax seems to be evolving fast

README.md
Thynix: Looks like it doesn't. cmd infers `.bat` but does not ignore `./`.

Thynix: PowerShell does ignore leading `./`. Can we assume people are using PowerShell?

Thynix: I guess we can tell them to use PowerShell for the purposes of these directions.

README.md
Thynix: In general I'd agree, but given how paranoid/careful elements of our userbase can be I'd like to include this section anyway to make it easier to not trust binaries we commit. I can add a note how it's not supported if the Gradle version is different from that of gradlew. Thoughts?

nextgens: Then get them to delete/re-generate the wrapper from the system's gradle... that way we can keep a single set of instructions for all users.

It should be "gradle wrapper ; goto 1" rather than "gradle jar"

Thynix: If they're doing it out of security concerns, I doubt they'll be all that enthused about downloading a non-checksum-verified jar using their PGP-verified system package. I'd be up for leaving them to figure it out?

nextgens: I wouldn't document how to do it with distro stuff at all.

It's not like if the dependencies gradle will fetch were authenticated to a higher standard ;)

Thynix: Fair enough. I've removed the section. Anything else?

nextgens:
LGTM

Thynix:
Merged as 799f7fd6bc833a0d9ebe20dd3488cda54250b839; thanks for your review!

PR #619, opened 2017-07-25
Add support for `gradle publish` to S3 Maven repo

Thynix:
Thanks to @D-sha for getting this started in #615 and @nextgens for pointing out the existing `git describe` support (hacky though it may be; dunno if it works on Windows) to avoid the need for Gradle plugins. (I am not aware of an automated mechanism for verifying Gradle plugins like `gradle-witness` does for dependencies.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,13 +13,38 @@ configurations {
     compile
 }
 
+group 'org.freenetproject'
+
 apply plugin: 'witness'
 apply plugin: 'java'
+apply plugin: 'maven-publish'
+
+archivesBaseName = 'freenet'
+version = '1'
 
 repositories {
     jcenter()
     mavenLocal() // TODO: use lib/ instead?
     maven { url 'https://mvn.freenetproject.org' }
+}
+
+
+// AWS credentials can be set below, gradle.properties or command line
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            from components.java
+        }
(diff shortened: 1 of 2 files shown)


Thynix:
I've [published jars](https://mvn.freenetproject.org/org/freenetproject/fred/build01475-207-g916e5/fred-build01475-207-g916e5.jar) with this and have yet to reproduce the failure locally. If none of you have ideas I'll increase Travis' verbosity at some point, but my schedule is filling up for the near future and I shouldn't be taking this time already.

D-sha:
I think this was similar to what i was having.  
gradle on my system is producing freenet-without-version-1.jar which doesnt get caught by [jarWithVersion](https://github.com/freenet/fred/blob/next/build.gradle#L85)  
ie i think the appended version number is the problem.  
Maybe you have something set locally that makes it work for you?

Thynix:
Are you using Windows?
-------- Original Message --------

D-sha:
I run Arch Linux, openjdk8
I attached a [log.txt](https://github.com/freenet/fred/files/1175697/log.txt) with the shell output with the commands hopefully it makes sense.
Using the ./gradlew wrapper makes no difference and gradle publish stops at the same point.

Cheers

nextgens:
I have made it happen in d8e6a25e64c74733368b04a667ef34c78b0b6807

PR #621, opened 2017-09-03
1479: preparing for plugin+l10n update

ArneBab:
This pull request tracks work on master to create a small release with plugin+l10n updates. This is necessary to avoid being blocked by big changes in next while our clearnet failover is still broken from the website infrastructure change.


src/freenet/pluginmanager/OfficialPlugins.java
```diff
@@ -110,7 +110,7 @@ public OfficialPlugins() {
 					.inGroup("communication")
 					.minimumVersion(18)
```
xor-freenet: Why not increment this?
Disclaimer: i haven't looked up the purpose of this value!
So check how previous releases by e.g. @Thynix  / toad_ did it.

xor-freenet: Also while you're at it you can increase the value of the minimumVersion() and of the USK at the below testing version of WoT.

NEWS.md
```diff
@@ -11,6 +11,8 @@ next ():
 
 master ():
 
+- plugins: WebOfTrust build0019
```
xor-freenet: I've invested hours in writing a comprehensive changelog, I would appreciate it very much if you could link it here and in any other announcements surrounding the release:
https://git.io/v5EsM

src/freenet/pluginmanager/OfficialPlugins.java
ArneBab: If I increment the minimumVersion, it will block startup if the key isn’t accessible (i.e. because the node does not have connections yet).

xor-freenet: Will it only block startup of WoT? That would be OK I suppose if the WoT build fixes significant issues - and in fact build0019 does fix a bug which can block startup of WoT for like half an hour (if not hours, the half an hour is only what I have observed), so nothing would be lost anyway.

ArneBab: IIRC it would block startup of fproxy — but recommendedVersion(19) won’t.

xor-freenet: > IIRC it would block startup of fproxy — but recommendedVersion(19) won’t.

Oh ok nice, then if you or @Thynix validate this to be correct then I would say we go for recommendedVersion(). But please do validate it somehow or otherwise do as previous RMs did as this stuff could have dangerous results like the one you just mentioned.

ArneBab: the first step is a testing release :)

ArneBab: (re version: I actually have notes on my freenet-release-management website which say that)

xor-freenet: Thanks. The WoT build has been released to the testing channel for a whole year, does it really need more testing?

xor-freenet: Oh I see you already marked it as recommended. Thanks.
Please do the same for the *Testing channel.

xor-freenet:
You've pushed changes to the core networking.
I thought we agreed there should be a release which solely fixes update.sh/update.cmd before we release any further stuff which can break the network!?

ArneBab:

> You've pushed changes to the core networking.
> I thought we agreed there should be a release which solely fixes update.sh/update.cmd before we release any further stuff which can break the network!?

I pushed a config change which should not be able to break the network,
except for typos which should turn up in testing. It just adjusts our
settings to provide better speeds with the slower network.

This change reduces the number of peers for the current slowest
connections from 10 to 7, which is enough by far to get good
connectivity. It should reduce flopping connections for the slowest and
the fastest peers without changing any functional behavior.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

src/freenet/l10n/freenet.l10n.fi.properties
```diff
@@ -908,7 +898,6 @@ PproxyToadlet.pluginNotFoundReloadTitle=Liitännäistä ei löydetty (uudelleenl
 PproxyToadlet.pluginNotFoundUpdatingTitle=Liitännäinen ei löytynyt (päivitetään)
 PproxyToadlet.pluginNotFoundUpdating=Liitännäinen, jota yritit päivittää, ei latautunut!
 PproxyToadlet.pluginSourceChoice=Voit noutaa lisäosan Freenetistä tai keskuspalvelimiltamme:
-PproxyToadlet.pluginSourceFreenet=Hae Freenetistä. Tämä ei ole jäljitettävissä ja turvallista, mutta voi olla hidasta.
```
ArneBab: This is strange. I’m not sure why this got deleted from the translations. I’ll have to go through the translation changes again to see whether they are all OK.

xor-freenet: I'm not sure how familiar you're with the way our l10n works so here is the most simple explanation:
Whenever an English source string changes we remove all existing translations of it so there always either is an up-to-date translation or none. Transifex does implement that IIRC.
So perhaps the English source string changed?

By the way, why are you still able to update translations of a master-based build anyway? Shouldn't Transifex already be updated to contain the English source strings for next, not master, and hence updating translations should be invalid until we deploy next?

ArneBab: I’m able to use that, because next and master differ by very little. I did all possible backports to avoid blocking our moving forward by stuff which blocks next.

It’s possible that the English source changed, yes. Thank you for the explanation!

NEWS.md
```diff
@@ -11,13 +11,22 @@ next ():
 
 master ():
 
+- optimized network settings for the new structure since the link
+  length fix : less peers for the same bandwidth should result in
+  higher throughput per connection. This allows for less powerful
+  devices to join (with low bandwidth settings) and should provide
+  better bandwidth utilization for very fast nodes.
+- Re-enable RSA-based ciphers
```
xor-freenet: Shouldn't changes in core security areas such as cryptography mention things such as
- why they were done
- which subsystems they affect
?

NEWS.md
xor-freenet: ^^^ Ping for the above comment.

xor-freenet:
I'm sorry it seems GitHub messed up the review UI to look like you're commenting on a single commit but it will actually not associate the comment with the commit - my previous comment about commit messages and bookmark stuff was about commit d689190f1c291e943550a05c1296803aef2dfa3b.

ArneBab:
noted now that I maintain the stats bookmark site, same for RSA

ArneBab:
@xor-freenet could you check (and maybe ack) the new changes?

PR #622, opened 2017-09-13
A really samll edit to make the build-clean.xml fit the README.build.md better

callmexss:
when i built the release edition 'build01478' i didn't past all the given tests. So i intended to skip the tests by following the instruction on the screen: 
`ant Dtests.skip=true` 
but it didn't work for me, it still executed the test progress. Than i looked up the related build files and i found in the `build-clean.xml` file, in which the value was "test.skip=true". After i edited it to "tests.skip=true", i finally got the "freenet.jar" which works fine for me. Thats all, this is my first PR, hope it can help :)

### DIFF
--- a/build-clean.xml
+++ b/build-clean.xml
@@ -308,7 +308,7 @@ maintainers may prefer to use this instead of build.xml.
 
 	<target name="package" depends="unit, package-only" description="build standard binary packages (Freenet daemon)"/>
 
-	<target name="unit-build" depends="build" unless="${tests.skip}">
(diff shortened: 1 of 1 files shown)


xor-freenet:
Thanks but development happens on the branch "next", not "master" - so what is on master can be outdated, and in fact in this case is: We got rid of the Ant builder in favor of a Gradle one.
(Gradle has native support for skipping tests with "-x test")

ArneBab:
Thank you! I fixed the README, since people will already depend on the value here.

PR #623, opened 2017-10-11
Travis CI: Fix for OpenJDK 7 by not using wrapper; shows our own tests fail on OpenJDK 7

xor-freenet:
**NOTICE**: OpenJDK 7 is still the default JRE on a maintained LTS release of a major operating system, namely Ubuntu 14.04, so it DOES make sense to test against it.

Fixes these failures of the Gradle wrapper:
https://travis-ci.org/xor-freenet/fred-staging/jobs/286721010#L461-L511

This shows that freenet.crypt.KeyGenUtilsTest fails on OpenJDK 7:
https://travis-ci.org/freenet/fred/jobs/334276152#L643-L656

Notice: Instead of this PR we could also try updating the Gradle wrapper
binaries to see whether it fixes the failure. I'm not sure how I'm
supposed to obtain them so I did this PR as it does have some benefits
to not use the wrapper:
- Gradle is already installed in the Travis environment so it's faster
to not download it using the wrapper.
- We outsource the maintenance of figuring out which Gradle version to
use to be compatible with the rest of Travis.

Notice: The failure to build on Java 9 is an issue with Travis CI, it will likely go away in a few days, see:
https://github.com/travis-ci/travis-cookbooks/pull/967

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ cache:
 
 before_install:
   - echo -e "org.gradle.parallel = true\norg.gradle.jvmargs=-Xms256m -Xmx1024m\norg.gradle.configureondemand=true\ntasks.withType(Test) {\n maxParallelForks = Runtime.runtime.availableProcessors()\n}" > gradle.properties
-  # Workaround for openjdk7: Don't use gradlew. When removing this also replace all instances of the "gradle" command in this file with "./gradlew"!
+  # No need to use the binary Gradle download script we ship, Travis CI has gradle installed already.
   - rm ./gradlew
 install: "gradle jar"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,15 +21,19 @@ cache:
   - $HOME/.gradle/caches/
   - $HOME/.gradle/wrapper/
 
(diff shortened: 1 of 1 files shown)


nextgens:
TL;DR;

This doesn't fix anything... get travis to go green if you want it to be considered for inclusion.

xor-freenet:
> This doesn't fix anything... get travis to go green if you want it to be considered for inclusion.

Oh it does fix the fact that your changes of https://github.com/freenet/fred/commit/ebb6e33de4a084e9604d1c5e2921b016f09c4d0d were hiding that our unit tests fail on java 7 instead of fixing the underlying breakage which the tests diagnose.
I'm not qualified to fix crypto code such as you are, I can only document that it is broken by shipping this PR.

Java 7 is still shipped and maintained on Linux distributions which are still in active support, e.g. Ubuntu 14.04.
Sticking our head into the sand and ignoring that our code breaks on it doesn't make the issue go away.

nextgens:
Again, this doesn't "fix" anything... and changes the behaviour of gradle from something predictable that we control to something unpredictable that we don't (with wide ranging effects, including breaking deterministic builds)...

Fundamentally, it fails to both:
1. demonstrate that there is a problem (a trivial one-line commit that shows that before the tests did pass and after they don't)
2. fix it

PR #625, opened 2017-10-17
Cryptbucket check nonce length

ArneBab:
addresses https://freenet.mantishub.io/view.php?id=6363

based on the patch therein. For review and testing.


nextgens:
AEADOutputStream talks about nonce[0] &= 0x7F but doesn't enforce it in its constructor; IMHO we want/need both

nextgens:
Let me be more clear: either we try to "fix" what we get (masking the bit we discard) or we enforce that we never create a bucket with it set... right now we don't.

ArneBab:
@nextgens could you file a new pull-request which does what you want?

nextgens:
I will try to have a look this week-end

PR #627, opened 2018-03-03
Fix freenet n2n handling

Bombe:
<@ArneBab> Hi, I created some patches to fix n2n handling (recommending files, dis- and reassembling large n2ns and only showing direct file transfer in advanced mode, because its limitations have confused several of my new darknet contacts and we have working upload recommendation, which is a better model for sharing, see thecommits for a rationale).


src/freenet/node/DarknetPeerNode.java
```diff
@@ -1466,19 +1479,49 @@ public int sendFileOffer(HTTPUploadedFile file, String message) throws IOExcepti
...
 		}
-		N2NTMUserAlert userAlert = new N2NTMUserAlert(this, text, fileNumber, composedTime, sentTime, receivedTime);
+		long msgid = fs.getLong("msgid", Long.MAX_VALUE);
+		List<UserAlert> merged = new ArrayList<>();
+		if (msgid != Long.MAX_VALUE) {
+			// merge existing user alerts
+			// TODO: This costs time linear to the number
+			// of known alerts, so receiving a large text
+			// message is quadratic in its size!
+			for (UserAlert userAlert : node.clientCore.alerts.getAlerts()) {
```
Bombe: I think I understand what you’re doing here but I also think it doesn’t work because you only locate the alerts with `composedTime + 1` and `composedTime - 1` which may not be enough by far.

I would recommend the following:

a) break this out into a separate method.
b) get the list of alerts, filter it (both by class and msgid), then sort it by composeTime.

Now merging adjacent alerts should be a breeze (and about O(n + n lg n), if I got that right).

src/freenet/node/useralerts/N2NTMUserAlert.java
```diff
@@ -22,38 +22,50 @@
...
+		super(true, null, null, null, null, UserAlert.MINOR, true, null, true, null);
+		this.messageText = message;
+		this.fileNumber = fileNumber;
+		this.composedTime = composedTime;
+		this.sentTime = sentTime;
+		this.receivedTime = receivedTime;
+		this.peerRef = sourcePeerNode.getWeakRef();
+		this.sourceNodeName = sourcePeerNode.getName();
+		this.sourcePeer = sourcePeerNode.getPeer().toString();
+		this.msgid = Long.MAX_VALUE;
```
Bombe: Is there a specific reason for choosing `Long.MAX_VALUE` as magic value for a missing msgid? I would prefer using `0` or `-1` but it’s not a deal breaker. :)

src/freenet/node/useralerts/N2NTMUserAlert.java
```diff
@@ -22,38 +22,50 @@
...
+		this.peerRef = sourcePeerNode.getWeakRef();
+		this.sourceNodeName = sourcePeerNode.getName();
+		this.sourcePeer = sourcePeerNode.getPeer().toString();
+		this.msgid = msgid;
+	}
+
+	public N2NTMUserAlert(DarknetPeerNode sourcePeerNode, String message, int fileNumber, long composedTime,
+			long sentTime, long receivedTime) {
+		super(true, null, null, null, null, UserAlert.MINOR, true, null, true, null);
+		this.messageText = message;
```
Bombe: Oh, also, maybe replace this constructor with a call to the other constructor?

src/freenet/node/useralerts/N2NTMUserAlert.java
```diff
@@ -106,21 +116,55 @@ private String l10n(String key, String pattern, String value) {
...
+
+	/**
+	 * @return the composedTime
+	 */
+	public long getComposedTime() {
+		return composedTime;
+	}
+
+	/**
+	 * @param composedTime the composedTime to set
```
Bombe: This seems to be in the wrong place. :)

src/freenet/l10n/freenet.l10n.de.properties
```diff
@@ -1037,7 +1037,7 @@ N2NTMToadlet.sendMessageShort=Nachricht senden
 N2NTMToadlet.sendStatus=N2NTM-Sende-Status
 N2NTMToadlet.sent=Nachricht wurde an den Partner gesendet
 N2NTMToadlet.sentTitle=Gesendet
-N2NTMToadlet.tooLong=N2NTMs sind auf 1024 Zeichen beschränkt
+N2NTMToadlet.tooLong=N2NTMs sind auf 131.072 Zeichen beschränkt
```
Bombe: I wouldn’t change this text for languages that you don’t know (putting this on the German text because I know that you actually do know it) because other languages may have different rules for numbers ending in -2 as opposed to numbers ending in -4. :)

src/freenet/node/Version.java
```diff
@@ -50,7 +50,7 @@
 	public static final String protocolVersion = "1.0";
 
 	/** The build number of the current revision */
-	private static final int buildNumber = 1478;
+	private static final int buildNumber = 1480;
```
Bombe: Even if this might be correct it should only be changed by a merge when we actually make a new release, I guess…

xor-freenet:
@ArneBab Thanks for providing this!

Related bugtracker entries, please include in review and close after merge:
https://freenet.mantishub.io/view.php?id=1857
https://freenet.mantishub.io/view.php?id=6178

I don't have time for reviewing the whole thing so I'd like to ask whether this fixes the issue which I consider the single most broken aspect of N2NTMs (as I don't see any indication of that in the commit messages):
Does this add serialization of unsent N2NTMs to disk so they are not lost if the user restarts the node / his machine before they have been sent?
If not could you perhaps provide that as well?

You can use Java serialization for it as fred already does use that for purge-d4bo. Or for coherency stick to what the stuff which serializes received messages uses. In both cases store them in the same place.
And please obtain a key from fred to encrypt the files on disk, we already have too many things which don't obey disk encryption.
Bugtracker entries:
https://freenet.mantishub.io/view.php?id=7008
https://freenet.mantishub.io/view.php?id=669
https://freenet.mantishub.io/view.php?id=1461

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1393,12 +1395,23 @@ public int sendDownloadFeed(FreenetURI URI, String description) {
...
-		fs.put("composedTime", now);
-		sendNodeToNodeMessage(fs, Node.N2N_MESSAGE_TYPE_FPROXY, true, now, true);
-		this.setPeerNodeStatus(System.currentTimeMillis());
+		long msgid = node.random.nextLong();
+		// split large messages
+		int requiredN2nCount = 1 + (message.length() / 1024);
+		String messagePart;
+		for (int i = 0; i < requiredN2nCount; i++) {
+			messagePart = message.substring(i * 1024,
+							Math.min((i+1) * 1024,
```
xor-freenet: Non-blocker:
You can use "import static" to be able to use self-explanatory functions like min() without prefixing them with the parent class' name. Java isn't that noisy anymore when using all its features to full extent....
In Eclipse putting the cursor on the function name and pressing CTRL+SHIFT+M automatically adds the import.

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1393,12 +1395,23 @@ public int sendDownloadFeed(FreenetURI URI, String description) {
...
+								 message.length()));
+			SimpleFieldSet fs = new SimpleFieldSet(true);
+			fs.put("type", Node.N2N_TEXT_MESSAGE_TYPE_USERALERT);
+			fs.putSingle("text", Base64.encodeUTF8(messagePart));
+			fs.put("msgid", msgid);
+			// increment compose time to allow sorting messages, this does not change the time for
```
xor-freenet: Some words are missing at the end of the comment.

src/freenet/node/useralerts/N2NTMUserAlert.java
xor-freenet: - The entropy of all the JavaDocs in this areas is very low anyway, they only repeat what's in the function names, we could save the space for something which really needs documentation.

- For docs which fit into one line I'd format them like this to require less scrolling:
```java
/** Some documentation */
public void function() {
```

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1466,19 +1479,58 @@ public int sendFileOffer(HTTPUploadedFile file, String message) throws IOExcepti
...
+                        // TODO: This costs time linear to the number
+			// of known alerts, so receiving a large text
+			// message is quadratic in its size! Worst
+			// case cost for transfer of a single huge
+			// n2ntm (when receiving every second message)
+			// is about 0.5 N*N, with N at most
+			// 128. However even sending a 64kiB file
+			// (DarknetPeerNode.java) does not cause
+			// noticeable load on the receiving system.
+                        synchronized (text) {
```
xor-freenet: Am I understanding the motivation of this code correctly like that:
Because you didn't write code to intercept split messages to not display a UserAlert until the full message is available the split messages are displayed immediately as partial incomplete UserAlerts and pieced together as more messages arrive?

If yes then this is a big usability issue which I'd consider a blocker:
By exposing the user to randomly cut messages without any apparent reason it forces the user experience to obey the internal technical architecture. It should be precisely the other way around: The internal technical architecture should be forced to comply to produce results which are pleasant to the user = Messages should not be displayed until they are complete.

(Or in any other words: Think about the dozen instant messengers which were popular in the past 20 years - did any of them do this? We'll make it rain bug reports upon ourselves if we ship this :) I'm actually amazed, you invented a new kind of hack here :)

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1393,12 +1395,23 @@ public int sendDownloadFeed(FreenetURI URI, String description) {
...
+		long msgid = node.random.nextLong();
+		// split large messages
+		int requiredN2nCount = 1 + (message.length() / 1024);
+		String messagePart;
+		for (int i = 0; i < requiredN2nCount; i++) {
+			messagePart = message.substring(i * 1024,
+							Math.min((i+1) * 1024,
+								 message.length()));
+			SimpleFieldSet fs = new SimpleFieldSet(true);
+			fs.put("type", Node.N2N_TEXT_MESSAGE_TYPE_USERALERT);
```
xor-freenet: This will sound theoretical at first but it has a real consequence already which I will explain later on:
You changed the network protocol here without introduction a protocol version field to indicate a different version or at least changing the ```type``` value here to something different.
This means it's impossible for peers to recognize whether they're talking to something which speaks the old or the new protocol version and and act accordingly - which can cause trouble in handling cases (or even bugs) where the protocol version needs to be identified.

The already existing consequence is that the UserAlert hack which I already commented is necessary: You cannot detect whether a message is split until more messages with the same ID arrive. So the first piece is displayed before others arrive even though it should be hidden until the message is complete.

I accept your desire to have this stay backwards compatible by not changing the ```type``` so I would suggest the following incremental modification to be backwards compatible *and* versioned:
Add an integer ```fragments``` which contains the total number of 1024 byte pieces the message will consist of, and an integer ```fragment``` which identifies the current fragment. 
If those aren't present (or both equal to 1) in a received message then assume it's the old format of non-split messages.

Then putting the message together also becomes as easy as allocating a String array of the size being equal to ```fragments``` and putting the fragments into the index ```fragment```. Don't forget to add code to throw away all messages of the given ```msgid``` if malicious intent is indicated by e.g. an excessive ```fragments``` count, String size, multiple messages claiming a different ```fragments``` for a given ```msgid```, or multiple messages using the same ```fragment``` number. Fred probably even has code to disconnect/punish peers if they do such things.

xor-freenet: Ah sorry I didn't notice that ```msgid``` serves as an indicator for the new protocol :|
Still adding fragment numbers will be necessary to resolve the issue of being able to detect when a message is complete.

xor-freenet: Arguably the fact that I missed that ```msgid``` indicates the new protocol shows that it's a poor indicator of a message being split, it's not very intuitive. Normally either every object of a certain class has an identifier or the class doesn't have an identifier, selectively only adding an identifier to some is weird. An identifier doesn't sound very related to splitting, does it? :|

Having an ID in all, even non-split, messages will perhaps be useful for other things anyway such as sending read reciepts so using it for this meaning would block those things in the future.

xor-freenet:
@Bombe 
> This looks a little bit like a hot mess, especially the part where you rework a part over several commits because “eh, that’s not good.” I’d prefer having a cleaned up version of this.

Isn't one of the primary purposes of version control precisely to have a **log** of what happened to the code in the past so if I run into some weirdness in the future I can reconstruct why it ended up that way from reading the log?

It wouldn't be the first time where I had needed to look up the history of code in order to understand the intentions of the author (including myelf being the author :) ....  I'd rather have Arne not delete the Git history.

ArneBab:

> @Bombe 
>> This looks a little bit like a hot mess, especially the part where you rework a part over several commits because “eh, that’s not good.” I’d prefer having a cleaned up version of this.
>
> Isn't one of the primary purposes of version control precisely to have a **log** of what happened to the code in the past so if I run into some weirdness in the future I can reconstruct why it ended up that way from reading the log?

> It wouldn't be the first time where I had needed to look up the history of code in order to understand the intentions of the author (including myelf being the author :) ....  I'd rather have Arne not delete the Git history.

I honestly don’t mind here, because Bombe is right that the patches are
the result from three long evenings of work, and they do not form a very
coherent story, and because it’s git we’re using, where convention is
more polishing history than preserving a worklog to re-investigate (if
you want that, along with tools which make it easier, you should really
be using Mercurial, where safe history is part of the philosophy).

A clean set of commits would be:

- fix recommending files as n2n
- transparently dissasemble and reassemble large n2ns
- only show option to send files in advanced mode
- cleanup whitespace

But to do that, I would take the full diff and create those 4 commits
from it one by one. It would be great if there were a way to have hidden
commits in git which are referenced as source be polished commits and
can be investigated when needed.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

ArneBab:

> xor-freenet commented on this pull request.
>
>
>
>> -		N2NTMUserAlert userAlert = new N2NTMUserAlert(this, text, fileNumber, composedTime, sentTime, receivedTime);
> +		long msgid = fs.getLong("msgid", Long.MAX_VALUE);
> +		List<UserAlert> merged = new ArrayList<>();
> +		if (msgid != Long.MAX_VALUE) {
> +			// merge existing user alerts
> +
> +                        // TODO: This costs time linear to the number
> +			// of known alerts, so receiving a large text
> +			// message is quadratic in its size! Worst
> +			// case cost for transfer of a single huge
> +			// n2ntm (when receiving every second message)
> +			// is about 0.5 N*N, with N at most
> +			// 128. However even sending a 64kiB file
> +			// (DarknetPeerNode.java) does not cause
> +			// noticeable load on the receiving system.
> +                        synchronized (text) {
>
> Am I understanding the motivation of this code correctly like that:
> Because you didn't write code to intercept split messages to not display a UserAlert until the full message is available the split messages are displayed immediately as partial incomplete UserAlerts and pieced together as more messages arrive?

Firstoff: did you already test the code?

Second: My goal is to improve reliability of n2ns. I do not want to kill
the total message just because a sub-message gets lost, otherwise this
change would *reduce* reliablility by up to factor 128.

Any replacement algorithm should not require all messages to arrive. It
should always degrade gracefully in case of transmission problems.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

xor-freenet:
@ArneBab 
> and because it’s git we’re using, where convention is more polishing history than preserving a worklog to re-investigate 

Do you have a source for that? I've never hat history rewriting is the standard in Git.
I was instead of the opinon that history rewriting is strongly discouraged. 

The whole point of it being a chain of hashes is that you can't fake that.
The whole point of version control is storing versions across development, it's even in the name of the software concept!
Rewriting history deletes versions.

> (if you want that, along with tools which make it easier, you should really be using Mercurial, where safe history is part of the philosophy). 

Git will happily tell me when someone changed the history, I don't think your pet peeve of advertising for Mercurial has any chance to succeed here :) 

FWIW please don't feel offended as I was trying to not offend but defend you against Bombe's wording which I consider excessively strong for code review - and as said unnecessary:
Writing bad code is something which happens during development, and the effects of not fully noticing all pieces of it while fixing it means that its consequences can persist - which is why it is actually a good thing to preserve the bad code in the Git history so it can be used for forensics of future bug hunting.

xor-freenet:
@ArneBab

> Firstoff: did you already test the code?

No. Race conditions of the type "have a higher probability of occuring if things are slow", such as this one, do not have to be proven by testing if they are a consequence of network code, such as this one:
Networking is by definition often something which is slow, it will certainly happen that the necessary slowness to provoke the race condition does occur in real life.


> Second: My goal is to improve reliability of n2ns. I do not want to kill the total message just because a sub-message gets lost, otherwise this change would *reduce* reliablility by up to factor 128. Any replacement algorithm should not require all messages to arrive. It should always degrade gracefully in case of transmission problems.

From the personal impression of having known you for a few years I'd blindguess you're not the kind of person to enjoy using every one of the dozens of instant messengers which have been popular during the past decades.

So please let me tell you:
Of the popular ones I've used ICQ, Jabber (both with like 3 different clients), SMS, StudiVZ, Wer-kennt-wen, Facebook, WhatsApp, Signal, Instagram, and probably quite a few others others I forgot , and I cannot remember that any single of them will deliver parts of a multi-message while it is incomplete.

This will definitely confuse users, they aren't used to this at all.
We should neither impose the support effort of this upon ourselves nor impose the confusion upon the users.

> Any replacement algorithm should not require all messages to arrive. It should always degrade gracefully in case of transmission problems.

You are right there, I fully agree.
But the proper approach to packet loss is queueing and resending the messages until they can be delivered (i.e. you'd send an ACK once you got a message, and the sender will repeat them until he gets the ACK for all), as I said in the other comment.
On IRC you said you already implemented this - so I would suppose there is even less reason to keep this highly confusing code, is there?

xor-freenet:
With regards to how to implement a retransmission mechanism for packet loss:
I would be surprised if the N2N file transfer code doesn't have a retransmission mechanism given that files are large and packet loss thus is likely to happen during a file transfer.
It uses the classes freenet.io.xfer.BulkTransmitter and BulkReceiver.
If they do contain a retransmission mechanism you could change the text messaging code to send the messages like files internally.

However from starting at them for quite a while to me it looks like it's possible that they don't even do retransmitting on their own, it looks BulkTransmitter does the following instead:
- keep a map of packets it has send and sends packets until all have been sent.
- then waits for until the remote peer sends a single ACK for the whole file. If that times out the transfer is marked as failed. I.e. there is no ACK for each file chunk, but only a "I have received all chunks"-ACK.

To me this looks like fred's low level UDP code (i.e. what consumes the Message object created by DMT.createFNPBulkPacketSend() in BulkTransmitter) has a mechanism for trying a certain amount of time to retransmit individual packets until they have been ACKed, and the BulkTransmitter relies upon that so it doesn't need a packet-level ACK but only one for whether all chunks of the file have been received.
I don't have the time for digging any deeper into the UDP code, I would recommend you to do so before you write your own ACK/retransmit stuff.

ArneBab:

> With regards to how to implement a retransmission mechanism for packet loss:
…
> However from starting at them for quite a while to me it looks like it's possible that they don't even do retransmitting on their own, it looks BulkTransmitter does the following instead:
> - keep a map of packets it has send and sends packets until all have been sent.
> - then waits for until the remote peer sends a single ACK for the whole file. If that times out the transfer is marked as failed. I.e. there is no ACK for each file chunk, but only a "I have received all chunks"-ACK.

That’s what they seem to do from the UI, too. Which isn’t good usability.

> To me this looks like fred's low level UDP code (i.e. what consumes the Message object created by DMT.createFNPBulkPacketSend() in BulkTransmitter) has a mechanism for trying a certain amount of time to retransmit individual packets until they have been ACKed, and the BulkTransmitter relies upon that so it doesn't need a packet-level ACK but only one for whether all chunks of the file have been received.
> I don't have the time for digging any deeper into the UDP code, I would recommend you to do so before you write your own ACK/retransmit stuff.

I did: There can be an ack per packet. Also we already had a queue
mechanism. I use both to avoid having to implement a mechanism of my
own: n2ntm packets now get removed from the queue when they are acked.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

ArneBab:
>> Any replacement algorithm should not require all messages to arrive. It should always degrade gracefully in case of transmission problems.
>
> You are right there, I fully agree.
> But the proper approach to packet loss is queueing and resending the messages until they can be delivered (i.e. you'd send an ACK once you got a message, and the sender will repeat them until he gets the ACK for all), as I said in the other comment.

This already happens, but I do not trust it 100%. It’s much more
reliable than what we had before, but if we require all messages to
arrive before we show anything, we increase the reliability requirement by two
orders of magnitude -- which could easily void the improvements made.

This is why I don’t agree with your requirement: It moves the goalpost.

I have an idea how this could be made more efficient, but it carries the
risk of having to dive into debugging, so I want to have this solid
improvement landed before I start breaking it.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

ArneBab:

> @ArneBab 
>> and because it’s git we’re using, where convention is more polishing history than preserving a worklog to re-investigate 
>
> Do you have a source for that? I've never hat history rewriting is the standard in Git.

Yes:

git push --force

The default is to allow that to rewrite remote branches.

Aside from this: I’ve been in the VCS wars from the very beginning, and
this is a core difference in philosophy between Git and
Mercurial. You’ll have to go dig for sources yourself, though. I did not
keep "trophies" of everrything :-)

>> (if you want that, along with tools which make it easier, you should really be using Mercurial, where safe history is part of the philosophy). 
>
> Git will happily tell me when someone changed the history, I don't think your pet peeve of advertising for Mercurial has any chance to succeed here :) 

This is not the same as safe history. This is just enabling you to keep
history consistent.

Besides: My pet peeve is projects which chose git and later realize that
Mercurial would have been a better fit for their requirements. The
tragedy of that is that it’s typically too late to change.

> FWIW please don't feel offended as I was trying to not offend but defend you against Bombe's wording which I consider excessively strong for code review - and as said unnecessary:
> Writing bad code is something which happens during development, and the effects of not fully noticing all pieces of it while fixing it means that its consequences can persist - which is why it is actually a good thing to preserve the bad code in the Git history so it can be used for forensics of future bug hunting.

I know that, but since you already commented on things I changed in
later commits (if I did not misunderstand your comments), I think Bombe
was right in practice.

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

ArneBab:
> Do you have a source for that? I've never hat history rewriting is the standard in Git.

*old soldier mode*
I’ve been in the trenches, and now you ask me to prove that what ripped
off my arm was a hand grenade and not some fluffy bunny?
*/old soldier mode*

:-)

src/freenet/clients/http/QueueToadlet.java
```diff
@@ -875,24 +875,24 @@ public boolean run(ClientContext context) {
...
 				form.addChild("br");
 				form.addChild("textarea",
-				        new String[]{"id", "name", "row", "cols"},
-				        new String[]{"descB", "description", "3", "70"});
+					new String[]{"id", "name", "row", "cols"},
+					new String[]{"descB", "description", "3", "70"});
 				form.addChild("br");
 				if (core.node.isFProxyJavascriptEnabled()) {
-					form.addChild("script", new String[] {"type", "src"}, new String[] {"text/javascript",  "/static/js/checkall.js"});
+					form.addChild("script", new String[] {"type", "src"}, new String[] {"text/javascript",	"/static/js/checkall.js"});
```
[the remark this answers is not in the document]
ArneBab: ah, now I see what you mean -- yikes :)

src/freenet/clients/http/QueueToadlet.java
```diff
@@ -2149,13 +2149,13 @@ private HTMLNode createRequestTable(PageMaker pageMaker, ToadletContext ctx, Lis
 					headerRow.addChild("th", l10n("reason"));
 					break;
 				case LAST_ACTIVITY:
-					headerRow.addChild("th").addChild("a", "href", (isReversed ? "?sortBy=lastActivity" : "?sortBy=lastActivity&reversed"),  l10n("lastActivity"));
+					headerRow.addChild("th").addChild("a", "href", (isReversed ? "?sortBy=lastActivity" : "?sortBy=lastActivity&reversed"),	 l10n("lastActivity"));
```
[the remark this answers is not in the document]
ArneBab: tabify, makes all our inconsistent double-spaces stand out :-)

src/freenet/node/useralerts/N2NTMUserAlert.java
ArneBab: nope, using Math.abs and -1 would have been nicer. I changed that.

src/freenet/node/useralerts/N2NTMUserAlert.java
ArneBab: oh, yes, definitely

src/freenet/node/useralerts/N2NTMUserAlert.java
ArneBab: @Bombe argl, leftovers from removing the setters … thanks :)

@xor-freenet yes, they can all go.

src/freenet/l10n/freenet.l10n.de.properties
ArneBab: uh, yes … I was sloppy here.

src/freenet/node/Version.java
ArneBab: I moved it out now -- thank you again for the review! I created a cleaned up commit series at https://github.com/freenet/fred/pull/628

src/freenet/node/DarknetPeerNode.java
ArneBab: This is not "because I did not write" such code, but because this solution is robust against many kinds of failure. Even if we lose a message in between, the user gets two large messages with only the middle part missing.

The synchronization is already fixed (yepp, that was just wrong).

ArneBab:
Closed in favor of https://github.com/freenet/fred/pull/628 which rewrote commits for easier review.

xor-freenet:
@ArneBab 
> > @ArneBab 
> > > and because it’s git we’re using, where convention is more polishing history than preserving a worklog to re-investigate
> >
> > Do you have a source for that? I've never hat history rewriting is the standard in Git.
>
> Yes: git push --force The default is to allow that to rewrite remote branches. 

The default **of GitHub**, which isn't necessarily the default of Git :)
And it can be disabled in GitHub's preferences, which we should do if we haven't.

Even if ```--force``` didn't exist having a central repository always means that it's owner can change the history. He owns the hard disk on which the stuff resides and he can mess with the contents as he pleases. Hiding that from the UI by removing ```--force``` doesn't change that fact, does it?
Better have it be possible explicitly in the UI so people are aware that it can happen and are prepared to deal with rogue maintainers doing it by refusing to pull that into their repositories.
Git will notice them if a forced push happens, so they can refuse. In fact you have to do manual intervention to be able to pull a forced push into your clone. See [here](https://stackoverflow.com/questions/14787776/how-to-pull-after-a-forced-git-push).

Beyond that ```git push``` isn't even the default mode of operation of Git (and force isn't the default of git push!). Git is supposed to be used in a pull-fashion:
Remember how Linux kernel development works where Git was invented: Developers are supposed to send patches to Linus, and he merges them if he considers them as mergeworthy. Pushing doesn't happen there.
GitHub and pushing to it was probably merely invented to cope with people's desire for centralization, but GitHub isn't Git.

ArneBab:
Let’s not discuss git history mutation any further. We obviously have vastly different perspectives on this question and even approach it completely differently.

PR #628, opened 2018-03-16
Improve n2ntms

ArneBab:
I created some patches to fix n2n handling (recommending files, dis- and reassembling large n2ns and only showing direct file transfer in advanced mode, because its limitations have confused several of my new darknet contacts and we have working upload recommendation, which is a better model for sharing, see the commits for a rationale).

This is a reworked version of https://github.com/freenet/fred/pull/627


src/freenet/node/DarknetPeerNode.java
```diff
@@ -1393,12 +1391,25 @@ public int sendDownloadFeed(FreenetURI URI, String description) {
...
+			messagePart = message.substring(i * 1024,
+							Math.min((i+1) * 1024,
+								 message.length()));
+			SimpleFieldSet fs = new SimpleFieldSet(true);
+			fs.put("type", Node.N2N_TEXT_MESSAGE_TYPE_USERALERT);
+			fs.putSingle("text", Base64.encodeUTF8(messagePart));
+			fs.put("msgid", msgid);
+			fs.put("requiredParts", requiredN2nCount);
+			fs.put("partIndex", i);
+			// increment compose time to allow sorting messages
```
xor-freenet: Shouldn't sorting be done by the partIndex?

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1466,19 +1477,97 @@ public int sendFileOffer(HTTPUploadedFile file, String message) throws IOExcepti
...
+			// TODO: This costs time linear to the number
+			// of known alerts, so receiving a large text
+			// message is quadratic in its size! Worst
+			// case cost for transfer of a single huge
+			// n2ntm (when receiving every second message)
+			// is about 0.5 N*N, with N at most
+			// 128. However even sending a 64kiB file
+			// (DarknetPeerNode.java) does not cause
+			// noticeable load on the receiving system.
+			synchronized (node.clientCore.alerts) {
```
xor-freenet: Now that you have partIndex numbers this should use them for assembling the message by putting the parts into a String[] with partIndex being the offset, and then squashing the array into a single String.

xor-freenet:
Since Arne and I cannot reach consensus on the following matter which I vote to block this PR I would please ask for at least one other reviewer to cast a vote on this so we can resolve our conflict of opinion:

(@ArneBab  you only need to read the last paragraph, you know the rest)

This will split the messages into 1024 byte chunks **and** upon partial delivery at the recipient of only some or even one chunk the message chunks will be displayed to the user immediately, in their chopped up state.
Chunks will be inserted into the chopped up message while they arrive, but if they don't arrive soon enough the messed up state of the message will stay visible to the user.

I consider this highly confusing and prone to cause a lot of support requests:
Of the ~ 10 popular instant messengers and social networks I've used over the past years cannot remember that any single of them will display parts of a multi-message while it is incomplete.

As Arne has implemented queuing and ACKing of the messages already fixing this should be less than 10 lines of code: It's a matter of setting a boolean for the visibility of the UserAlert to false until the message is complete. IIRC such a boolean may exist already.

PR #629, opened 2018-03-17
Fix wrapper for new windows installs

TheSeeker:
Moving wrapper.conf into $\wrapper broke parsing/saving of the memory usage option.

If I did things right, it should check $\wrapper\wrapper.conf first, then fall back to $\wrapper.conf if not found.


ArneBab:
Looks good, I still need to test whether it actually does create no problems on GNU/Linux, though.

TheSeeker:
Incomplete; setWrapperProperty requires modification too.

PR #630, opened 2018-03-19
fix memory limit setting on windows v2

TheSeeker:
Previous patch fixed open/read, but write was still broken.

This patch includes write, and worksforme (actually run-tested it this time).   From what I'm seeing online, java has turned / in paths into whatever is the appropriate separator char for decades, and should pose no problems.  Hard coding \\ is what you want to avoid.

### DIFF
--- a/src/freenet/config/WrapperConfig.java
+++ b/src/freenet/config/WrapperConfig.java
@@ -40,10 +40,13 @@ public class WrapperConfig {
 			Logger.normal(WrapperConfig.class, "Cannot alter properties: not running under wrapper");
 			return false;
 		}
-		File f = new File("wrapper.conf");
+		File f = new File("wrapper/wrapper.conf");
 		if(!f.exists()) {
-			Logger.normal(WrapperConfig.class, "Cannot alter properties: wrapper.conf does not exist");
-			return false;
+			f = new File("wrapper.conf");
+			if(!f.exists()) {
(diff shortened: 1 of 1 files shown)


src/freenet/config/WrapperConfig.java
```diff
@@ -127,10 +136,10 @@ public static synchronized boolean setWrapperProperty(String name, String value)
 		}
 		
 		if(!newConfig.renameTo(oldConfig)) {
-			File oldOldConfig = new File("wrapper.conf.old");
+			File oldOldConfig = new File(wrapperDir + "/wrapper.conf.old");
 			if(oldOldConfig.exists() && !oldOldConfig.delete())
 				try {
-					oldOldConfig = File.createTempFile("wrapper.conf", ".old.tmp", new File("."));
+					oldOldConfig = File.createTempFile(wrapperDir + "wrapper.conf", ".old.tmp", new File("."));
```
Thynix: Thoughts on passing wrapperDir instead of "." for `directory` and not concatenating here?

TheSeeker: Because then I'd have to re-do the testing done above and duplicate even more code in order to support wrapper being possibly in two locations.

Of course I just now noticed I missed the / on line 142. should be   wrapperDir + "/wrapper.conf"  or  wrapperDir, "wrapper.conf"

PR #631, opened 2018-04-07
Update dependencies for 1480-pre1

Thynix:
We're going to merge `master` and `next`. `next` needs JNA, and there's been Bouncy Castle security updates.


nextgens:
Hmm,  no jna.order ... are you sure that it's optional?

PR #632, opened 2018-05-12
Rekey the update URIs to ensure they are RFC6979

Thynix:
This rekeys URIs for both EoL and newer versions of Java, which is not supported by existing code. This separates the legacy URI - for EoL JVMs going forward - from the "previous URI."

Once this is reviewed and ready to merge I can distribute the private keys before merging if that seems preferable.

I tested that the key config updated as expected, but not UOM. It behaves as expected when first running Java 7, subsequently running Java 8, and first running Java 8. (The script I used, which is specific to Arch, is [here](https://gist.github.com/Thynix/8d6f1ff2526dc83d25184d8dfb48bb6b).)

### DIFF
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -78,17 +78,24 @@ import freenet.support.io.FileUtil;
 public class NodeUpdateManager {
 
 	/**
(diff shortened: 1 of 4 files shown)


Thynix:
I've sent the shares and keys.

PR #633, opened 2018-06-16
Fix regression: add relative jna tmpdir path to wrapper if not existing

ArneBab:
This allows running multiple nodes on the same machine.

### DIFF
--- a/src/freenet/node/updater/UpdateDeployContext.java
+++ b/src/freenet/node/updater/UpdateDeployContext.java
@@ -256,7 +256,7 @@ public class UpdateDeployContext {
 		}
 		// ensure that we have an entry for the JNA tempdir
 		if (!writtenJnaTmpDir) {
-			bw.write("wrapper.java.additional."+count+"=-Djna.tmpdir=./tmp/"+'\n');
+			bw.write("wrapper.java.additional."+count+"=-Djava.io.tmpdir=./tmp/"+'\n');
 		}
 		
 		for(String s : otherLines)
--- a/src/freenet/node/updater/UpdateDeployContext.java
+++ b/src/freenet/node/updater/UpdateDeployContext.java
@@ -256,7 +256,7 @@ public class UpdateDeployContext {
 		}
 		// ensure that we have an entry for the JNA tempdir
 		if (!writtenJnaTmpDir) {
-			bw.write("wrapper.java.additional."+count+"=-Djava.io.tmpdir=./tmp/"+'\n');
+			bw.write("wrapper.java.additional."+count+"=-Djna.tmpdir=./tmp/"+'\n');
 		}
 		
 		for(String s : otherLines)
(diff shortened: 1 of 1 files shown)


nextgens:
Surely the right fix here is to configure it only when required AKA: not for the majority of users!

ArneBab:
@nextgens is there merit to using the global tempdir?

nextgens:
That's the expected behaviour

ArneBab:
Since it is completely unexpected that you cannot just run two instances of the same program without getting into trouble, that does not sound right. I don't think anyone among us expected that.

Or did you know that the switch to JNA would break nodes for everyone who runs more than one on the same system?

nextgens:
I didn't know that it prevents nodes from being run concurrently.... and I am still unclear as of why that might be.

Can you please describe what the actual problem is? and how do the symptoms manifest? Neither the commit message nor this PR help in this regard... and I am sure that the change you propose are far from being free of side effects.

ArneBab:
The actual effect is that JNA creates a folder in /tmp/ and uses it. A second instance of Freenet tries the same and fails because the folder is already in use.

Essentially this happens: https://stackoverflow.com/a/11461767/7666

And I do not consider chmod -R 777 /tmp/jna to be a good solution.

We might be able to replace java.io.tmp= by jna.tmpdir= to reduce the side-effects (see  https://docs.datastax.com/en/dse/5.1/dse-admin/datastax_enterprise/security/secTmp.html )

nextgens:
So if I read it right, it's only happening when you are trying to run concurrent nodes using different user IDs... and the root cause of the problem is that JNA uses mktemp() into a hardcoded folder name (instead of using mktempdir() first).

If setting jna.tmpdir is enough, I strongly suggest that we do that

nextgens:
I can't find any bug reported upstream related to this... have you reported one?

A quick glance at the code tells me that there are two code-path that are likely to be problematic:
https://github.com/java-native-access/jna/blob/d42375cc4569058de70db25fd6a802f11642c35f/src/com/sun/jna/Native.java#L1277
https://github.com/java-native-access/jna/blob/ea9de70cb73ca0794120666e8ed9794d88ea5c81/contrib/platform/src/com/sun/jna/platform/win32/COM/tlb/TlbImp.java#L138

I guess it's just a matter of modernizing the code to use java.nio.file.Files.createTempDirectory()

ArneBab:
Setting jna.tmpdir does not suffice.

here’s a bug report: https://freenet.mantishub.io/view.php?id=7020 — I reported this purely in IRC a few weeks ago because it came up just as I did the last checks before doing a test release I had hoped to be the final before the real release. We then found a solution, tested it, and I now implemented that.

nextgens:
I've reported the bug upstream... since it's where it needs to be fixed. I am still not happy with deploying the suggested workaround to all users.

https://github.com/java-native-access/jna/issues/985

ArneBab:
Thank you for reporting the bug upstream!

I’m also not happy that we have to roll out a workaround, but it’s better than shipping a version where we know that it will break the setup for some of our users.

The next step now is a test release on a separate update key, so they still have some time to fix this before we roll out the release to our users, and if they push a fix fast enough, we can easily undo this.

PR #635, opened 2018-06-30
commit COMPRESS-449 workaround

ArneBab:
for https://freenet.mantishub.io/view.php?id=6921

### DIFF
--- a/src/freenet/client/ArchiveManager.java
+++ b/src/freenet/client/ArchiveManager.java
@@ -4,7 +4,7 @@
 package freenet.client;
 
 import java.io.BufferedOutputStream;
-import java.io.DataOutputStream;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


src/freenet/client/ArchiveManager.java
```diff
@@ -350,8 +350,12 @@ public void run() {
 
 			if(ARCHIVE_TYPE.ZIP == archiveType)
 				handleZIPArchive(ctx, key, is, element, callback, gotElement, throwAtExit, context);
-			else if(ARCHIVE_TYPE.TAR == archiveType)
-				handleTARArchive(ctx, key, is, element, callback, gotElement, throwAtExit, context);
+			else if(ARCHIVE_TYPE.TAR == archiveType) {
+                                // COMPRESS-449 workaround, see https://freenet.mantishub.io/view.php?id=6921
+                                byte[] buf = new byte[(int) archiveSize];
+                                is.read(buf);
```
nextgens: This is the wrong way to go about fixing it.

1) there's no need to have a copy of the data
2) checking how many bytes were actually read is not optional !
3) this diverges from what upstream does for no good reason ! 

Wrap it in a SkipShieldingInputStream like they do or fix the root cause of the problem (find out what exception is thrown, by which type of InputStream and why).
https://github.com/apache/commons-compress/blob/master/src/main/java/org/apache/commons/compress/utils/SkipShieldingInputStream.java

I am worried that this might mask some more significant bug hidden somewhere on one of our support classes (encryption, ...)

nextgens:
LGTM; I'd still prefer if we had the associated unit test to go with it though

PR #638, opened 2018-09-18
Feature/remove htl 0

Bombe:
The success rates are always 0 so there’s no reason to show this line.


ArneBab:
Looks good. Thank you!

PR #640, opened 2018-10-09
[PATCH] Fix memory leak with BouncyCastleProvider

desyncr:
Fix issue with BounceCastleProvider introduced in abad64d133ed9a674d5f666f48db178a85652b9e causing node shutdown with OOM error.

It was opted to use a static initialization method to setup provider using the `static` initializer.

I've a node running on this branch and it seems to be OK following the reproduction steps provided in the mailing list.

### DIFF
--- a/src/freenet/crypt/KeyGenUtils.java
+++ b/src/freenet/crypt/KeyGenUtils.java
@@ -36,6 +36,10 @@ import freenet.support.Fields;
  */
 public final class KeyGenUtils {
 
+    static {
+        java.security.Security.insertProviderAt(new BouncyCastleProvider(), 1);
+    }
+
     /**
      * Generates a public/private key pair formatted for the algorithm specified
      * and stores the keys in a KeyPair. Can not handle DSA keys.
@@ -47,7 +51,7 @@ public final class KeyGenUtils {
             throw new UnsupportedTypeException(type);
         }
         try {
-            KeyPairGenerator kg = KeyPairGenerator.getInstance(type.alg, new BouncyCastleProvider());
+            KeyPairGenerator kg = KeyPairGenerator.getInstance(type.alg);
             kg.initialize(type.spec);
             return kg.generateKeyPair();
         } catch (NoSuchAlgorithmException e) {
@@ -68,7 +72,7 @@ public final class KeyGenUtils {
             throw new UnsupportedTypeException(type);
         }
         try {
-            KeyFactory kf = KeyFactory.getInstance(type.alg, new BouncyCastleProvider());
+            KeyFactory kf = KeyFactory.getInstance(type.alg);
             X509EncodedKeySpec xks = new X509EncodedKeySpec(pub);
             return kf.generatePublic(xks);
         } catch (NoSuchAlgorithmException e) {
@@ -123,7 +127,7 @@ public final class KeyGenUtils {
             throw new UnsupportedTypeException(type);
         }
         try {
-            KeyFactory kf = KeyFactory.getInstance(type.alg, new BouncyCastleProvider());
+            KeyFactory kf = KeyFactory.getInstance(type.alg);
 
             PublicKey pubK = getPublicKey(type, pub);
 
@@ -170,7 +174,7 @@ public final class KeyGenUtils {
      */
     public static SecretKey genSecretKey(KeyType type){
         try{
-            KeyGenerator kg = KeyGenerator.getInstance(type.alg, new BouncyCastleProvider());
+            KeyGenerator kg = KeyGenerator.getInstance(type.alg);
             kg.init(type.keySize);
             return kg.generateKey();
         } catch (NoSuchAlgorithmException e) {


src/freenet/crypt/KeyGenUtils.java
```diff
@@ -36,6 +36,10 @@
  */
 public final class KeyGenUtils {
 
+    static {
+        java.security.Security.insertProviderAt(new BouncyCastleProvider(), 1);
```
desyncr: Alternatively instead of modifying security we could save the instance in a property.

desyncr:
There is my setup to reproduce the issue: https://github.com/desyncr/freenet-oom-reproduction-setup

nextgens:
Can't we just revert that commit ? Why is it necessary in the first place ?

desyncr:
It fixes this issue: https://freenet.mantishub.io/view.php?id=7023

nextgens:
It's forcefully disabling other (potentially hardware-accelerated) providers... to make a bunch of unit tests pass on a JVM we've never officially supported ? Are we even using the missing ciphers ?

desyncr:
If I understand it correctly it doesn't disable other providers, but prioritizes `BouncyCastle`.
To remove the side effect it may be more conservative to instantiate a single `BouncyCastleProvider` in the static initialization and save it at the class level, passing it as argument as the current implementation.

What do you think of that?

nextgens:
I think that we need to establish whether that hack is required in the first place... and if it is, your patch is fine, provided we switch to Security.addProvider() to add it to the list without prioritizing it

desyncr:
I think @ArneBab found out the issue. I could set up an environment to reproduce it anyways.

> provided we switch to Security.addProvider() to add it to the list without prioritizing it

Would not cause the patch to actually not work for what it's indented to?

As I understand it needs to use BC provider rather than any other available, right? For this is I think passing the provider as an argument rather than registering this provider is the most conservative patch we could come up with.

I'll try out both methods, `addProvider` and an `static property` while I setup the environments.

nextgens:
@desyncr thank you for your work !

I don't think that this is what's going on... IIRC openjdk7 introduced ECC ciphers and then removed them by removing the SunEC provider... So we need at least "one provider" that supports ECC for the node to run (and the tests to pass) when that JVM is in use.

Now, fred uses freenet/crypt/JceLoader (loaded from freenet/node/Node) to ensure that BC is "somewhere to be found". Tests on the other hand do not (and explicitly load BC as required, using the static construct like your PR).

I am arguing that abad64d is most likely wrong: it's most likely unnecessary for "normal" operations of fred and patching the wrong codepath: if https://freenet.mantishub.io/view.php?id=7023 is the issue, the bugfix should have affected only the test (and not fred's code!). @ArneBab shouldn't have committed it in the first place.

@ArneBab can you confirm/infirm ?

desyncr:
I see. Very clear explanation.

> Now, fred uses freenet/crypt/JceLoader (loaded from freenet/node/Node) to ensure that BC is "somewhere to be found".

What would be the best way to make BC available on the tests?

nextgens:
My understanding is that it's already there:
https://github.com/freenet/fred/blob/next/test/freenet/crypt/KeyGenUtilsTest.java#L98

If that doesn't work, I'm keen to know why... it may need to be in setUp() like what we do in:
https://github.com/freenet/fred/blob/next/test/freenet/crypt/HMACTest.java#L31

desyncr:
Thank you! I'm looking into it!

ArneBab:
@nextgens can you file a pull request which fixes the tests to test our actual code path? If yes, I’m perfectly happy to revert both fixes.

Because the point is: The tests failed.

desyncr:
@ArneBab I'm working on fixing the tests for openjdk7: https://github.com/desyncr/fred/pull/1

nextgens:
@ArneBab abad64d (your code) introduces two regressions:
1) the memory leak
2) it disables other (potentially hw/accelerated) providers for *everyone*, including those that do not run a 8 year old JVM !

Neither are acceptable and what you've merged only addresses the first point

nextgens:
@ArneBab moreover, I am still not convinced that we do use ECC ciphers with the KeyGenUtils API... sure the tests fail... but are we using that code in fred at all ? I don't think so (at least not yet). The sane thing to do would be to revert the various "fixes" you have committed and avoid those tests until we make JDK8 mandatory.

Try "rgrep KeyGenUtils src/"

desyncr:
I think @nextgens is correct. It seems an issue with `openjdk`+`gradle` rather than `KeyGenUtils`. The issue was solved not because we force` BounceCastleProvider` but because it was installed manually and present for `gradle` to use.

In order to fix the failing tests we need to
- Apply the changes commented in https://github.com/travis-ci/travis-ci/issues/8503#issuecomment-339943812 (I'm testing this right now) -- only for openjdk7
- Remove the BounceCastleProvider fix

I'm working on it and will try to create a new PR today.

desyncr:
I was able to get  a clean build and tests on openjdk7 with java 7u181 (docker/gradle), travis runs on java 7u151 and seems to have issues. Not sure how to update to 7u181 on travis so I'm testing out with docker.

ArneBab:
We are still supporting Java 7 for this and the next build, afterwards we can drop this. Therefore I’d rather be safe.

However we can apply this fix on Java 7 only, since it is not a problem on Java 8.

desyncr:
@ArneBab I just finishing up testing travis.yml changes to support this: https://travis-ci.org/desyncr/fred

ArneBab:
I added a commit to make the fix conditional on the java version: https://github.com/freenet/fred/pull/642/commits/88a8d723719c4f21d41a47825c186c789b2688eb

PR #642, opened 2018-10-10
Reuse bc instance to avoid memory leak

ArneBab:


ArneBab:
I no longer see leaking memory with this change.

desyncr:
This looks good. I'm running on this branch and I see no issues so far.

desyncr:
@ArneBab Take into account it's a *development* issue and not an issue with Fred -- `gradle` issues with openjdk7 (missing SunEC) and java `7u151` AFAIK.

So it can be solved by working around issues with Travis-CI environment (see what I'm trying to do at https://github.com/freenet/fred/pull/644).

What do you think?

nextgens:
@ArneBab what's your rationale for re-inventing 
https://github.com/freenet/fred/blob/next/src/freenet/support/JVMVersion.java ?

Doing JVM-version parsing at runtime on a hot code-path ... and putting that logic under freenet/crypto is just insane.

Do the version parsing just once (use a static construct, set a final boolean and use that), possibly using the existing support classes.

This is all too complicated for fixing an issue that has yet to be proven to exist... Can you bootstrap a node with that broken JVM and see whether segment fetches/inserts (over 2MB compressed) work ? If they do (and I think they do work) this isn't an issue at all. The tests are exercising a cipher combination fred doesn't even use yet !

nextgens:
Oh and please use explicit commits messages and comments (possibly referencing bugs and or PRs) ... it's really hard to follow what you are doing.

Thynix:
I agree with @nextgens - what bug is this working around? Do we really have to work around it? Do we have a handle on the performance impact of querying the JVM version a lot? Thoughts on initializing a boolean once on startup instead?

Bigger picture: we no longer support Java 7 - I'd advocate either closing this PR, or, if we're feeling very kind to those on Java 7, merging this to a Java 7-only branch without the runtime check and include it in another release to the Java 7 update key.

ArneBab:
We can definitiely cut this now. I don’t see any reason to keep next java7 compatible now that 1484 is released.

ArneBab:
Thank you for your reviews!

PR #644, opened 2018-10-11
Support openjdk7 builds on Travis-ci

desyncr:
- Use latest java/jdk version (7u181) for gradle
- Add openjdk7 to jdk list, use matrix for faster builds (sudo: false vs sudo: required)
- Remove BouncyCastleProvider usage in KeyGenUtils
- Install and register BounceCastleProvider for openjdk7
- Use custom java.security file for openjdk7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,13 @@
 language: java
 
 sudo: false
+matrix:
+  include:
+    - jdk: oraclejdk9
+    - jdk: oraclejdk8
+    - jdk: openjdk8
+    - jdk: openjdk7
+      sudo: true
 
 env:
   global:
@@ -21,13 +28,18 @@ cache:
   - $HOME/.gradle/caches/
   - $HOME/.gradle/wrapper/
 
-before_install: echo -e "org.gradle.parallel = true\norg.gradle.jvmargs=-Xms256m -Xmx1024m\norg.gradle.configureondemand=true\ntasks.withType(Test) {\n maxParallelForks = Runtime.runtime.availableProcessors()\n}" > gradle.properties
+before_install:
+  - if [[ "${JAVA_HOME}" == *java-7-openjdk* ]]; then
+        sudo apt-get update && sudo apt-get install openjdk-7-jdk;
+        jdk_switcher use openjdk7;
+        export _JAVA_OPTIONS="-Xmx2048m -Xms512m -Djava.security.properties==${PWD}/java.security-openjdk-7";
+        sudo wget "https://bouncycastle.org/download/bcprov-ext-jdk15on-159.jar" -O "${JAVA_HOME}/jre/lib/ext/bcprov-ext-jdk15on-159.jar";
+        fi
+  - echo -e "org.gradle.parallel = true\norg.gradle.jvmargs=-Xms256m -Xmx1024m\norg.gradle.configureondemand=true\ntasks.withType(Test) {\n maxParallelForks = Runtime.runtime.availableProcessors()\n}" > gradle.properties
+
 install: "./gradlew jar"
 
-jdk:
-  - oraclejdk8
-  - oraclejdk9
-  - openjdk8
+script: "./gradlew check"
 
 before_deploy:
   - "./gradlew javadoc ; true"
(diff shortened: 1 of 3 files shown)


java.security-openjdk-7
```diff
@@ -0,0 +1,900 @@
+#
```
desyncr: This java.security is the default for required a custom order of `security.providers`, otherwise is pristine. I could just `perl/sed` the default file.

```
security.provider.1=sun.security.provider.Sun
security.provider.2=sun.security.rsa.SunRsaSign
security.provider.3=sun.security.ec.SunEC
security.provider.4=com.sun.net.ssl.internal.ssl.Provider
security.provider.5=com.sun.crypto.provider.SunJCE
security.provider.6=org.bouncycastle.jce.provider.BouncyCastleProvider
security.provider.7=sun.security.jgss.SunProvider
security.provider.8=com.sun.security.sasl.Provider
security.provider.9=org.jcp.xml.dsig.internal.dom.XMLDSigRI
security.provider.10=sun.security.smartcardio.SunPCSC
# the NSS security provider was not enabled for this build; it can be enabled
# if NSS (libnss3) is available on the machine. The nss.cfg file may need
# editing to reflect the location of the NSS installation.
security.provider.11=sun.security.pkcs11.SunPKCS11 
```

Thynix: Is there a reason it wouldn't work to specify `-Djava.security.properties=/path/to/overrides` instead of modifying the systemwide file?

Thynix: Oh, is it because it's a pain to control how Gradle calls Java?

desyncr: I didn't know there was such an option. I believe I just need to change:

```
org.gradle.jvmargs=-Xms256m -Xmx1024m
```

to

```
org.gradle.jvmargs=-Xms256m -Xmx1024m -Djava.security.properties=/path/to/overrides
```

Thanks for the heads up.

.travis.yml
```diff
@@ -21,13 +28,18 @@ cache:
   - $HOME/.gradle/caches/
   - $HOME/.gradle/wrapper/
 
-before_install: echo -e "org.gradle.parallel = true\norg.gradle.jvmargs=-Xms256m -Xmx1024m\norg.gradle.configureondemand=true\ntasks.withType(Test) {\n maxParallelForks = Runtime.runtime.availableProcessors()\n}" > gradle.properties
+before_install:
+  - if [[ "${JAVA_HOME}" == *java-7-openjdk* ]]; then
+        sudo apt-get update && sudo apt-get install openjdk-7-jdk;
+        jdk_switcher use openjdk7;
+        export _JAVA_OPTIONS="-Xmx2048m -Xms512m -Djava.security.properties==${PWD}/java.security-openjdk-7";
```
desyncr: I can't use `gradle.properties` to setup `_JAVA_OPTIONS`. I had to resort to overwriting this variable instead.

Thynix:
I'm sorry I didn't think of this before, but does it make sense to still support Java 7?

At least from looking at the updater it seems we're intending to require [Java 8](https://github.com/freenet/fred/blob/build01483/src/freenet/support/JVMVersion.java#L23) at minimum, and Debian Stretch has been stable for quite some time and has Java 8.

Bombe:
I am very much in favour of using Java 8 as minimum requirement.

Thynix:
I now see that in the [build 1484 pull request](https://github.com/freenet/fred/pull/649#issue-235195011) Arne says

> I hope we can also get it fixed that Freenet loses uploads on restart. Because after that we can switch to Java 8.

Bombe:
Wait, our moving to a newer Java version depends on some random bug (that probably nobody is actively working on)? Is there a reason for that?

ArneBab:
That bug is fixed in 1484.

The reason for delaying until after that fix is that this means that we
leave Java 7 people with a version of Freenet that was built from
Gradle. Without fixing that bug, many people would stick to 1480 (many
currently do).

David Roden <notifications@github.com> writes:

> Wait, our moving to a newer Java version depends on some random bug (that probably nobody is actively working on)? Is there a reason for that?


--
Unpolitisch sein
heißt politisch sein
ohne es zu merken

desyncr:
Knowing that its fixed already I'm closing this now.

PR #646, opened 2018-11-10
Hotfixes for 1481: Reduce peer count scaling to reduce CPU load during routing

ArneBab:
This is an emergency fix because some of our users with the biggest bandwidth report big problems with CPU load. They report that the update from 1480 to 1481 increased their CPU load by factor 2 to factor 3. Since in the merge from master to merge we kept the min bandwidth requirements of 10kB/s, the lowest peer count will be 4, which still suffices for scaling (as shown by the link length fix).

Risk: Users might complain that their target peers are reduced, because they don’t understand that this does not reduce the bandwidth they can use (because the bandwidth per peer is increased if the peer-count decreases network-wide, and caching should actually become more effective, countering potential problems due to longer paths).


ArneBab:
bff667b fixes the comment, thank you for your review!

PR #654, opened 2019-03-22
improvement in Winterfacey

Olezha:
What was changed:
1. Logo doesn't get out from menu block
2. In dropdown menu options take 1 line, not 2 lines. So dropdown menu extends to the length of an option
3. Blocks became of the same size within 1 line
4. Blocks are centered (before they were left-side)
5. Buttons are displayed on the right next to the fields

after_changes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ next ():
 - replace handler.outputHandler.queue by handler.send - thanks to patheticcockroach
 - add "send confidential message" button to friends page - thanks to Redwerk
 - update plugin Freemail_wot to v0.2.7.4 with better detection of contacts missing from WoT - thanks to Redwerk
+- improvement in Winterfacey
 
 1483:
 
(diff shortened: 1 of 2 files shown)


ArneBab:
How does this cope with blocks that have more than 2 lines of text?

ArneBab:
How does it look with a notification that a new version is available?

Olezha:
> How does this cope with blocks that have more than 2 lines of text?

The entire line will be the height of the largest element.

> How does it look with a notification that a new version is available?

Can you tell me how to enable this notification for the test?

ArneBab:
> Can you tell me how to enable this notification for the test?

You can install version 1483 and then add the theme as CSS override on http://localhost:8888/config/fproxy?fproxyAdvancedMode=2

ArneBab:
> The entire line will be the height of the largest element.

Sounds good.

For the logo I personally like seeing it jump from white into dark, but I’m not a professional designer so if you have the note from a designer, I’d merge that.

Olezha:
It is a notification and I also temporalily added text into the Enzo's Index block to check how it looks like
Screenshot from 2019-05-21 13-43-31

ArneBab:
Looks good.

PR #659, opened 2019-05-05
Improve RSS/feed sniffing (supersedes #599)

Bombe:
This more resembles the sniffing done by Firefox and should lead to fewer false positives.

This pull request is intended as a cleaned up version of the pull request that was intended as a more readable alternative to #598 and included unit tests. It also has `next` merged into it as it’s been lying around for two and a half years.


ArneBab:
Looks good!

PR #660, opened 2019-05-10
Testing-refactoring

mbyrenh:
The changes in this request aim to make it easier to test Freenet's different parts in isolation.

Please see [my email](https://www.mail-archive.com/devl@freenetproject.org/msg55130.html) on the Freenet dev list for details.


src/freenet/node/FailureTable.java
```diff
@@ -287,8 +287,8 @@ public boolean isExpired() {
 	 */
 	public void onFound(KeyBlock block) {
 		if(logMINOR) Logger.minor(this, "Found "+block.getKey());
-		if(!(node.enableULPRDataPropagation || node.enablePerNodeFailureTables)) {
-			if(logMINOR) Logger.minor(this, "Ignoring onFound because enable ULPR = "+node.enableULPRDataPropagation+" and enable failure tables = "+node.enablePerNodeFailureTables);
+		if(!(node.enableULPRDataPropagation() || node.enablePerNodeFailureTables())) {
```
Olezha: Probably there should be node.isULPRDataPropagationEnabled()

Bombe:
Yay, refactoring!

Just one general remark: please don’t create separate interfaces for package-private stuff. Package-private access was a bad idea to start with. Make it public or protected, please. :)

Bombe:
Oh, and another thing (from your aforementioned email): calling methods in subclasses before the constructor has finished is risky. For the case in your email: don’t construct `MyClass` in the constructor, _especially_ when all you need for it is the given `param`, just make it a constructor parameter and create it on the outside, i.e. in your test.

src/freenet/io/comm/UdpSocketHandlerImpl.java
```diff
@@ -0,0 +1,472 @@
...
+import freenet.io.AddressTracker;
+import freenet.io.comm.Peer.LocalAddressException;
+import freenet.node.Node;
+import freenet.node.NodeImpl;
+import freenet.node.PrioRunnable;
+import freenet.support.Logger;
+import freenet.support.io.NativeThread;
+import freenet.support.transport.ip.IPUtil;
+
+public class UdpSocketHandlerImpl implements PrioRunnable, PortForwardSensitiveSocketHandler, UdpSocketHandler {
```
Olezha: On the renamed classes, I do not see the difference. This happened because in one commit there is an interface with the old class name. It's hard to see what the changes were there.

test/freenet/node/PeerNodeTest.java
```diff
@@ -0,0 +1,191 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
```
Olezha: Hmm

src/freenet/node/FailureTable.java
mbyrenh: That method name would also be fine for me. Shall I rename it and add a corresponding commit?

src/freenet/io/comm/UdpSocketHandlerImpl.java
mbyrenh: I agree that this is problematic. I probably didn't use git correctly when renaming the files. Currently, I see two ways to handle that

(1) You can use a diff tool like meld to compare the *Impl-File with the original file from the next branch. It's not a nice way but since it is "only" seven classes that were renamed, it might be doable. I will avoid such problems in any following commits.

(2) I can try to replace the problematic commits with proper commits. However, I don't know how much effort this would actually require. I would probably create a new branch and cherry-pick the non-problematic commits while replacing the problematic ones.

Which solution would you prefer?  Of course, I'm also open for other solutions.

test/freenet/node/PeerNodeTest.java
mbyrenh: Argh, sorry. I will add a commit that removes the template text.

src/freenet/io/comm/UdpSocketHandlerImpl.java
Olezha: I am a new user and therefore can not recommend something. I can only say that I read PR excluding such files and did not meet anything bad. I also liked the addition of the jacoco tool.

mbyrenh: Okay, no problem. Thank you for the feedback!

I just looked a bit into solution (2) and my current understanding is that Git's renaming detection does not work well with the steps I performed (renaming as well as modifying the class' java-file and additionally add a new file with the old name that now contains the interface).
As an alternative, instead of renaming each class to "*Impl", I could keep the class names and create an interface "IMyClass" for each class "MyClass". This way, I don't move existing code around, which should make commit diffs cleaner. Furthermore, having an interface called "IMyClass" for each class might generally be a nicer approach than having an "*Impl"-class for each interface.

Olezha: Personally, I prefer the `Impl` option. And on the wiki, I saw a link to the Google code convention as adopted here ..
If you rename the classes and make a commit and then make the interfaces as a separate commit, I think diffs will be visible more cleaner.

Bombe: I think both `IMyClass` and `MyClassImpl` suck.

I prefer an interface with a clean name (_especially_ no Hungarian-type prefixes) and a name for the implementation that alludes to what sets this implementation apart from all the other implementations. I am aware that some (or maybe most?) of the implementations here will probably have no more defining aspect than being the only implementation of the interface in questions but even a `DefaultMyClass` (`MyClass` is a horrible name to illustrate this 😃) is better than a `MyClassImpl`.

Names have meaning! Please don’t waste this opportunity by using some fixed scheme that a machine could apply. :(

Then of course there’s also the question whether we actually need all this if we only want to write tests. Mockito can mock classes, too, and as long as there are accessor methods for things, mocking those is not an issue, either.

mbyrenh: Besides testing, my intention of creating these interfaces was also to make it easier to replace different parts with other implementations for performance comparison (e.g. to compare different load balancing algorithms or routing algorithms) or maybe even enable re-using most of Freenet's code in a discrete, event-based simulation model). I chose to start with writing tests in order to familiarize myself better with the inner workings of Freenet and the dependencies among the different classes.

mbyrenh: When I created this pull request, I was not aware of the pull requests created by Toad, which also aim to make it easier to test and perform simulations of Freenet. I guess it does not make sense to further discuss my changes as long as Toad's pull requests have not been dealt with. I am willing to help you with reviewing Toad's changes. However, I consider it unrealistic that a single person can review all the changes of [pull request 576](https://github.com/freenet/fred/pull/576/) anytime soon. If you would be willing to participate in the review, we can split the work. Instead of everyone reviewing single commits, I consider it more realistic that we review the [accumulated changes of a each file](https://github.com/freenet/fred/pull/576/files) and resolve conflicts, if existing.

Olezha: I would like to clarify regarding workflow with PR of not my origin (how to commit resolved conflicts).

Bombe: @mbyrenh unless all logical changesets are always restricted to a single file (possible but unlikely, I guess), that way lies madness.

mbyrenh: @Bombe Why exactly do you consider this approach problematic? My main concern with reviewing commit-by-commit is that I may spend a lot of time with reviewing and discussing a commit only to discover later on that there is another commit in the PR that changes the code again and renders the previous review and discussion obsolete.

Bombe: @mbyrenh because the unit “commit” has usually been chosen for a very good reason. You don’t want to split a commit into two parts because it touches two files, and you don’t want to lump 2 commits together because they touch the same file. Files and commits are two different axises, and this is valid for both writing and reviewing.

And even if a later commit partially invalidates an earlier one, the history is still important in order to fully understand why the changes were made the way they were. I would never give up that knowledge. 

If you are worried about later commits obsoleting earlier ones, get the commits in order before reviewing, i.e. remove/combine reverts and commits that are like 15 tries of something in a row until it worked.

mbyrenh:
@Bombe I get your point but I was also hoping to find a way to split that huge pull request into smaller, hopefully mostly independent bits that can be worked on in parallel by multiple persons, so that there is at least a chance that it will be reviewed and merged in the foreseeable future. With your comments, I currently only see the option that a single person goes through commit by commit and so far, it does not look as if there is anyone that has enough spare time to work through the full request alone. If you have a better idea, I am open for suggestions. Otherwise, it might even be worth considering to set a deadline such that if no one comes forward until then, the pull request will be closed without merging so that it does not have to be discussed again and again whenever someone wants to do something similar.

Bombe:
@mbyrenh of course, smaller chunks are better to review but splitting by file is not the way to go in these situations – unless there are no changes in file A which depend on changes in file B.

We could split this up by tackling one field/method at a time so that all changes related to that field are in a single commit. This would make the changes less inconvenient to review; it would require someone to cut new commits from this PR. :/

(And that’s completely ignoring the fact that _all of this_ is public interface and changing it can break any number of plugins… but for fixing that we’d need a new plugin API.)

mbyrenh:
Since the discussion is not related to this pull request anymore and I guess that it is not going to be merged, I am closing this pull request now.

PR #661, opened 2019-05-14
WIP m3u filter hardcoded hostport

ArneBab:
This makes m3us playlists (aka streaming over Freenet) work for mp3, flac, ogg vorbis and theora, if Freenet uses the default host+port, but uses a hardcoded host+port combination.

To actually work for all setups it would have to include the actual host+port in the filtered m3u, but that would leak information which could be used to find out whether someone runs a node.

For reference why this would be dangerous, here’s the guide to create the most secure setup for Freenet: https://d6.gnutella2.info/freenet/USK@sUm3oJISSEU4pl2Is9qa1eRoCLyz6r2LPkEqlXc3~oc,yBEbf-IJrcB8Pe~gAd53DEEHgbugUkFSHtzzLqnYlbs,AQACAAE/random_babcom/396/#UseFreenetasproxysecureagainstspyingattacks

### DIFF
--- a/src/freenet/client/filter/GenericReadFilterCallback.java
+++ b/src/freenet/client/filter/GenericReadFilterCallback.java
@@ -192,7 +192,7 @@ public class GenericReadFilterCallback implements FilterCallback, URIProcessor {
 		// Convert localhost uri's to relative internal ones.
 		
 		String host = uri.getHost();
-		if(host != null && (host.equals("localhost") || host.equals("127.0.0.1")) && uri.getPort() == 8888) {
+		if(host != null && (host.equals("localhost") || host.startsWith("127."))) {
 			try {
 				uri = new URI(null, null, null, -1, uri.getPath(), uri.getQuery(), uri.getFragment());
 			} catch (URISyntaxException e) {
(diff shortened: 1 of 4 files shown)


src/freenet/client/filter/HTMLFilter.java
```diff
@@ -2814,14 +2814,23 @@ private boolean isStandardLinkType(String token) {
 			String hreflang = getHashString(h, "hreflang");
 			String charset = null;
 			String maybecharset = null;
-            /* TODO: get the type from the filename. Currently we only
-             * have a filter for mp3, so this is the simplest possible
-             * solution.*/
-            String type = "audio/mpeg";
 
 			String src = getHashString(h, "src");
```
Bombe: Please extract this whole block into a method, I’m sure the containing method is large enough already… :)

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -192,7 +192,7 @@ public String processURI(String u, String overrideType, boolean forBaseHref, boo
 		// Convert localhost uri's to relative internal ones.
 		
 		String host = uri.getHost();
-		if(host != null && (host.equals("localhost") || host.equals("127.0.0.1")) && uri.getPort() == 8888) {
+		if(host != null && (host.equals("localhost") || host.startsWith("127."))) {
```
Bombe, line 195: This will also match `127.freenetproject.org`. 😃

PR #662, opened 2019-05-16
File upload

Olezha:
- upload (large file) is getting broken down after the restart. I found the limit and increased it
- File compression (upload preparation step)
    - work a little on compression and stop to measure the effect of compression and continue compressing only if compression makes sense
    - If compression took a lot of time, do not try other compression algorithms
    - these parameters are moved to Freenet Core Settings

There was a discussion on the FMS: Freenet > Upload file 15 Gigabytes


src/freenet/support/Fields.java
```diff
@@ -357,6 +358,62 @@ public static String secToDateTime(long time) {
 		return dateString;
 	}
 
+	/**
+	 * Accepted format is {integer}{d|h|min|s}
+	 */
+	public static int durationToMillis(String duration) {
```
Olezha: You can skip reading (for saving time) these 2 methods as they were transferred and rewritten in the next commit.

src/freenet/support/compress/Bzip2Compressor.java
```diff
@@ -78,6 +73,9 @@ public long compress(InputStream is, OutputStream os, long maxReadLength, long m
 				read += x;
 				if(cos.written() > maxWriteLength)
 					throw new CompressionOutputSizeException();
+
+				if (++i == iterationToCheckCompressionRatio)
+					checkCompressionEffect(read, cos.written(), minimumCompressionPercentage);
```
ArneBab: please add curly braces around the if body.

test/freenet/support/FieldsDurationTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Tests parsing of duration value.
+ */
+public class FieldsDurationTest extends TestCase {
+
+    /**
+     * Duration input with and without various d|h|min|s.
+     */
+    private static final String[] durations = { "2d", "3h", "20m", "56s", "1h30m" };
```
ArneBab: You can use a parametrized test to make this show all failing tests, and to make it trivial to add more tests. https://github.com/junit-team/junit4/wiki/Parameterized-tests

Bombe: Please do not write parameterized tests.

- Parameterized tests can’t tell you any more than that two values didn’t match but they offer no context. Tests are supposed to be documentation, by means of the name of the test method and the used variable names.
- It’s easy to mix up values in long lists and so a broken test or broken code can stay undetected for longer. That’s easily avoidable by creating a method for each value pair.
- They encourage people to add more tests that add no value but increase the number of tests.

ArneBab: For a parser like this, I disagree.

If you use the parametrized tests runner, then you always have pairs of {input, output}, ideally each on its own line, and you can annotate the variables to give good output. For a test with lots of edge-case-values like a parser, that’s a really efficient way to test, and clear to read.

For a parser, parametrized tests are documentation by example. Due to the conciseness of one input-output pair per line, it’s easier to spot mistakes, because the values are not buried in several lines of repetitive ceremony.

And having the test-values in one block makes it easier to spot unnecessary duplication of tested logic, because you can grasp all the test values with one glance.

To show that by way of example:

```java
    @Parameters(name = "{index}: duration string: {0}, millis: {1}")
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][] {
                 // single units
                 { "2d", 2 * 24 * 3600 * 1000},
                 { "3h", 3 * 3600 * 1000},
                 { "20m", 20 * 60 * 1000},
                 { "56s", 56 * 1000},
                 // mixed units
                 { "1h30m", 5400 * 1000}
           });
    }
```

PR #663, opened 2019-05-22
peers-offers/*.fref files

Olezha:
If there are *.fref files in a peers-offers/ folder, ask user whether to connect to them. That would be a step towards introduction bundles: You can manually build an installer which auto-connects to you.
https://www.mail-archive.com/devl@freenetproject.org/msg55104.html

Implementation question (QA): In a case when 2 or more .fref files are added to peers-offers folder - should a user get as many message windows as there are peers? This will allow setting level of trust and ability to see and connect to this friend for each peer separately.


src/freenet/clients/http/ConnectionsToadlet.java
```diff
@@ -641,6 +637,30 @@ public void handleMethodPOST(URI uri, final HTTPRequest request, ToadletContext
...
+			if (Boolean.parseBoolean(request.getPartAsStringFailsafe("peers-offers-files", 5))) {
+				File[] files = core.node.runDir().file("peers-offers").listFiles();
+				if (files != null && files.length > 0) {
+					StringBuilder peersOffersFilesContent = new StringBuilder();
+					for (final File file : files) {
+						if (file.isFile()) {
+							String filename = file.getName();
+							if (filename.contains(".") && "fref".equals(filename.substring(filename.lastIndexOf(".") + 1))) {
+								try (Scanner scanner = new Scanner(file)) {
+									peersOffersFilesContent.append(scanner.useDelimiter("\\A").next());
```
Bombe: `\A` is the boundary match for “beginning of the input,” what is that supposed to accomplish here?

src/freenet/clients/http/ConnectionsToadlet.java
```diff
@@ -641,6 +637,30 @@ public void handleMethodPOST(URI uri, final HTTPRequest request, ToadletContext
...
+
+			if (Boolean.parseBoolean(request.getPartAsStringFailsafe("peers-offers-files", 5))) {
+				File[] files = core.node.runDir().file("peers-offers").listFiles();
+				if (files != null && files.length > 0) {
+					StringBuilder peersOffersFilesContent = new StringBuilder();
+					for (final File file : files) {
+						if (file.isFile()) {
+							String filename = file.getName();
+							if (filename.contains(".") && "fref".equals(filename.substring(filename.lastIndexOf(".") + 1))) {
+								try (Scanner scanner = new Scanner(file)) {
```
Bombe: This creates a `Scanner` using the platform’s default charset. Please specify a charset or don’t use `Scanner`, it’s a rather horrible API, anyway.

src/freenet/clients/http/ConnectionsToadlet.java
```diff
@@ -641,6 +637,30 @@ public void handleMethodPOST(URI uri, final HTTPRequest request, ToadletContext
...
+							}
+						}
+					}
+					reftext = peersOffersFilesContent.toString();
+				}
+
+				try {
+					node.config.get("node").set("peersOffersDismissed", true);
+				} catch (InvalidConfigValueException | NodeNeedRestartException e) {
+					e.printStackTrace();
```
Bombe: Nope, either handle the exception or re-throw it.

src/freenet/node/Node.java
```diff
@@ -756,6 +746,8 @@ public void set(String val) throws InvalidConfigValueException {
 	
 	private boolean enableRoutedPing;
 
+	private boolean peersOffersDismissed;
```
Bombe: Is there a way to keep this functionality out of `Node`? It’s huge enough as it is…

src/freenet/node/Node.java
```diff
@@ -2539,6 +2533,46 @@ public void realRun() {
...
 	}
 
+	// peers-offers/*.fref files
+	private void processPeersOffersFrefFiles(SubConfig nodeConfig, int configOptionSortOrder) {
+		nodeConfig.register("peersOffersDismissed", false, configOptionSortOrder, true, true,
+				"Node.peersOffersDismissed", "Node.peersOffersDismissedLong", new BooleanCallback() {
+
+					@Override
+					public Boolean get() {
+						synchronized(Node.this) {
```
Bombe: Please don’t synchronize on the `Node` itself, this would allow any plugin to block access to this.

src/freenet/node/Node.java
```diff
@@ -2539,6 +2533,46 @@ public void realRun() {
 		System.out.println("Node constructor completed");
 	}
 
+	// peers-offers/*.fref files
+	private void processPeersOffersFrefFiles(SubConfig nodeConfig, int configOptionSortOrder) {
```
Bombe: This method both handles the configuration and processes the files but its name does not reflect that. Either change the method or change the name. :)

src/freenet/node/Node.java
```diff
@@ -2539,6 +2533,46 @@ public void realRun() {
...
+		if (peersOffersDismissed) return;
+
+		File[] files = runDir.file("peers-offers").listFiles();
+		if (files != null && files.length > 0) {
+			StringBuilder frefFiles = new StringBuilder();
+			String prefix = "";
+			for (final File file : files) {
+				if (file.isFile()) {
+					String filename = file.getName();
+					if (filename.contains(".") && "fref".equals(filename.substring(filename.lastIndexOf(".") + 1))) {
```
Bombe: Uh… `filename.endsWith(".fref")`?

src/freenet/node/Node.java
```diff
@@ -2539,6 +2533,46 @@ public void realRun() {
...
+						frefFiles.append(prefix).append(file.getName());
+						prefix = ", ";
+					}
+				}
+			}
+			this.clientCore.alerts.register(new PeersOffersUserAlert(frefFiles.toString(), this));
```
Bombe: Also, I would totally move the generation of this `String` into the `UserAlert` and just hand in the relevant files to the `UserAlert`. (Probably using a fake-factory for the alert, i.e. a static `createAlert(node, files)` method.)

src/freenet/node/useralerts/PeersOffersUserAlert.java
```diff
@@ -0,0 +1,85 @@
...
+                        new String[] {"type", "name", "value"},
+                        new String[] {"submit", "add", "Connect"});
+
+        return content;
+    }
+
+    @Override
+    public boolean isValid() {
+        if (node.config.get("node").getBoolean("peersOffersDismissed"))
+            valid = false;
```
Bombe: This should be set in `onDismiss`, should it not? It looks like it should.

src/freenet/node/useralerts/PeersOffersUserAlert.java
```diff
@@ -0,0 +1,85 @@
...
+        return valid;
+    }
+
+    @Override
+    public String dismissButtonText() {
+        return "No, thanks";
```
Bombe: This would like to be translated, too. :)

src/freenet/node/useralerts/PeersOffersUserAlert.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public void onDismiss() {
+        try {
+            node.config.get("node").set("peersOffersDismissed", true);
+        } catch (InvalidConfigValueException | NodeNeedRestartException e) {
+            e.printStackTrace();
```
Bombe: Again, please no `printStackTrace()`.

src/freenet/node/useralerts/PeersOffersUserAlert.java
```diff
@@ -0,0 +1,85 @@
...
+                new String[] {"action", "method"},
+                new String[] {"/friends/", "post"});
+        form.addChild("input",
+                new String[] {"type", "name", "value"},
+                new String[] {"hidden", "formPassword", node.clientCore.formPassword});
+        form.addChild("input",
+                new String[] {"type", "name", "value"},
+                new String[] {"hidden", "peers-offers-files", "true"});
+
+        ConnectionsToadlet.drawPeerTrustInputForAddPeerBox(form);
```
Bombe: Looks like this method should not live in that particular toadlet if it’s used from here as well.

src/freenet/node/Node.java
Olezha: One option is to create a `configuration` package and create classes there that will contain groups of them. It makes no sense to move something separately, but if there is a lot ..

src/freenet/node/useralerts/PeersOffersUserAlert.java
Olezha: I processed all the comments except for the issue of moving the `nodeConfig` functionality from the `Node` class. Thank you for correcting me!

PR #664, opened 2019-05-31
clean js

Olezha:
This is on page 'Connections to friends'
Page functions after refactoring as before, in particular in 'My Friends (trusted peers added by me)' table > 'Private Note' fields (when a user puts/changes input in 'Private Note' field and loses focus, page is refreshed and input is saved)


ArneBab:
looks good.

Bombe:
Well, looks horrible but it did that before, too, and it’s actually slightly less horrible now. :)

PR #665, opened 2019-05-31
/test page

Olezha:
This page was investigated on its functions and problems that it has.
It was discovered that it contains 3 tests (MIME Inline, Number of connections, Javascript)
Each of these tests had bugs that were fixed.
1) MIME Inline
Before fix > was invalid because contained not needed characters (in particular, ====)
After fix > become valid and is shown on other websites where inline format is allowed. It is correct that on Freenet it is not displayed as inline format is not allowed on Freenet.
2) Number of connections
Before fix > last picture which alt = 'fail' was in a wrong format (in particular, gif) that is why was not displayed
After fix > last picture which alt = 'fail' is in a correct format (in particular, png) that is why is displayed
3) Javascript
Before fix > picture in JavaScript block was in a wrong format (in particular, gif) that is why was not displayed
After fix > picture in JavaScript block is in a correct format (in particular, png) that is why is displayed

To sum up, before fix all 3 tests gave to user wrong results. With fixes page gives correct results.
Seems like it is necessary to add the fixes or to remove the page. Without fixes it gives completely wrong test results.

results


ArneBab:
looks good

src/freenet/clients/http/BrowserTestToadlet.java
```diff
@@ -175,43 +175,50 @@
...
+
 		// #### Test MIME inline
+		/* for test (for allow <img src="data:...) add "; img-src 'self' data:"
+		 * to freenet.clients.http.ToadletContextImpl#generateCSP return statement */
 		ctx.getPageMaker().getInfobox("infobox-warning", "MIME Inline", contentNode, "mime-inline-test", true).
-			//addChild("img", new String[]{"src", "alt"}, new String[]{"data:image/gif;base64,"+imgWarningMime, "Your browser is probably safe."});
-			addChild("img", new String[]{"src", "alt"}, new String[]{"?mimeTest", "Your browser is probably safe."});
-		
+			addChild("img", new String[]{"src", "alt"}, new String[]{"data:image/gif;base64,"+imgWarningMime, "Your browser is probably safe."});
+//			addChild("img", new String[]{"src", "alt"}, new String[]{"?mimeTest", "Your browser is probably safe."});
```
Bombe: Please don’t add commented code.

PR #666, opened 2019-06-13
Changes are not saved in 'Web interface port' in Configurations

devalexk:
Changes are not saved in 'Web interface port' in Configurations was fixed.


Bombe:
That is insufficient. You would also need to close the current server socket and re-open it on the new port which would instantly break the web interface for the user.

With a lot of force this could be made to work but I don’t think this is worth it.

PR #668, opened 2019-07-02
Additional checks in Theora Filter

Olezha:


src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+		if (ISLEAF == 1) {
+			if (HTS.length == 32)
+				throw new UnknownContentTypeException("HTS[hti] = " + Arrays.toString(HTS) +
+						"; HTS[hti] is already 32 - The stream is undecodable.");
+			int TOKEN = input.readInt(5);
+
+			HTS = Arrays.copyOf(HTS, HTS.length + 1);
+			HTS[HTS.length - 1] = TOKEN;
+		} else {
+			HBITS += 0;
```
Bombe: This is the same as `HBITS += "0"` but more confusing. (Also, still not a big fan of modifying method parameters.)

Bombe: On second thought, does this have to be a String at all? It looks like you’re just using it for counting, the actual value doesn’t seem to matter, except for logging…

src/freenet/client/filter/TheoraBitstreamFilter.java
```diff
@@ -1,13 +1,14 @@
...
+import freenet.support.Logger;
+
 import java.io.IOException;
 import java.util.ArrayList;
-
-import freenet.support.Logger;
-import freenet.support.Logger.LogLevel;
+import java.util.List;
 
 public class TheoraBitstreamFilter extends OggBitstreamFilter {
```
Bombe: I would also totally love some tests for this class.

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+
 					input.readFully(magicHeader);
-					Logger.minor(this, "Header type: "+magicHeader[0]);
-					if(magicHeader[0] != -127) throw new DataFilterException("Header type: " + magicHeader[0]);
-					try {
-						checkMagicHeader(magicHeader);
-					} catch (UnknownContentTypeException e) {
-						throw new DataFilterException(e.getType());
-					}
+					checkMagicHeader(magicHeader, (byte) 0x81); // -127
```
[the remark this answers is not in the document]
Olezha: spec say `0x81` - end of page 47 or https://git.io/fj6qg

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+					if (FMBH == 0) throw new UnknownContentTypeException("Header FMBH: " + FMBH);
+
+					int PICW = input.readInt(24);
+					if (PICW > FMBW*16) throw new UnknownContentTypeException("Header PICW: " + PICW + "; FMBW: " + FMBW);
+
+					int PICH = input.readInt(24);
+					if (PICH > FMBH*16) throw new UnknownContentTypeException("Header PICH: " + PICH + "; FMBH: " + FMBH);
+
+					int PICX = input.readInt(8);
+					if (PICX > FMBW*16-PICX) throw new UnknownContentTypeException("Header PICX: " + PICX + "; FMBW: " + FMBW + "; PICX: " + PICX);
```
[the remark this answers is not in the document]
Olezha: This is from page 43 point 9 or https://git.io/fj6qt

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+					int NOMBR = input.readInt(24);
+					int QUAL = input.readInt(6);
+					int KFGSHIFT = input.readInt(5);
+
+					int PF = input.readInt(2);
+					if (PF == 1) throw new UnknownContentTypeException("Header PF: " + PF);
+
+					int Res = input.readInt(3);
+					if (Res != 0) throw new UnknownContentTypeException("Header Res: " + Res);
+
```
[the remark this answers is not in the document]
Olezha: Now I skip ignored values. Should I think about some other streamlining?

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+		checkMagicHeader(magicHeader, (byte) 0x81); // -127
+
+		int vendorLength = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+		byte[] vendor = new byte[vendorLength];
+		input.readFully(vendor);
+		Logger.minor(this, "Vendor string is: " + new String(vendor, StandardCharsets.UTF_8));
+		int numberOfComments = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+		for (long i = 0; i < numberOfComments; i++) {
+			int commentLength = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+			byte[] comment = new byte[commentLength];
```
Bombe: Hmm, this does allow a denial-of-service attack on every node trying to filter this file… maybe we should restrict these kinds of allocations to some kind of limit? I have no idea whether we already do this anywhere else… @nextgens @Thynix @ArneBab do we?

src/freenet/client/filter/TheoraPacketFilter.java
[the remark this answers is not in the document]
Bombe: I totally would write a method like this:

```
void checkHeaderField(String name, int sizeInBits, Predicate<Int> validator) {
    int value = input.readInt(sizeInBits);
    if (!validator.apply(value)) {
        throw new UnknownContentTypeException("Header " + name + ": " + value);
    }
 }
```
That would cut down most of that method to lines like:

```
checkHeaderField("PF", 2, v -> v != 1);
checkHeaderField("Res", 3, v -> v == 0);
```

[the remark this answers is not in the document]
Olezha: But now we have Java 7 language level. It is possible to write something like this, but what we can have on code level (one line instead two) and on stack.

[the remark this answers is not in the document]
Bombe: Argh, shit, we’re _still_ on 7? Okay, then disregard that. :/

src/freenet/client/filter/TheoraPacketFilter.java
Olezha: FRED has protection on this type of bugs https://git.io/fj6zx
I also added limits here

src/freenet/client/filter/TheoraBitstreamFilter.java
Olezha: I need some video for tests. What about this one SSK@gwe30-18LwLnAUuQqwZvxSeRk70LmxzGzYQtrnifknY,Nrw4BGOZ7H4hGyfCUVkZGuK9tLp9T6G7odPXOU5aj~Y,AQACAAE/small.ogv

src/freenet/client/filter/TheoraPacketFilter.java
Bombe: That is totally not what I mean. I was thinking more along the lines of…
```
if (commentLength > 16 * 1024) { // amount is negotiable
    throw new UnknownContentException("Comment too large!");
}
```
Then again, if we only read the comment for logging, just read 50 bytes to print them and skip the rest? I’m really not sure we should read the data at all if we’re not using it.

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+
+	private CodecPacket parseCommentHeader(BitInputStream input) throws IOException {
+		Logger.minor(this, "COMMENT_HEADER");
+
+		byte[] magicHeader = new byte[1 + magicNumber.length];
+		input.readFully(magicHeader);
+		checkMagicHeader(magicHeader, (byte) 0x81); // -127
+
+		int vendorLength = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+		byte[] vendor = new byte[vendorLength];
```
Bombe: And this one I’d also like to check, along with all other occurences of using a user-supplied number to allocate memory. :)

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+		int numberOfComments = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+		for (long i = 0; i < numberOfComments; i++) {
+			int commentLength = input.readInt(32, ByteOrder.LITTLE_ENDIAN);
+			byte[] comment = new byte[commentLength];
+			input.readFully(comment);
+			Logger.minor(this, "Comment string is: " + new String(comment));
+		}
+
+		try {
+			input.readBit();
```
Bombe: And the same is valid for this here… if we don’t do anything with the information, what do we get it for? Either redundant bits are an error (in which case we should throw) or redundant bits are totally fine.

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,287 @@
...
+		}
+
+		expectedPacket = Packet.FRAME;
+	}
+
+	private void checkMagicHeader(byte[] typeAndMagicHeader, byte expectedType) throws IOException {
+		if (typeAndMagicHeader[0] != expectedType)
+			throw new UnknownContentTypeException("Header type: " + typeAndMagicHeader[0] + ", expected: " + expectedType);
+
+		for (int i=0; i < magicNumber.length; i++) {
```
Bombe: Skipped the formatter here? :)

src/freenet/client/filter/TheoraPacketFilter.java
Olezha: Spec doesn't say that redundant bits are an error or not. So as we skip comments at next lines of code, let's skip reading them. I removed this fragment.

src/freenet/client/filter/TheoraPacketFilter.java
Olezha: I rewrote this method

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,321 @@
...
+    }
+
+    // The minimum number of bits required to store a positive integer `a` in
+    // two’s complement notation, or 0 for a non-positive integer a.
+    private int ilog(int a) {
+        if (a <= 0) {
+            return 0;
+        }
+
+        int n = 0;
```
Bombe: It just occured to me that good old `Integer` has a `bitCount` method which does the same, probably. Maybe that could be used? (It doesn’t mention what it does with negative numbers but I assume it treats them as any other number.)

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -1,148 +1,321 @@
...
-				throw new UnknownContentTypeException("Packet header magicNumber[" + i + "]: " + typeAndMagicHeader[i+1]);
-		}
-	}
+    static final byte[] magicNumber = new byte[]{'t', 'h', 'e', 'o', 'r', 'a'};
+
+    enum Packet {
+        IDENTIFICATION_HEADER, COMMENT_HEADER, SETUP_HEADER, FRAME
+    }
+
+    private Packet expectedPacket = Packet.IDENTIFICATION_HEADER;
```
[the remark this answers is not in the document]
ArneBab: regarding the stash of Theora videos: You can find some here https://wiki.xiph.org/List_of_Theora_videos#Short_Films

Also yes, I have a stash of self-created ones, i.e. http://infinite-hands.draketo.de/filme/Infinite_Hands-2008-Thusnelda-2009-09-18.ogv

(@Bombe thank you for taking over this code review!)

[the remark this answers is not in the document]
Olezha: @Bombe it is should be new for every stream (Ogg file can contain also others, e.g. Vorbis).
I did not touch this place, but it is somewhere here: https://github.com/freenet/fred/blob/next/src/freenet/client/filter/OggFilter.java#L51

src/freenet/client/filter/TheoraBitstreamFilter.java
Bombe: As soon as you can integrate some of the videos Arne linked in the other comment, I think this should be good to go.

src/freenet/client/filter/TheoraPacketFilter.java
Olezha: Like this allocation not rarely occurs before calling `readFully` method. I launched the search and there are places where the `Integer.MAX_VALUE` can be present according the syntax. But also every pace should be logically checked, is it true.

So, probably this can be registered as an issue in the tracker.

nextgens: Sure, we use the same pattern elsewhere... but in this instance the code is in a content filter: we are parsing presumed-malicious streams... and allocating too much memory doing so will DoS the node (OutOfMemory exceptions...).

Surely the vendor length (and other similar allocations in any of our filters) can be restricted to something sane size-wise ;)

Thynix: At this point we don't know the complete length of the stream being filtered?

Thynix: I'm not confident we can place much of a bound on this, without adding it as an arbitrary limitation. From [the spec](https://www.theora.org/doc/libtheora-1.0alpha6/structtheora__comment.html):

> In filling in this structure, theora_decode_header() will null-terminate the user_comment strings for safety. However, the bitstream format itself treats them as 8-bit clean, and so the length array should be treated as authoritative for their length.

nextgens: A content filter doesn't have to be spec abiding ... it has to be safe.

Bombe: I’m totally with nextgens here. We don’t have to restrict it to 20 chars but something sensible, like, I don’t know, 1k, or maybe 10k shouldn’t catch too many videos (one would have to check with stuff that’s in the wild) and still prevent the file from DoSing the node.

Olezha: So, in this PR I totally removed this place (in COMMENT_HEADER state we check only magicHeader and do not read comments). Additionally I created an issue to check other places https://freenet.mantishub.io/view.php?id=7108. So I am marking this place as resolved at this PR.

src/freenet/client/filter/TheoraPacketFilter.java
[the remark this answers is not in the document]
Bombe: Sorry for dragging this up again but I think @ArneBab recently mentioned that we can actually move to Java 8 so I’d be extremely happy if this could be the first code in Fred to use lambdas ’n shit. :)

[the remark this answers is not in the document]
Bombe: And I think after this I don’t have any more comments and/or requests. :) Thanks for sticking with this!

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -249,6 +206,15 @@ private void verifyTypeAndHeader(BitInputStream input, int expectedHeaderType) t
         }
     }
 
+    private int checkHeaderField(String headerName, String fieldName,
+                                 BitInputStream input, int sizeInBits, Predicate<Integer> validator) throws IOException {
+        int value = input.readInt(sizeInBits);
+        if (validator.test(value)) {
```
Bombe: A validator should return `true`if the given value is valid, something needs to be the other way around here. :)

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -53,37 +55,37 @@ public CodecPacket parse(CodecPacket packet) throws IOException {
     private void verifyIdentificationHeader(BitInputStream input) throws IOException {
         verifyTypeAndHeader("Identification", input, 0x80); // expected -128
 
-        checkHeaderField("Identification", "VMAJ", input, 8, v -> v != 3);
+        checkHeaderField("Identification", "VMAJ", input, 8, not(v -> v != 3));
```
Bombe: The usage of `not` makes these harder to read than if they had been inverted, I think.

Olezha: On this line I selected this style because it is closer to spec, I think. Spec says:

> Read an 8-bit unsigned integer as VMAJ.  If VMAJ is not 3, then stop.This stream is not decodable according to this specification.

So, '**not 3**' and `v -> v != 3` looks closer then `v -> v == 3`. So can I please known your opinion one more time?
I will recheck other lines and think some `not`-s should be removed.

Bombe: Mostly my argument is that using the `not` function on all of the expressions does make some of them harder to read. Sure, for `not(v -> v != 3)` that is easy but an expression like (the completely made up) `not(v -> v != 3 && v != 4 || v != 17)` I have a tough time understanding which values are actually valid.

Olezha: I also removed `VREV == 1` check, as:

> Read an 8-bit unsigned integer as VREV.  If VREV is greater than 1, then this stream may contain optional features or interpretational changes documented  in  a  future  version  of  this  specification.   **Regardless**  of  the value of VREV, the stream is decodable according to this specification.

src/freenet/client/filter/TheoraPacketFilter.java
```diff
@@ -57,17 +57,17 @@ private void verifyIdentificationHeader(BitInputStream input) throws IOException
...
         checkHeaderField("Identification", "VMIN", input, 8, not(v -> v != 2));
-        checkHeaderField("Identification", "VREV", input, 8, not(v -> v > 1));
-        int FMBW = checkHeaderField("Identification", "FMBW", input, 16, not(v -> v == 0));
-        int FMBH = checkHeaderField("Identification", "FMBH", input, 16, not(v -> v == 0));
+        input.skip(8); // skip VREV
+        int FMBW = checkHeaderField("Identification", "FMBW", input, 16, v -> v > 0);
+        int FMBH = checkHeaderField("Identification", "FMBH", input, 16, v -> v > 0);
         checkHeaderField( "Identification", "PICW", input, 24, not(v -> v > FMBW * 16));
         checkHeaderField("Identification", "PICH", input, 24, not(v -> v > FMBH * 16));
         checkHeaderField("Identification", "PICX", input, 8, not(v -> v > FMBW * 16 - v));
```
Bombe: Hmm, is this not the same as `2 * v > FMBW * 16` which would be the same as `v > FMBW * 8`?

Bombe:
I think I don’t have anything more to argue about. Thanks for putting up with me! :)

src/freenet/client/filter/TheoraPacketFilter.java
Olezha: This is the same. And where does this come from:

> Read an 8-bit unsigned integer as PICX.  This MUST be no greater than (FMBW∗16−PICX).

Should it be simplified?

Bombe: No, doesn’t have to be. But there’s an unnecessary space two lines above so if you could run the formatter on anything you’ve changed? Also, for Java 8 you’d need to pull in next so the Travis build will run.

Olezha: I already merged next here. Probably these lines should be changed:
https://github.com/freenet/fred/blob/next/build.gradle#L58

Bombe: Those lines were changed in next but you botched the merge, changing them back. :)

Olezha:
> Thanks for putting up with me!

Thank you so much! This review was very useful to me.

PR #669, opened 2019-07-05
Redesign first time wizard UI

Olezha:
I want to show a new form from issue 6020. Please do not look at the template engine; I will replace it with `HTMLNode`
https://freenet.mantishub.io/view.php?id=6020


Bombe:
I would totally love to have a decent template engine in Fred, I don’t know, FreeMarker, or Thymeleaf… :)

Olezha:
We have compared both engines and think that either of them is good for our purposes.

Does anyone in core team have a strong preference in favour of FreeMaker? If not, we would suggest to use Thymeleaf, since we worked with it before. However, since FreeMaker offers similar features, so we are totally fine to go with it.

Please let us know if it is ok to start using Thymeleaf. Also, if you are not sure, let us know if you’d like a more detailed comparison to make the choice.

ArneBab:
According to https://github.com/mbosecke/template-benchmark and https://smarterco.de/java-benchmark-thymeleaf-2.1.4-vs-thymeleaf-3.0-snapshot/ , Thymeleaf is quite a bit behind other templating engines in terms of performance. Do you have experience with [pebble](https://pebbletemplates.io/)? That seems to be roughly factor 10 faster than Thymeleaf and has syntax similar to django and makes it easy to build derived pages and to make it easy for designers to adapt pages.

In our case such performance might actually matter, because the server runs locally, so there’s little network overhead, and because many people run Freenet on pretty weak machines (and it would be great to get Freenet onto Freedomboxes).

src/freenet/clients/http/staticfiles/js/firsttimewizard.js
```diff
@@ -0,0 +1,126 @@
...
+        currHeight = el.clientHeight * (getComputedStyle(el).display !== 'none');
+        ref = currHeight > this.height / 2 ? [this.height, 0] : [0, this.height];
+        start = ref[0];
+        end = ref[1];
+        disp = end - start;
+        this.el.classList[end === 0 ? 'remove' : 'add']('open');
+        this.el.style.cssText = "overflow: hidden; display: block; padding-top: 0; padding-bottom: 0";
+        init = (new Date).getTime();
+        repeat = function() {
+        var instance, ref1, repeatLoop, results, step;
```
ArneBab: here the indentation seems to be different from the logical code — the function body should be more indented.

src/freenet/clients/http/staticfiles/js/firsttimewizard.js
```diff
@@ -0,0 +1,126 @@
...
+        instance = (new Date).getTime() - init;
+        step = start + disp * instance / time;
+        if (instance <= time) {
+            el.style.height = step + 'px';
+        } else {
+            el.style.cssText = "display: " + (end === 0 ? 'none' : 'block');
+        }
+        repeatLoop = requestAnimationFrame(repeat);
+        if (ref1 = Math.floor(step), indexOf.call((function() {
+            results = [];
```
ArneBab: please extract this function from the if-condition to make the code easier to follow. Maybe something like shouldCancelAnimationFrame(step)

src/freenet/clients/http/wizardsteps/BandwidthLimit.java
```diff
@@ -25,4 +39,31 @@ public BandwidthLimit(long downBytes, long upBytes, String descriptionKey, boole
...
+	 * Calculate download and upload limit
+	 * @param bytesPerMonth monthly bandwidth limit
+	 */
+	public BandwidthLimit(long bytesPerMonth) {
+		/*
+		 * Fraction of total limit used for download. Asymptotically from 0.5 at the minimum cap to 0.8.
+		 *
+		 * FIXME: Why do we do this? It does not actually work, since
+		 * download cannot be larger than upload for any long amount
+		 * of time.
```
ArneBab: While browsing, Freenet uses realtime mode which allows for short-term bursts of much higher download bandwidth usage. People are sometimes using that to boost a specific download (by simply accessing the file interactively).

src/freenet/clients/http/wizardsteps/BandwidthManipulator.java
```diff
@@ -69,48 +69,32 @@ protected BandwidthLimit getCurrentBandwidthLimitsOrNull() {
...
-			downstreamBytes = -1;
-		} else if (downstreamBits < 8*KiB) {
-			//Nonsensically slow.
-			System.err.println("Detected downstream of "+downstreamBits+" bits/s is nonsensically slow, ignoring.");
-			downstreamBytes = -1;
-		} else {
-			downstreamBytes = downstreamBits/8;
-		}
+		if (downstreamBits < 8*KiB)
+			throw new Exception("Detected downstream of "+downstreamBits+" bits/s is nonsensically slow, ignoring.");
```
ArneBab: Maybe throw an IllegalStateException instead of a generic Exception?

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
```diff
@@ -0,0 +1,295 @@
...
+            model.put("connectToStrangers", connectToStrangers.length() > 0 ? "checked" : "");
+            model.put("haveMonthlyLimit", haveMonthlyLimit.length() > 0 ? "checked" : "");
+            model.put("downloadLimit", downloadLimit);
+            model.put("uploadLimit", uploadLimit);
+            model.put("bandwidthMonthlyLimit", bandwidthMonthlyLimit);
+            model.put("storageLimit", storageLimit);
+            model.put("setPassword", setPassword.length() > 0 ? "checked" : "");
+
+            if (downloadLimitDetected == null || uploadLimitDetected == null)
+                detectBandwidthLimit();
```
ArneBab: please always add curly braces around the if-body. This isn’t done everywhere in the existing code, but where we touch code, we should fix it.

Olezha:
> Do you have experience with [pebble](https://pebbletemplates.io/)? That seems to be roughly factor 10 faster than Thymeleaf and has syntax similar to django and makes it easy to build derived pages and to make it easy for designers to adapt pages.
>
> In our case such performance might actually matter, because the server runs locally, so there’s little network overhead, and because many people run Freenet on pretty weak machines (and it would be great to get Freenet onto Freedomboxes).

Pebble also looks fine but I don't have experience with it (this is not a problem but I can't say how comfortable it might be).

Because the server runs locally, this is not loaded part unlike ordinary sites (there is much less page requests). But performance matters anyway.

To sum up:
FreeMarker - oldest
Thymeleaf - now more popular https://java.libhunt.com/compare-pebble-vs-thymeleaf
Pebble - faster and smaller

It seems to me that all options are okey, and only personal preferences matter. I would like to try Pebble or choose Thymeleaf as I tried it before and everything was okey. Maybe you have any preferences in this?

src/freenet/clients/http/wizardsteps/BandwidthLimit.java
nextgens: Upload is limited because maxing it out increases latency... http://bufferbloat.net/

And fred (line most layered P2Ps) deals very poorly with high-latency links

src/freenet/clients/http/wizardsteps/BandwidthManipulator.java
nextgens: Don't use a RuntimeException! IllegalValueException is what you want

src/freenet/clients/http/wizardsteps/BandwidthManipulator.java
```diff
@@ -69,49 +69,37 @@ protected BandwidthLimit getCurrentBandwidthLimitsOrNull() {
...
-		Logger.normal(this, "bandwidthIndicator reports downstream " + downstreamBits + " bits/s and upstream "+upstreamBits+" bits/s.");
+		Logger.normal(bwIndicator, "bandwidthIndicator reports downstream " + downstreamBits + " bits/s and upstream "+upstreamBits+" bits/s.");
 
-		int downstreamBytes, upstreamBytes;
+		if (downstreamBits < 0 || upstreamBits < 0) {
+			throw new Exception("Reported unavailable.");
+		}
 
 		//For readability, in bits.
 		final int KiB = 8192;
```
Olezha: Hh, 8 KiB? Why do we multiply below?

src/freenet/clients/http/wizardsteps/BandwidthLimit.java
Olezha: Can I add this explanation to the comment and change FIXME to Q&A?

ArneBab:
Feel free to chose what works best to you.

Olezha:
In addition to pebble, **slf4j** and **unbescape** libraries were also required. I do not know exactly how they should be distributed properly. I put them in freenet-ext.jar If something is wrong, please let me know to correct

src/freenet/clients/http/wizardsteps/BandwidthLimit.java
ArneBab: Sure.

ArneBab:
> In addition to pebble, **slf4j** and **unbescape** libraries were also required. I do not know exactly how they should be distributed properly. I put them in freenet-ext.jar If something is wrong, please let me know to correct

We need to upload them and add them to dependencies.properties so they can be retrieved from freenet upon update (and before activating the update).

ArneBab:
marked as afterNextRelease because adding the dependencies needs more testing — so I’d like to merge this for 1486.

src/freenet/clients/http/wizardsteps/BandwidthLimit.java
```diff
@@ -25,4 +39,32 @@ public BandwidthLimit(long downBytes, long upBytes, String descriptionKey, boole
...
+	 * @param bytesPerMonth monthly bandwidth limit
+	 */
+	public BandwidthLimit(long bytesPerMonth) {
+		/*
+		 * Fraction of total limit used for download. Asymptotically from 0.5 at the minimum cap to 0.8.
+		 *
+		 * Q: Why do we do this? It does not actually work, since download cannot be larger than upload
+		 *  for any long amount of time.
+		 * A: Upload is limited because maxing it out increases latency... http://bufferbloat.net/
+		 *  And fred (line most layered P2Ps) deals very poorly with high-latency links
```
Olezha: like

ArneBab:
I had to do 1486 as emergency release, because Java on Windows broke with our jna. We’re working on improving our dependencies workflow to make this PR safer to release (currently dependencies are scattered all over the place, so every updated or added dependency is risky).

ArneBab:
I resolved the conflicts via the online editor. I hope it didn’t break anything.

PR #672, opened 2019-09-15
Turn CompressionRatioException into a checked exception to ensure tha…

ArneBab:
…t it gets handled wherever it could appear.

Handle it in Key.java.

This extends upon https://github.com/freenet/fred/pull/662 — which kept to the style that was there before.

Using an exception for control flow isn’t nice, but with a checked exception we can at least be sure that it does not pollute other code.

It is caught in the InsertCompressor and in Key (Key was missing).

### DIFF
--- a/src/freenet/support/compress/Compressor.java
+++ b/src/freenet/support/compress/Compressor.java
@@ -16,13 +16,13 @@ import freenet.support.api.BucketFactory;
  * This is for single-file compression (gzip, bzip2) as opposed to archives.
  */
 public interface Compressor {
-	
+
 	String DEFAULT_COMPRESSORDESCRIPTOR = null;
 
 	enum COMPRESSOR_TYPE implements Compressor {
-	    // WARNING: Changing non-transient members on classes that are Serializable can result in 
+	    // WARNING: Changing non-transient members on classes that are Serializable can result in
 	    // restarting downloads or losing uploads.
-	    
+
 		// Codecs will be tried in order: put the less resource consuming first
 		GZIP("GZIP", new GzipCompressor(), (short) 0),
 		BZIP2("BZIP2", new Bzip2Compressor(), (short) 1),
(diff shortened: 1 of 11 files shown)


src/freenet/support/compress/Compressor.java
```diff
@@ -132,22 +132,25 @@ public static void getCompressorDescriptor(StringBuilder sb) {
...
 		@Override
-		public Bucket compress(Bucket data, BucketFactory bf, long maxReadLength, long maxWriteLength) throws IOException, CompressionOutputSizeException {
+		public Bucket compress(Bucket data, BucketFactory bf, long maxReadLength, long maxWriteLength)
+				throws IOException, CompressionOutputSizeException, CompressionRatioException {
 			return compressor.compress(data, bf, maxReadLength, maxWriteLength);
 		}
 
 		@Override
-		public long compress(InputStream is, OutputStream os, long maxReadLength, long maxWriteLength) throws IOException, CompressionOutputSizeException {
+		public long compress(InputStream is, OutputStream os, long maxReadLength, long maxWriteLength)
```
Olezha: This method shouldn't throw `CompressionRatioException`. But the next one should. The next method should be [described](https://github.com/redwerk/fred/commit/26389c1f152b086a6449d7b7f95a432e620e4270#diff-127ee35d8f28c09f0b9ba01bcd477492R202) in a different way. After this not the best [realization](https://github.com/redwerk/fred/commit/26389c1f152b086a6449d7b7f95a432e620e4270#diff-dddbae15c941315e0069a6f233029e27R11) of their interaction might happen.

Another implementation approach: to add some callback and to make a decision about compression continuation at higher level. But in this case a mechanism of compression process will also be needed. It may cause more code without any benefit.

It seems to okey when `compress` with `minimumCompressionPercentage` throw exceptions because it should compress data at least with the set effect, but cannot do this.

ArneBab:
Damn, I accidently pushed this unfinished when merging the cleanup pull-request :-(

ArneBab:
I’m having a look at https://github.com/freenet/fred/compare/next...redwerk:frn-80#diff-8a56acd98e351799ac2e80ab29b8de0bR87

Could you file a pull-request?

PR #674, opened 2019-09-17
Upgrade of the connection speed should be offered to a user

Olezha:
Suggest increasing Bandwidth Limit if rise of speed was auto-detected

> offer upgrade of the connection speed on upgrade, if auto-detected speed is much higher than the set speed, or even better: if the detected speed increased significantly since the last offer.


ArneBab:
Thank you for your pull-request! This is a feature we’ve needed for a long time.

Asking once a day might be a bit too much, though. A natural time to ask would be when Freenet gets updated to a new version.

Please merge next into your branch so we get the full CI (it contains fixes for our travis setup).

src/freenet/node/BandwidthManager.java
```diff
@@ -0,0 +1,109 @@
...
+                    int detectedInputBandwidth = bandwidthIndicator.getDownstreamMaxBitRate() / 8;
+                    int detectedOutputBandwidth = bandwidthIndicator.getUpstramMaxBitRate() / 8;
+
+                    int currentInputBandwidth = node.config.get("node").getInt("inputBandwidthLimit");
+                    int currentOutputBandwidth = node.config.get("node").getInt("outputBandwidthLimit");
+
+                    if (detectedInputBandwidth > currentInputBandwidth * 3 &&
+                            detectedInputBandwidth > lastOfferedInputBandwidth * 3 ||
+                        detectedOutputBandwidth > currentOutputBandwidth * 3 &&
+                                detectedOutputBandwidth > lastOfferedOutputBandwidth * 3) {
```
Olezha: Speed check happens once a day, yes. But the alert will appear if the speed is significantly greater than previously offered one.

ArneBab: that sounds good; I had missed that — thank you!

PR #675, opened 2019-10-02
CompressionRatioException restriction

Olezha:
Related to #662 and  #672

### DIFF
--- a/src/freenet/support/compress/AbstractCompressor.java
+++ b/src/freenet/support/compress/AbstractCompressor.java
@@ -19,6 +19,9 @@ abstract class AbstractCompressor implements Compressor {
 
     void checkCompressionEffect(long rawDataVolume, long compressedDataVolume, int minimumCompressionPercentage)
         throws CompressionRatioException {
+        assert rawDataVolume != 0;
+        assert minimumCompressionPercentage != 0;
+
         long compressionPercentage = 100 - compressedDataVolume * 100 / rawDataVolume;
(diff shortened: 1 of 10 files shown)


src/freenet/support/compress/AbstractCompressor.java
```diff
@@ -7,12 +7,21 @@
...
     }
 
     void checkCompressionEffect(long rawDataVolume, long compressedDataVolume, int minimumCompressionPercentage)
         throws CompressionRatioException {
+        assert rawDataVolume != 0;
+        assert minimumCompressionPercentage != 0;
+
         long compressionPercentage = 100 - compressedDataVolume * 100 / rawDataVolume;
         if (compressionPercentage < minimumCompressionPercentage) {
             throw new CompressionRatioException("Compression has no effect. Compression percentage: " + compressionPercentage);
```
Olezha: > Using an exception for control flow isn’t nice

This is where `CompressionRatioException` came from.
Maybe it's worth changing.

ArneBab:
We should add a new compatibility mode, similar to this: https://github.com/freenet/fred/pull/677/commits/bebc400bf2731e2b77a446a3b3599671ef690d00

(I’ll need review by nextgens or bombe for this)

ArneBab:
Looks good. No need to add a new compatibility mode since the old LZMA has been deprecated for almost a decade. Here’s the change to remove it from checked methods: https://github.com/freenet/fred/pull/679

PR #676, opened 2019-10-03
Fix config option with dimension (duration)

Olezha:
I noticed a bug that I added. This is a fix, but not ideal. Therefore, I will be glad of any help.
Related points:
- Two toString
  - config.IntOption#toDisplayString
  - config.IntOption#toString
- Single toInt - config.IntOption#parseString
- ConfigToadlet#L329

### DIFF
--- a/src/freenet/support/Fields.java
+++ b/src/freenet/support/Fields.java
@@ -721,15 +721,11 @@ public abstract class Fields {
 			case SIZE:
 				return parseInt(s);
 			case DURATION:
-				try {
-					long durationInMillis = TimeUtil.toMillis(s);
-					if ((int) durationInMillis == durationInMillis) {
-						return (int) durationInMillis;
-					}
-					throw new ArithmeticException("integer overflow");
-				} catch (NumberFormatException e) {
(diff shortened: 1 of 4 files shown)


src/freenet/support/Fields.java
```diff
@@ -721,10 +721,15 @@ public static int parseInt(String s, Dimension dimension) throws NumberFormatExc
...
-					return (int) durationInMillis;
-				throw new ArithmeticException("integer overflow");
+				try {
+					long durationInMillis = TimeUtil.toMillis(s);
+					if ((int) durationInMillis == durationInMillis) {
+						return (int) durationInMillis;
+					}
+					throw new ArithmeticException("integer overflow");
+				} catch (NumberFormatException e) {
+					return parseInt(s);
```
ArneBab: shouldn’t the NumberFormatException be caught where the method is used — and for example lead to ignoring he new value?

Olezha: yes, that’s much better. the Fields class is not relevant to this issue. and the fix code is next to the cause.

PR #681, opened 2019-11-03
Survive missing bookmarks

ArneBab:
Bug: Freenet failed to start when the bookmarks.dat file was missing. This fixes the NPE.

### DIFF
--- a/src/freenet/clients/http/bookmark/BookmarkManager.java
+++ b/src/freenet/clients/http/bookmark/BookmarkManager.java
@@ -367,14 +367,18 @@ public class BookmarkManager implements RequestClient {
 
 			sfs = toSimpleFieldSet();
 		}
-		try (FileOutputStream fos = new FileOutputStream(backupBookmarksFile)){
-
+		FileOutputStream fos = null;
+		try {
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/bookmark/BookmarkManager.java
```diff
@@ -369,19 +367,14 @@ public void storeBookmarks() {
...
 		}
-		FileOutputStream fos = null;
-		try {
-			fos = new FileOutputStream(backupBookmarksFile);
+		try (FileOutputStream fos = new FileOutputStream(backupBookmarksFile)){
+
 			sfs.writeToBigBuffer(fos);
-			fos.close();
-			fos = null;
 			if(!FileUtil.renameTo(backupBookmarksFile, bookmarksFile))
```
nextgens: I'd bet that this breaks on windows. If you haven't closed the file, odds are you can't move it.

ArneBab:
This avoids a NPE if there is no bookmarks file. Nothing more, except for the cleanup with try-with-resources that was actually wrong.

PR #685, opened 2019-12-01
Sanitize fdtracker unserialization

ArneBab:
This might fix the startup loop bug likely caused as a side effect of fixing the lose-all-downloads-on-restart bug.

Symptoms of the bug: Massive memory requirement on startup, continuously restart.

### DIFF
--- a/src/freenet/support/io/PooledFileRandomAccessBuffer.java
+++ b/src/freenet/support/io/PooledFileRandomAccessBuffer.java
@@ -4,6 +4,7 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.io.RandomAccessFile;
 import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


ArneBab:
Tested by BadLuckBob now, but did not resolve the problem :-/

gits7r:
I have run into the same bug, massive memory requirement on startup, continuously restarts but I did not have any ongoing downloads or anything in the download queue. In fact, I did not download anything at all from that particular node never (not just before the last restart before the continuous restarts. So, I doubt there was some queue that needed deserialization when I tried to start fred and did not work. We have to check and make sure.

Is `fdtracker` also used by other kind of (automated) downloads such as datastore refresh and not just for manual downloads requested by users by presenting the SSK (or keys)?

ArneBab:
This is in the PooledFileRandomAccessBuffer, so yes, this is used for example when inserting splitfiles.

ArneBab:
Closed in favor of https://github.com/freenet/fred/pull/686 which actually worked.

PR #686, opened 2019-12-25
move the deserialization guard from FDTracker into PooledFileRandomAccessBuffer

ArneBab:
Hypothesis: deserialization of PooledFileRandomAccessBuffers could cause multiple deserialization of the fds variable, including its closables, so it could deserialize the buffers multiple times.

Fix: transient fds, and deserialization that sets the static FDTracker.

This was not measured.


gits7r:
How can this bug occur even if you have no downloads queue at all, or never downloaded anything from freenet? From the datastore sync or what else can trigger the same bug?

ArneBab:
Because PooledFileRandomAccessBuffers are used in several other places when it’s about working with data.

nextgens:
lgtm

ArneBab:
Thank you for your review!

PR #688, opened 2019-12-30
Update links to jSite

Bombe:
This updates the links to jSite to the current version of both the JAR file and the freesite. Additionally, it also removes the link to the old, non-HTTPS download site that doesn’t exist anymore.


ArneBab:
looks good — thank you!

PR #689, opened 2020-01-16
Higher bandwith to darknet peers

mkb2091:
Not sure if this has already been considered and decided against, but I thought giving higher bandwidth to darknet peers would aid darknet-only nodes relying on hybrid nodes.

It could also make it more beneficial to create darknet connections when both nodes are hybrid since they would prioritize each others bandwidth over other nodes.

I just chose 3 as the multiplier, but not sure if it should be configurable, or calculated at run time, or a different hard-coded number


Bombe:
Please don’t mix completely unrelated things, i.e. separate the bookmark update into its own PR.

mkb2091:
Sorry, hadn't spotted I'd done it. I made the branch after the bookmark update. I've removed that commit

mkb2091:
I'm not sure how the commits I made could have caused  "Could not HEAD 'https://repo.maven.apache.org/maven2/com/android/tools/build/gradle/maven-metadata.xml'. Received status code 403 from server: Forbidden"

ArneBab:
Looks good. This has been on the "we should try this" list for a long time.

Thank you for stepping up!

PR #690, opened 2020-01-24
Translation;  Using Transifex for translations

hernic:
Hello,

here is a message from the Transifex Team

> The relevance of the translation
pirgooo
jan. 24, 2020, 10:31 matin
Hello friends. Judging by the statistics, nobody is practically engaged in translation now. Is the project still relevant? And in the text itself the data is very old.
Do the authors of the project somehow coordinate with translation teams?

https://www.transifex.com/otf/teams/37175/discussions/19499/

by the way what is the state of Transifex translation integratoin into fred ?

Thanks


Bombe:
This pull request is a bit of a mess, don’t you think? Is it even a real pull request or did you just want to start a discussion? I don’t actually see any recent commits on it, only a lot of things that don’t have anything to do with the translation.

Regarding the translation on Transifex:
* I don’t think pirgooo is a member of the Transifex team – unless by “Transifex team” you mean “team of Freenet translators” and not “people working for Transifex.”
* I myself have recently started using Transifex, both for Freenet as well as for Sone and while it’s true that translations are only slowly coming in it will be that way for everything that is not anonymous and relying on participation of our users. I have no idea yet how Freenet is being integrated with the translations and this discussion is probably important but a fake pull request is the wrong place for this. Please open an issue for it.

Bombe:
Okay, apparently we don’t even have issues activated here, that’s a pity. We do have [an issue tracker on MantisHub](https://bugs.freenetproject.org/), though.

PR #691, opened 2020-01-29
prefix links in m3u-files with host+port of the node to allow for local playback with arbitrary keys

ArneBab:
This leaks host+port of the node into the file,
but otherwise we would not be able to support arbitrary keys:
we would have to stick to files within the same prefix (i.e. SSK@…/) as the m3u file.


ArneBab:
^ this gives me working m3u playlists over Freenet, usable like normal web-radio. It should get us to the point where we can live-stream audio.

src/freenet/client/filter/M3UFilter.java
```diff
@@ -141,7 +141,7 @@ public void readFilter(InputStream input, OutputStream output, String charset, H
                                 // impaired.
                                 if (cb instanceof GenericReadFilterCallback) {
                                     try {
-                                        filtered = ((GenericReadFilterCallback)cb).makeURIAbsolute(filtered);
+                                        filtered = ((GenericReadFilterCallback)cb).makeURIAbsoluteWithHostAndPort(filtered);
```
Bombe, line 144: Oh, and there is no test for this, either. So before adding new tests for this feature you would have to fix the test which relies on an incorrect base URI being handed in to the filter in order to test the URL resolving.

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -280,6 +281,18 @@ public String makeURIAbsolute(String uri) throws URISyntaxException{
 		return baseURI.resolve(URIPreEncoder.encodeURI(uri).normalize()).toASCIIString();
 	}
 
+	public String makeURIAbsoluteWithHostAndPort(String uri) throws URISyntaxException{
+		// FIXME: this breaks abstraction really badly and depends on implementation details.
+		String fproxyAuthority = "";
+		if (linkFilterExceptionProvider instanceof  SimpleToadletServer) {
+			fproxyAuthority = ((SimpleToadletServer) linkFilterExceptionProvider).getURL();
```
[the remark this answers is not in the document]
ArneBab: I will — thank you!

ArneBab:
Here’s an example with all working keys:

`USK@sDzl0PU80nmJrgyx2kRtXROaAfgYa8~cQQXCIblGQWw,6WvK00wFVFfky60q9kEmFxUi3t9fL3RV18dgswGbcU0,AQACAAE/freenet-stream-mp3s/6/stream.m3u?type=text/plain`

With this change you can now actually play the stream with 
`mpv http://localhost:8888/freenet:USK@sDzl0PU80nmJrgyx2kRtXROaAfgYa8~cQQXCIblGQWw,6WvK00wFVFfky60q9kEmFxUi3t9fL3RV18dgswGbcU0,AQACAAE/freenet-stream-mp3s/7/stream.m3u`

src/freenet/client/filter/M3UFilter.java
ArneBab: This should now be fixed; I now use the host (IP+Port) from the request headers.

src/freenet/client/filter/GenericReadFilterCallback.java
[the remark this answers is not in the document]
ArneBab: This is fixed, too — also by moving to using the data from the request.

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -280,6 +281,18 @@ public String makeURIAbsolute(String uri) throws URISyntaxException{
 		return baseURI.resolve(URIPreEncoder.encodeURI(uri).normalize()).toASCIIString();
 	}
 
+	public String makeURIAbsoluteWithHostAndPort(String uri) throws URISyntaxException{
+		// FIXME: this breaks abstraction really badly and depends on implementation details.
```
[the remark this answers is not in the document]
ArneBab: this is now gone :-)

test/freenet/client/filter/m3u/safe.m3u
```diff
@@ -15,7 +15,7 @@ http://127.0.0.1:8888/CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKW
 http://127.0.0.1:8888/CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKWMvedueAKLkyv0S83xOrGUT4S7FaN5J3nI,AAMC--8/next.m3u
 /CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKWMvedueAKLkyv0S83xOrGUT4S7FaN5J3nI,AAMC--8/Infinite-Hands-free-software.mp3
 /CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKWMvedueAKLkyv0S83xOrGUT4S7FaN5J3nI,AAMC--8/next.m3u
-CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKWMvedueAKLkyv0S83xOrGUT4S7FaN5J3nI,AAMC--8/Infinite-Hands-free-software.mp3
+/CHK@qtpnyOzrg~aiAbnw6AoKbX8ZlTCwznjY4G1y-JiiCgg,b0wkEgRPKWMvedueAKLkyv0S83xOrGUT4S7FaN5J3nI,AAMC--8/Infinite-Hands-free-software.mp3
```
ArneBab, line 18: I no longer support having plain keys without / in front. This wasn’t well-defined (unclear whether it would be a file in the base-uri which is named after the key or a key of its own).

test/freenet/client/filter/m3u/unsafe_madesafe.m3u
```diff
@@ -1,4 +1,4 @@
-http://localhost:8888/external-link/?_CHECKED_HTTP_=http://betray.user/simplepathname.mp3
+#bad-uri-removed
```
ArneBab, line 1: same here: no longer supported, since the simpler "just remove it" works sufficiently well (and we have no GUI for "let’s check whether the user really wants this" anyway, so this part of the test wasn’t correct, but is now).

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -274,7 +260,49 @@ public String processURI(String u, String overrideType, boolean forBaseHref, boo
...
+			throw new CommentException(l10n("couldNotParseURIWithError", "error", e1.getMessage()));
+		}
+		if (uri.getHost() == null) {
+			String scheme = uri.getScheme() != null ? uri.getScheme() : "http";
+			return scheme + "://" + forceHostPort + filtered;
+		}
+		return filtered;
+	}
+
+	private String processBookmark(HTTPRequest req) throws CommentException {
```
ArneBab, line 283: this is just drive-by cleanup.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -224,8 +235,17 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
...
 	 *             If data is invalid (e.g. corrupted file) and the filter have no way to recover.
 	 */
-	public static FilterStatus filter(InputStream input, OutputStream output, String typeName, URI baseURI, FoundURICallback cb, TagReplacerCallback trc , String maybeCharset, LinkFilterExceptionProvider linkFilterExceptionProvider) throws UnsafeContentTypeException, IOException {
-		return filter(input, output, typeName, maybeCharset, new GenericReadFilterCallback(baseURI, cb,trc, linkFilterExceptionProvider));
+	public static FilterStatus filter(
+			InputStream input,
+			OutputStream output,
+			String typeName,
+			URI baseURI,
+			String host,
```
ArneBab, line 243: change: added host.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -200,8 +203,16 @@ public static FilterMIMEType getMIMEType(String mimeType) {
...
 	 *             If data is invalid (e.g. corrupted file) and the filter have no way to recover.
 	 */
-	public static FilterStatus filter(InputStream input, OutputStream output, String typeName, URI baseURI, FoundURICallback cb, TagReplacerCallback trc , String maybeCharset) throws UnsafeContentTypeException, IOException {
-		return filter(input, output, typeName, baseURI, cb, trc, maybeCharset, null);
+	public static FilterStatus filter(
+			InputStream input,
+			OutputStream output,
+			String typeName,
+			URI baseURI,
+			String host,
```
ArneBab, line 211: change: added host.

ArneBab:
To simplify review, disable display of whitespace changes. IntelliJ auto-cleaned up a lot of empty-except-for-spaces lines.

Bombe:
> To simplify review, disable display of whitespace changes. IntelliJ auto-cleaned up a lot of empty-except-for-spaces lines.

Yeah, don’t do any large formatting changes *in the middle* of a whole bunch of changes, this will make this quite unhandy to review…

ArneBab:
Those formatting changes came in unintentionally via IntelliJ auto-fixes and are pretty hard to undo after the commit is done :-/

I could try a rebase -i with squash and edit.

src/freenet/client/filter/FilterCallback.java
```diff
@@ -26,18 +26,28 @@
 	 */
 	public String processURI(String uri, String overrideType, boolean noRelative, boolean inline) throws CommentException;
 
+	/**
```
Bombe, line 29: Adding methods to public interfaces is dangerous… I didn’t find a match in anything I have checked out but we can’t really know what we’ll break with this as long as the interface is public. 😕

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -274,7 +260,49 @@ public String processURI(String u, String overrideType, boolean forBaseHref, boo
...
+		URI uri;
+		String filtered;
+		try {
+			filtered = processURI(makeURIAbsolute(u), overrideType, true, false);
+			uri = URIPreEncoder.encodeURI(filtered).normalize();
+		} catch (URISyntaxException e1) {
+			if(logMINOR) Logger.minor(this, "Failed to parse URI: "+e1);
+			throw new CommentException(l10n("couldNotParseURIWithError", "error", e1.getMessage()));
+		}
+		if (uri.getHost() == null) {
```
Bombe, line 276: I’m not sure I understand this block… in what cases does a URI not have a host? Do basically all freenet URIs not have a host?

src/freenet/client/filter/GenericReadFilterCallback.java
```diff
@@ -274,7 +260,49 @@ public String processURI(String u, String overrideType, boolean forBaseHref, boo
...
+			String scheme = uri.getScheme() != null ? uri.getScheme() : "http";
```
Bombe, line 277: You know, this might not be enough… we would need to get the protocol from the toadlet container as well, including using the X-Forwarded-Proto HTTP header. (Use case: I have freenet running behind a reverse proxy that does SSL for me so fred doesn’t know anything about that but the proxy sets a header fred can use to reconstruct valid URLs.)

src/freenet/client/filter/GenericReadFilterCallback.java
Bombe: Okay, after briefly checking `processURI` I think it’s clear to me now. :)

src/freenet/client/filter/M3UFilter.java
```diff
@@ -116,65 +117,63 @@ public void readFilter(InputStream input, OutputStream output, String charset, H
...
+                                    // HTML content instead).
+                                    if (!filtered.contains(ExternalLinkToadlet.PATH)
+                                        && !filtered.contains(ExternalLinkToadlet.magicHTTPEscapeString)) {
+                                        if (filtered.contains("?")) {
+                                            filtered += "&";
+                                        } else {
+                                            filtered += "?";
+                                        }
+                                        filtered += "max-size=" + Long.valueOf(
+                                            MAX_LENGTH_NO_PROGRESS).toString();
```
Bombe, line 161:
```suggestion
                                        filtered += "max-size=" + String.valueOf(MAX_LENGTH_NO_PROGRESS);
```

Bombe (changes requested):
Not completely done, will review more tomorrow!

src/freenet/client/async/USKRetriever.java
```diff
@@ -133,7 +133,7 @@ public void onSuccess(StreamGenerator streamGenerator, ClientMetadata clientMeta
 				pipeOut = new PipedOutputStream(pipeIn);
 				decompressorManager = new DecompressorThreadManager(pipeIn, decompressors, maxLen);
 				pipeIn = decompressorManager.execute();
-				ClientGetWorkerThread worker = new ClientGetWorkerThread(new BufferedInputStream(pipeIn), output, null, null, null, false, null, null, null, context.linkFilterExceptionProvider);
+				ClientGetWorkerThread worker = new ClientGetWorkerThread(new BufferedInputStream(pipeIn), output, null, null, ctx.host,null, false, null, null, null, context.linkFilterExceptionProvider);
```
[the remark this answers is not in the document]
Bombe: (And there’s three more of those above.)

src/freenet/client/filter/M3UFilter.java
```diff
@@ -116,65 +117,63 @@ public void readFilter(InputStream input, OutputStream output, String charset, H
...
-                                // clients could be tricked into
-                                // accessing local files (and some
-                                // just don't work, especially not
-                                // with downloaded files). This
-                                // can however make downloaded
-                                // files leak information about
-                                // the local setup (host and
-                                // port).
+                            boolean lineIsEmpty = fileIndex == 0;
+                            if (!lineIsEmpty) {
```
Bombe, line 121: If `fileIndex` is `0` at this point it means that not a single byte was read for this line which means that the first character read was a New Line which also means we already aborted this loop in line 96.

src/freenet/node/NodeClientCore.java
```diff
@@ -1980,7 +1980,7 @@ public String getMyName() {
 	public FilterCallback createFilterCallback(URI uri, FoundURICallback cb) {
 		if(logMINOR)
 			Logger.minor(this, "Creating filter callback: " + uri + ", " + cb);
-		return new GenericReadFilterCallback(uri, cb,null, toadletContainer);
+		return new GenericReadFilterCallback(uri, cb, toadletContainer);
```
Bombe, line 1983: I would suggest reverting this and instead getting rid of that additional constructor because I don’t really see a pressing need for it… I believe an additional constructor will cause more confusion than a `null` parameter.

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -1111,7 +1112,8 @@ private void charsetTestUnsupported(String charset) throws DataFilterException,
 		detectedCharset = ContentFilter.detectCharset(bytes, bytes.length, cssMIMEType, null);
 		assertTrue("Charset detected \""+detectedCharset+"\" should be unknown testing unsupported charset \""+charset+"\" from ContentFilter.detectCharset bom=\""+bomCharset+"\"", charset == null || "utf-8".equalsIgnoreCase(detectedCharset));
 		try {
-			FilterStatus filterStatus = ContentFilter.filter(inputStream, outputStream, "text/css", new URI("/CHK@OR904t6ylZOwoobMJRmSn7HsPGefHSP7zAjoLyenSPw,x2EzszO4Kqot8akqmKYXJbkD-fSj6noOVGB-K2YisZ4,AAIC--8/1-works.html"), null, null, null);
+			FilterStatus filterStatus = ContentFilter.filter(inputStream, outputStream, "text/css", new URI("/CHK@OR904t6ylZOwoobMJRmSn7HsPGefHSP7zAjoLyenSPw,x2EzszO4Kqot8akqmKYXJbkD-fSj6noOVGB-K2YisZ4,AAIC--8/1-works.html"),
+          null,null, null, null);
```
Bombe, line 1116: Missed another space here. :)

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -282,7 +282,9 @@ public void testEvilCharset() throws IOException {
 		FileOutputStream fos;
 		try {
 			ArrayBucket out = new ArrayBucket();
-			filter.readFilter(new ArrayBucket(total).getInputStream(), out.getOutputStream(), "UTF-16", null, null);
+			filter.readFilter(new ArrayBucket(total).getInputStream(), out.getOutputStream(), "UTF-16", null,
+          null,
+          null);
```
Bombe, line 287: Adding these two lines doesn’t do *anything* for readability here…

test/freenet/client/filter/JPEGFilterTest.java
```diff
@@ -22,7 +22,9 @@ public void testThatAThumbnailExtensionCodeIsPreserved() throws IOException {
 		byte[] jpegFile = createValidJpegFileWithThumbnail();
 		InputStream inputStream = new ByteArrayInputStream(jpegFile);
 		ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-		jpegFilter.readFilter(inputStream, outputStream, "UTF-8", new HashMap<String, String>(), new NullFilterCallback());
+		jpegFilter.readFilter(inputStream, outputStream, "UTF-8", new HashMap<String, String>(),
+        null,
+        new NullFilterCallback());
```
Bombe, line 27: Same here, readability didn’t increase.

test/freenet/client/filter/M3UFilterTest.java
```diff
@@ -18,9 +18,10 @@
     };
 
     private static final String BASE_URI_PROTOCOL = "http";
+    private static final String HOST_PORT = "localhost:8888";
     private static final String BASE_URI_CONTENT = "localhost:8888";
     private static final String BASE_KEY = "USK@0I8gctpUE32CM0iQhXaYpCMvtPPGfT4pjXm01oid5Zc,3dAcn4fX2LyxO6uCnWFTx-2HKZ89uruurcKwLSCxbZ4,AQACAAE/FakeM3UHostingFreesite/23/";
-    private static final String BASE_URI = BASE_URI_PROTOCOL+"://"+BASE_URI_CONTENT+'/'+BASE_KEY;
+    private static final String BASE_URI = '/'+BASE_KEY;
```
Bombe, line 24: With this change, `BASE_URI_PROTOCOL` and `BASE_URI_CONTENT` are now unused; please remove them.

Bombe: Also, spaces around operators, please! :)

Bombe (changes requested):

src/freenet/client/FetchContext.java
```diff
@@ -114,24 +114,27 @@
...
 			long curMaxTempLength, int maxMetadataSize, int maxRecursionLevel, int maxArchiveRestarts, int maxArchiveLevels,
 			boolean dontEnterImplicitArchives,
 			int maxSplitfileBlockRetries, int maxNonSplitfileRetries, int maxUSKRetries,
 			boolean allowSplitfiles, boolean followRedirects, boolean localRequestOnly,
 			boolean filterData, int maxDataBlocksPerSegment, int maxCheckBlocksPerSegment,
 			BucketFactory bucketFactory,
 			ClientEventProducer producer,
-			boolean ignoreTooManyPathComponents, boolean canWriteClientCache, String charset, String overrideMIME) {
-		this.blocks = null;
+			boolean ignoreTooManyPathComponents, boolean canWriteClientCache, String charset, String overrideMIME,
```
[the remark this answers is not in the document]
ArneBab: It’s now always called in the constructor, because that’s how it should already have been :-)

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -580,6 +580,7 @@ private void innerHandleMethodGET(URI uri, HTTPRequest httprequest, ToadletConte
 		}
 
 		FetchContext fctx = getFetchContext(maxSize);
+		fctx.host = ctx.getHeaders().get("host");
```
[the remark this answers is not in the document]
ArneBab: Did so — it’s now going into getFetchContext() which uses it in the constructor.

src/freenet/client/async/ClientGetWorkerThread.java
```diff
@@ -36,6 +36,7 @@
 public class ClientGetWorkerThread extends Thread {
 
 	private InputStream input;
+	private String host;
```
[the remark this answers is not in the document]
ArneBab: It’s final(-ly) hostAndPort now — thank you!

src/freenet/client/filter/ContentFilter.java
```diff
@@ -247,11 +267,12 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
 	 * @throws IllegalStateException
 	 *             If data is invalid (e.g. corrupted file) and the filter have no way to recover.
 	 */
-	public static FilterStatus filter(InputStream input, OutputStream output, String typeName, String maybeCharset, FilterCallback filterCallback) throws UnsafeContentTypeException, IOException {
+	public static FilterStatus filter(InputStream input, OutputStream output, String typeName, String maybeCharset, String hostPort, FilterCallback filterCallback) throws UnsafeContentTypeException, IOException {
 		if(logMINOR) Logger.minor(ContentFilter.class, "Filtering data of type"+typeName);
 		String type = typeName;
 		String options = "";
 		String charset = null;
+		// mimeType params
```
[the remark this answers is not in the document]
ArneBab: :-)

I now replaced the comment by actually giving the variable a useful name: otherMimeTypeParams.

src/freenet/client/filter/M3UFilter.java
ArneBab: Except if the line was terminated by \r\n (I tried the cleaner version before).

Bombe: Oh, right. Carry on then!

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -984,6 +984,13 @@ public void run() {
 		}
 	}
 
+	private String getSchemeHostAndPort(ToadletContext ctx) {
```
Bombe, line 987: This feels unfinished because it doesn’t do what the name says it’ll do and it’s also not used like it does what the name says it does. Please finish.

Bombe (changes requested):

src/freenet/client/async/SingleFileFetcher.java
```diff
@@ -915,7 +915,7 @@ public void onSuccess(StreamGenerator streamGenerator, ClientMetadata clientMeta
 					pipeOut.connect(pipeIn);
 					DecompressorThreadManager decompressorManager =  new DecompressorThreadManager(pipeIn, decompressors, maxLen);
 					pipeIn = decompressorManager.execute();
-					ClientGetWorkerThread worker = new ClientGetWorkerThread(new BufferedInputStream(pipeIn), output, null, null, null, false, null, null, null, context.linkFilterExceptionProvider);
+					ClientGetWorkerThread worker = new ClientGetWorkerThread(new BufferedInputStream(pipeIn), output, null, null , ctx.getHostAndPort(),null, false, null, null, null, context.linkFilterExceptionProvider);
```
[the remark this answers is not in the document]
ArneBab: ah, yes, damn all that efficient automation :-) Thank you!

test/freenet/client/filter/CSSParserTest.java
ArneBab: ah… that’s not even visible in IntelliJ (because of parameter-name showing magic).

src/freenet/clients/http/FProxyToadlet.java
ArneBab: yes, it is unfinished. The required changes for Forwarded grew bigger.

test/freenet/client/filter/ContentFilterTest.java
ArneBab: They are an indication again that IntelliJ does not actually do AST manipulation for everything, but instead does a lot of things with plain string replacement. Otherwise it would either have all parameters on their own lines or all on the same line or a consistent spreading over two lines.

src/freenet/clients/http/FProxyToadlet.java
ArneBab: Now it’s in, but I cannot test it, because I don’t have a proxy to check.

src/freenet/client/filter/GenericReadFilterCallback.java
ArneBab: I hope that this is implemented correctly now.

test/freenet/client/filter/M3UFilterTest.java
ArneBab: done :-)

test/freenet/client/filter/CSSParserTest.java
ArneBab: fixed

src/freenet/node/NodeClientCore.java
ArneBab: you’re right — a single-use constructor doesn’t really help here.

src/freenet/client/filter/M3UFilter.java
ArneBab: It’s even simpler: I can just leave out the explicit valueOf, since the long will be cast to a string anyway.

src/freenet/client/filter/FilterCallback.java
ArneBab: Here I think I actually need it.

src/freenet/client/FetchContext.java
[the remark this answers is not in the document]
ArneBab: finally final now :-)

src/freenet/client/FetchContext.java
```diff
@@ -404,6 +413,11 @@ public FetchContext(DataInputStream dis) throws StorageFormatException, IOExcept
         cooldownRetries = dis.readInt();
         cooldownTime = dis.readLong();
         ignoreUSKDatehints = dis.readBoolean();
+        s = dis.readUTF();
```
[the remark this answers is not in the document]
ArneBab: I don’t really understand this — does it need additional checks beyond what is done for the other fields?

src/freenet/client/FetchContext.java
```diff
@@ -114,24 +114,27 @@
...
 	 * RequestScheduler.COOLDOWN_PERIOD, because ULPRs will ensure rapid success
 	 * with that interval or less. */
 	private long cooldownTime;
 
 	/** Ignore USK DATEHINTs */
 	public boolean ignoreUSKDatehints;
-	
-	public FetchContext(long curMaxLength,
+	/** host and port (authority) */
+  public String host;
```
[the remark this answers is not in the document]
ArneBab: it’s now actually schemeHostAndPort :-)

src/freenet/client/filter/ContentDataFilter.java
```diff
@@ -12,19 +12,20 @@
...
-	 * IMPORTANT Implementation note: The InputStream may be a PipedInputStream 
-	 * (or conceivably even a network stream). Implementations MUST NOT ASSUME 
+	 * IMPORTANT Implementation note: The InputStream may be a PipedInputStream
+	 * (or conceivably even a network stream). Implementations MUST NOT ASSUME
 	 * that input.available() == 0 => EOF!
 	 * @param input Stream to read potentially unsafe data from.
 	 * @param output Stream to write safe (but possibly incomplete) data to.
 	 * @param charset Character set of the data if appropriate for this MIME type.
 	 * @param otherParams Other type parameters if appropriate.
+	 * @param hostPort
```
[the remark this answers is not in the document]
ArneBab: added now — thanks!

test/freenet/client/filter/M3UFilterTest.java
```diff
@@ -56,6 +59,7 @@ public void testSuiteTest() throws IOException {
                 assertTrue("Creating URI from BASE_URI " + BASE_URI + " failed", false);
             }
         }
+        assertTrue("All files should be readable, but at least " + unreadableFiles.size() + " files could not be read: " + unreadableFiles.toString(), unreadableFiles.isEmpty());
```
Bombe, line 62: No, this is getting worse all the time, don’t add more asserts, let the test crash & burn if *the test setup* fails. :)

Bombe (changes requested):

src/freenet/client/filter/ContentFilter.java
```diff
@@ -247,18 +267,19 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
...
 		String charset = null;
-		HashMap<String, String> otherParams = null;
+		HashMap<String, String> otherMimeTypeParams = null;
 		input = new BufferedInputStream(input);
 
 		// First parse the MIME type
 
 		int idx = type.indexOf(';');
 		if(idx != -1) {
+			// TODO: extract into parseHeaderOptions method, also needed for the Forwarded header.
```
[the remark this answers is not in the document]
ArneBab: Arg, no, I fixed this but forgot to remove the TODO.

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -984,6 +989,49 @@ public void run() {
...
+						? headers.get("X-Forwarded-Proto")
+				    : uriScheme != null && !uriScheme.trim().isEmpty() ? uriScheme : "http");
+		String host = forwarded.getOrDefault("proto",
+				headers.containsKey("X-Forwarded-For")
+						? headers.get("X-Forwarded-For")
+				    : uriHost != null && !uriHost.trim().isEmpty() ? uriHost : headers.get("host"));
+		return protocol + "://" + host;
+	}
+
+	private Map<String, String> parseForwardedHeader(String forwarded) {
```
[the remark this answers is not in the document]
ArneBab: Ah, yes, I misread the spec as having , between multiple entries between the fields.

Do you have a suggestion where to put the implementation?

test/freenet/client/filter/M3UFilterTest.java
ArneBab: This gives better output if multiple files are missing. The output would be misleading if it said "one file not found" while in reality no single file can be found.

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -984,6 +989,49 @@ public void run() {
...
+	private String getSchemeHostAndPort(ToadletContext ctx) {
+		MultiValueTable<String, String> headers = ctx.getHeaders();
+		Map<String, String> forwarded = parseForwardedHeader(headers.get("Forwarded"));
+		String uriScheme = ctx.getUri().getScheme();
+		String uriHost = ctx.getUri().getHost();
+		String protocol = forwarded.getOrDefault("proto",
+				headers.containsKey("X-Forwarded-Proto")
+						? headers.get("X-Forwarded-Proto")
+				    : uriScheme != null && !uriScheme.trim().isEmpty() ? uriScheme : "http");
+		String host = forwarded.getOrDefault("proto",
```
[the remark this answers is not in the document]
ArneBab: arg, yes … fixed.

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -984,6 +989,49 @@ public void run() {
...
+				headers.containsKey("X-Forwarded-For")
```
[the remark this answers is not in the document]
ArneBab: Thank you for checking! Fixed.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -189,9 +189,12 @@ public static FilterMIMEType getMIMEType(String mimeType) {
 	 *            Stream to write filtered data to
 	 * @param typeName
 	 *            MIME type for input data
+	 * @param hostAndPort
+	 *        HOST and PORT from the request
 	 * @param maybeCharset
 	 * 			  MIME type of the referring document, as a hint, some types,
 	 * 			  such as CSS, will inherit it if no other data is available.
+	 * @param linkFilterExceptionProvider
```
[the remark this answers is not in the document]
ArneBab: Thank you for catching! Refactoring to add and later remove sadly left the comment around.

src/freenet/client/FetchContext.java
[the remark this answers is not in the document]
ArneBab: In ClientGet this is read via the checksumReaderWithLength — if the length is the length of a single object, then this is safe.

src/freenet/client/FetchContext.java
```diff
@@ -114,24 +114,28 @@
...
 	private long cooldownTime;
 
 	/** Ignore USK DATEHINTs */
 	public boolean ignoreUSKDatehints;
-	
-	public FetchContext(long curMaxLength,
+
+  /** host and port (authority) */
+  private final String schemeHostAndPort;
+
```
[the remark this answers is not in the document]
ArneBab: dafug, comment becoming outdated; replaced with a more useful comment now.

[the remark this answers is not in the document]
Bombe: Awesome!

src/freenet/client/FetchContext.java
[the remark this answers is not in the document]
Bombe: Okay, did you check that we always use the correct length? Is it hardcoded? Does it need to be adjusted? Or is the correct length otherwise known beforehand?

src/freenet/clients/http/FProxyToadlet.java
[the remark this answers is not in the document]
Bombe: In an extra class (probably named `ForwardedHeader`, parses either a single header, maybe even multiple headers, gives you access to all values of which you can then simply use the first), which I would probaby put in… [browses code] `freenet.clients.http.utils`?

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -15,8 +16,11 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
```
Bombe, line 21: Is this import actually used? I can’t locate it anywhere in the changes for this file…

src/freenet/clients/http/FProxyToadlet.java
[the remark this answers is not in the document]
Bombe: We can also split this here…
1. ignore `Forwarded` header and only use `X-Forwarded-Proto/Host`,
1. then add `Forwarded` header later.

test/freenet/client/filter/M3UFilterTest.java
Bombe: No, we’re going down the wrong path here. Assertions and their nice messages are for the actual tests; “counted 5 but expected 8.” The test setup itself should (IMHO) throw exceptions because there’s not necessarily anything wrong with the actual code. Failed assertions are invitations for being ignored.

src/freenet/client/FetchContext.java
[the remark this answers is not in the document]
ArneBab: This is used in a PrependLengthOutputStream which skips 8 bytes, so that’s the maximum size. As far as I understood the implementation, the size is written on closing.

But it might be an old version of the datastructure, and then this would throw a EOFException. I now added handling for that (if it doesn’t have a field for schemeHostAndPort, then the field is not set).

src/freenet/clients/http/FProxyToadlet.java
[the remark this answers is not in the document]
ArneBab: I do think you’re right with that. That’s too much complexity for this pull-request.

src/freenet/clients/http/FProxyToadlet.java
ArneBab: No, it isn’t … along with some more imports. Optimized the imports.

test/freenet/client/filter/M3UFilterTest.java
ArneBab: I think giving more complete info wins here, but that opinion isn’t a particularly strongly held one :-). So it now throws directly when hitting the first missing file.

src/freenet/clients/http/FProxyToadlet.java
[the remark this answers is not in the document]
ArneBab: It now only uses the X- headers.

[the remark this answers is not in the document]
ArneBab: I moved the Forwarded-Parsing into https://github.com/freenet/fred/pull/716

src/freenet/clients/http/FProxyToadlet.java
Bombe: You can use `curl` or `wget` to create a request with the respective header fields set, a proxy is not strictly necessary. :)

test/freenet/client/filter/M3UFilterTest.java
```diff
@@ -35,16 +34,14 @@ public void testSuiteTest() throws IOException {
             try {
                 ibo = resourceToBucket(original);
             } catch (IOException e) {
-                System.out.println(original + " not found, test skipped");
-                unreadableFiles.add(original);
-                continue;
+                System.out.println(original + " not found");
```
Bombe, line 37: If you remove the whole try-catch block here you can also get rid of parts of the `resourceToBucket` method. Crash & burn! \o/

ArneBab: I want the filename in the error message — that’s the whole reason for the try-catch block.

Bombe: Why?

* If this test fails during `gradle test` with an exception you run it exactly once in the IDE to find out why it’s failing. The filename in the exception will not yield any meaningful information (in the sense that it saves you a lot of time).
* You could add the filename to the exception you create in the `resourceToBucket` method, this way it will show up without the try-catch block.

src/freenet/clients/http/FProxyToadlet.java
Bombe: Heh, while it is totally safer to only allow known values for those headers it will break my setup (because Fred doesn’t know anything about being reverse-proxied). That can probably be added in a future PR with a config option.

However, I still would like to see this extracted into an extra component and properly unit-tested.

ArneBab: For your setup you’ll need to add the proxy-url to the bindTo addresses. Then fred will allow it. This is a stop-gap measure, but it should work.

Bombe: That’s… not how reverse proxies work, and I won’t bind Fred to an external IP address. Anyway, later PR. :)

test/freenet/client/filter/M3UFilterTest.java
ArneBab: you’re right — resourceToBucket now preserves the filename and this makes the code much nicer. Thank you!

src/freenet/clients/http/FProxyToadlet.java
ArneBab: sounds good — I extracted the code now, added some tests and found that default values for configured IP and Port were missing (that case should not be possible, but "should not").

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
```diff
@@ -0,0 +1,72 @@
...
+            : fProxyPortConfig;
+        // allow all bindToHosts
+        Set<String> safeHosts = new HashSet<>(bindToHosts);
+        // also allow bindTo hosts with the fProxyPortConfig added
+        safeHosts.addAll(safeHosts.stream()
+                         .map(host -> host + ":" + port)
+                         .collect(Collectors.toList()));
+
+        // check uri host and headers
+        // TODO: parse the Forwarded header, too. Skipped here to reduce the scope.
```
Bombe, line 56: Instead of TODOs littered all over the code I’d prefer that a ticket were created, either in Github or in Mantis. :)

src/freenet/pluginmanager/OfficialPlugins.java
```diff
@@ -123,9 +123,9 @@ public OfficialPlugins() {
 					.loadedFrom("USK@QeTBVWTwBldfI-lrF~xf0nqFVDdQoSUghT~PvhyJ1NE,OjEywGD063La2H-IihD7iYtZm3rC0BP6UTvvwyF5Zh4,AQACAAE/WebOfTrustTesting.jar/17");
 			addPlugin("FlogHelper")
 					.inGroup("communication")
-					.minimumVersion(35)
+					.minimumVersion(36)
 					.usesXml()
-					.loadedFrom("CHK@3Ht6otYHogKxa1Z8SC0hz46kV9q1qaSBsjbJxuwPosU,txky3jF33oWU9n6YiGMphiHA1V9q3i4ZN0VPtgElSaI,AAMC--8/FlogHelper-v35.jar");
+					.loadedFrom("CHK@vk9NRGTCBqOzQs2S~PhlJCAf3uYmoMguNB5a5fx3wC4,bx8ewRk-OJPwXOj5P1PXUxuuK7ZxuNLHXsMJJH9xuVg,AAMC--8/FlogHelper.jar");
```
Bombe, line 128: Can you please watch out that you’re not mixing in unrelated changes into your PRs all the time? I can help you to move unrelated commits out of the PR branches if you like because this makes it really hard to okay them…

src/freenet/support/MultiValueTable.java
```diff
@@ -1,13 +1,15 @@
 package freenet.support;
+import java.text.MessageFormat;
```
Bombe, line 2: Unused imports, both of them. :)

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -340,7 +340,8 @@ public static String HTMLFilter(String data, boolean alt) throws Exception {
 		ArrayBucket output = new ArrayBucket();
 		InputStream inputStream = input.getInputStream();
 		OutputStream outputStream = output.getOutputStream();
-		ContentFilter.filter(inputStream, outputStream, typeName, baseURI, null, null, null);
+		ContentFilter.filter(inputStream, outputStream, typeName, baseURI,
```
Bombe, line 343: Unnecessary line break.

test/freenet/client/filter/PNGFilterTest.java
```diff
@@ -112,7 +112,9 @@ public void testSuiteTest() throws IOException {
 			}
 
 			try {
-				filter.readFilter(ib.getInputStream(), new NullBucket().getOutputStream(), "", null, null);
+				filter.readFilter(ib.getInputStream(), new NullBucket().getOutputStream(), "", null,
+            null,
```
Bombe, line 116: Unnecessary line breaks.

test/freenet/clients/http/utils/UriFilterProxyHeaderParserTest.java
```diff
@@ -0,0 +1,95 @@
...
+  private final String fProxyBindTo;
+  private final String uriScheme;
+  private final String uriHost;
+  private final MultiValueTable<String, String> headers;
+  private String resultUriPrefix;
+
+  @Parameterized.Parameters(
+      name = "{index}: portConfig=\"{0}\", bindTo=\"{1}\", uriScheme=\"{2}\", uriHost=\"{3}\", headers={4}, expected=\"{5}\"")
+  public static Collection constructorParametersAndResult () {
+    return Arrays.asList(new Object[][] {
```
Bombe, line 27: I am not a big fan of these parameterized tests because if test 17 of those fails I have _no_ idea whatsover how that test differs from the other ones, i.e. what makes _this_ test so special that it fails… I’d prefer one method per actual test, e.g. “testThatInvalidHostInHeaderIsReplacedByFproxyHost” and similarly-named tests.

Bombe (changes requested):

src/freenet/client/filter/ContentFilter.java
```diff
@@ -275,9 +294,9 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
 				if(before.equals("charset")) {
 					charset = after;
 				} else {
-					if (otherParams == null)
-						otherParams = new LinkedHashMap<String, String>();
-					otherParams.put(before, after);
+					if (otherMimeTypeParams == null)
+						otherMimeTypeParams = new LinkedHashMap<String, String>();
```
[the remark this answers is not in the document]
ArneBab: good point

src/freenet/clients/http/staticfiles/js/checkall.js
```diff
@@ -1,9 +1,13 @@
-// copypaste javascript from http://stackoverflow.com/a/7251104/7666
+// @license magnet:?xt=urn:btih:1f739d935676111cfff4b4693e3816e664797050&dn=gpl-3.0.txt GPL-v3-or-Later
 function checkAll(bx, classname) {
-  var cbs = document.getElementsByClassName(classname)[0].getElementsByTagName("input");
-  for(var i=0; i < cbs.length; i++) {
-    if(cbs[i].type == "checkbox") {
-      cbs[i].checked = bx.checked;
+  var matching = document.getElementsByClassName(classname);
+  for (var matchingElement of matching) {
```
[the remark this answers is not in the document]
ArneBab: Yes, the changed behavior now checks all elements with the given class, not just the first. For multiple tables on the same page, that can check all checkboxes of all tables, but since the submit button only takes the correct form into account, this leads to correct behavior.

The previous behavior made the checkall box ineffective for the second table on a page.

src/freenet/clients/http/staticfiles/js/progresspage.js
```diff
@@ -1,3 +1,4 @@
+// @license magnet:?xt=urn:btih:1f739d935676111cfff4b4693e3816e664797050&dn=gpl-3.0.txt GPL-v3-or-Later
```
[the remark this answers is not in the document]
ArneBab: That is allowed, yes (fred is GPLv2 or later, so it can be shipped as GPLv3 or later), but even if allowed, it *should* still be GPLv2 or later here.

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
```diff
@@ -0,0 +1,72 @@
...
+        this.uriScheme = uriScheme;
+        this.uriHost = uriHost;
+        this.headers = headers;
+    }
+
+    public String getSchemeHostAndPort() {
+        Set<String> safeProtocols = new HashSet<>(Arrays.asList("http", "https"));
+
+        List<String> bindToHosts = Arrays.asList(
+                                                 fProxyBindToConfig.split(",")).stream()
```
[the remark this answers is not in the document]
ArneBab: yikes, yes :-)

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
ArneBab: I now added a ticket: https://bugs.freenetproject.org/view.php?id=7138

test/freenet/clients/http/utils/UriFilterProxyHeaderParserTest.java
ArneBab: In this case (due to name = "{index}: …"), you get information how the data in the test changes. But I can change this with a moderate amount of work. Watch me crank out Emacs macros :-)

src/freenet/clients/http/staticfiles/js/progresspage.js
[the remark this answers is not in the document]
Bombe: Aren’t there parts that are GPLv2 (without “later”)? I think I can remember some arguments about that many, many moons ago…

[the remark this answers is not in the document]
ArneBab: I dearly hope that not — if there are, the change I did should remove any problems (gplv2+).

src/freenet/pluginmanager/OfficialPlugins.java
ArneBab: sorry again — this should not have gone in here :-(

src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -984,6 +987,23 @@ public void run() {
 		}
 	}
 
+	private String getSchemeHostAndPort(ToadletContext ctx) {
+		// retrieve config from froxy
+		SubConfig fProxyConfig = core.node.config.get("fproxy");
+
+		String fProxyPort = String.valueOf(fProxyConfig.getInt("port"));
+		String fProxyBindTo = fProxyConfig.getString("bindTo");
```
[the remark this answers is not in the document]
ArneBab: That’s much nicer — thank you!

src/freenet/client/filter/M3UFilter.java
```diff
@@ -116,65 +117,63 @@ public void readFilter(InputStream input, OutputStream output, String charset, H
...
-                                if (cb instanceof GenericReadFilterCallback) {
-                                    try {
-                                        filtered = ((GenericReadFilterCallback)cb).makeURIAbsolute(filtered);
-                                    } catch (URISyntaxException e) {
-                                        filtered = badUriReplacement;
+                                    // mirroring tools like `wget -mk`
+                                    // strip the absolute path again,
+                                    // so mirroring should not be
+                                    // impaired.
+                                    filtered = cb.processURI(uriold, subMimetype,
```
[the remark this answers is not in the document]
ArneBab: you’re right — I totally blame automatism, and me not checking it hard enough. Thank you for catching it!

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import freenet.support.MultiValueTable;
+
+public class UriFilterProxyHeaderParser {
```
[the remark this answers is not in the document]
ArneBab: I moved usages to a static parse-method, returning a SchemeAndHostWithPort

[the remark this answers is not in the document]
Bombe: That’s a lot more complicated than what I had in mind and still doesn’t address most of the points.

Move the parsing code *into* the parse method. Remove the member fields and the constructor. You don’t really need much more than the parse method and the container for the returned values, everything else distracts from what this does.

We’re getting close! :)

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
```diff
@@ -0,0 +1,99 @@
...
+
+        SchemeAndHostWithPort(
+            String scheme,
+            String host
+        ) {
+            this.scheme = scheme;
+            this.host = host;
+        }
+
+        public String toString() {
```
Bombe, line 57: Hmm. Yes, this does what’s advertised but I’m hesitant to have “real” business logic in a method that was originally meant for debugging. However, feel free to let it this way, it doesn’t feel important enough. :)

test/freenet/clients/http/utils/UriFilterProxyHeaderParserTest.java
```diff
@@ -0,0 +1,342 @@
...
+        0,
+        false,
+        false,
+        "",
+        "",
+        new DummyStringCallback());
+    try {
+      option.setValue(value);
+    } catch (InvalidConfigValueException e) {
+      e.printStackTrace();
```
Bombe, line 301: No, no, no. This is auto-generated try-catch block, shame on you! :) Please let the exception bubble up and fail the test.

Bombe (changes requested):
We’re nearing the end of it! :)

src/freenet/pluginmanager/OfficialPlugins.java
ArneBab: I got it moved into another branch … worked this time.

test/freenet/clients/http/utils/UriFilterProxyHeaderParserTest.java
ArneBab: *cowers in shame* … :-)

After doing that I remembered why I did not: Every test method now throws an Exception. But it’s more robust, so it did get better.

src/freenet/clients/http/utils/UriFilterProxyHeaderParser.java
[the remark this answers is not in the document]
Bombe: \o/

Bombe (approved):
That looks good! Is there already a ticket for the next changes in the header-parsing code? As soon as there is, this is good to go, IMHO.

ArneBab:
There’s no ticket for the next changes, but a pull-request: https://github.com/freenet/fred/pull/716

ArneBab:
Thank you for helping me make this work! You’re awesome!

Bombe:
> There’s no ticket for the next changes, but a pull-request: #716

Even better! :)

PR #692, opened 2020-02-09
increase max size for passthrough, because our windows installer is >…

ArneBab:
…70 MiB and this seems to limit UOM file transfer

### DIFF
--- a/src/freenet/clients/http/FProxyToadlet.java
+++ b/src/freenet/clients/http/FProxyToadlet.java
@@ -88,7 +88,7 @@ public final class FProxyToadlet extends Toadlet implements RequestClient {
 	// ?force= links become invalid after 2 hours.
 	private static final long FORCE_GRAIN_INTERVAL = HOURS.toMillis(1);
 	/** Maximum size for transparent pass-through. See config passthroughMaxSizeProgress */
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -88,7 +88,7 @@
 	// ?force= links become invalid after 2 hours.
 	private static final long FORCE_GRAIN_INTERVAL = HOURS.toMillis(1);
 	/** Maximum size for transparent pass-through. See config passthroughMaxSizeProgress */
-	public static long MAX_LENGTH_WITH_PROGRESS = (50*1024*1024 * 11) / 10; // 50MiB plus a bit due to buggy inserts
+	public static long MAX_LENGTH_WITH_PROGRESS = (100*1024*1024 * 11) / 10; // 100MiB plus a bit due to buggy inserts, because our Windows installer is >70 MiB nowadays
```
Bombe: The comment “plus a bit” in combination with the way the parens are set makes this line a bit harder to digest than necessary. `(100*1024*1024) * 11 / 10` should be a bit better to read.

ArneBab: adjusted, thank you!

PR #693, opened 2020-05-28
Bugfix/0007132 freenet uri accepts invalid usk keys

Bombe:
https://freenet.mantishub.io/view.php?id=7132

USKs like “USK@/broken/0” are accepted by FreenetURI’s constructor but shouldn’t.


ArneBab:
looks good — thank you!

PR #698, opened 2020-07-26
Fix peer options box

Bombe:
The incorrect l10n prefix was used to render the boxes, and it’s not necessary anyway, as trust and visibility only ever apply to darknet peers, anyway, so the prefix was removed entirely.


ArneBab:
Looks good! Thank you!

PR #699, opened 2020-07-27
Hand in request client directly

Bombe:
As the `ClientCallback`’s only used functionality is to return a `RequestClient`, the client can simply be handed in instead of the callback, removing the need for fake callbacks.


ArneBab:
That’s much nicer! Thank you!

PR #701, opened 2020-08-05
Set context class loader when terminating plugin

Bombe:
This will fix a number of weird “can not load class/file”
exceptions on shutdown in plugins.

### DIFF
--- a/src/freenet/pluginmanager/PluginInfoWrapper.java
+++ b/src/freenet/pluginmanager/PluginInfoWrapper.java
@@ -142,12 +142,19 @@ public class PluginInfoWrapper implements Comparable<PluginInfoWrapper> {
 		unregister(manager, reloading);
 		// TODO add a timeout for plug.terminate() too
 		System.out.println("Terminating plugin "+this.getFilename());
+
+		// set the plugin’s class loader as context class loader
+		ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
+		Thread.currentThread().setContextClassLoader(plug.getClass().getClassLoader());
+
 		try {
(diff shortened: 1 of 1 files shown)


desyncr:
LGTM 👍

src/freenet/pluginmanager/PluginInfoWrapper.java
```diff
@@ -142,9 +142,15 @@ public void startShutdownPlugin(PluginManager manager, boolean reloading) {
...
 		System.out.println("Terminating plugin "+this.getFilename());
+
+		// set the plugin’s class loader as context class loader
+		ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
+		Thread.currentThread().setContextClassLoader(plug.getClass().getClassLoader());
+
 		try {
 			plug.terminate();
 		} catch (Throwable t) {
+			Thread.currentThread().setContextClassLoader(originalClassLoader);
```
desyncr, line 153: Wouldn't need to restore the original class loader on a successful termination? (maybe add a `finally` block?)

Bombe: That’s debatable… :) 
* This thread will also be used to terminate all other plugins so the context class loader will be overwritten again, anyway.
* In theory, most things shouldn’t even use the class loader.

But yeah, we should be consistent here, either reset it always or never but only in case of an exception. :)

desyncr:
@ArneBab Could you take a look? I think it's ready to merge.

ArneBab:
It looks good to me, too. I’ll merge it ASAP.

PR #702, opened 2020-08-07
Adjust version check: Android is not EOL

ArneBab:


desyncr:
LGTM 👍

src/freenet/support/JVMVersion.java
```diff
@@ -34,7 +36,12 @@
 	    Pattern.compile("^0*(\\d+)(?:\\.0*(\\d+)(?:\\.0*(\\d+)(?:[_.]0*(\\d+))?)?)?.*$");
 
 	public static boolean isEOL() {
-		return isEOL(getCurrent());
+		return !isAndroid() // on android the version checks are done on the App level, so we do not check here.
+			&& isEOL(getCurrent());
+	}
+
+	public static boolean isAndroid() {
```
Bombe, line 43: Is there any reason this is `public`?

desyncr: I believe it's because isEOL is public static as well.

desyncr: May be it could be removed and in-lining Platform.isAndroid()

Bombe: I’d be content with the method being `private`… but you’re right, with a static import we could totally skip adding this method.

src/freenet/support/JVMVersion.java
```diff
@@ -3,6 +3,8 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import com.sun.jna.Platform;
```
desyncr, line 6:
```suggestion
import static com.sun.jna.Platform.isAndroid;
```

src/freenet/support/JVMVersion.java
```diff
@@ -34,7 +36,12 @@
...
 
 	public static boolean isEOL() {
-		return isEOL(getCurrent());
+		return !isAndroid() // on android the version checks are done on the App level, so we do not check here.
+			&& isEOL(getCurrent());
+	}
+
+	public static boolean isAndroid() {
+		return Platform.isAndroid();
 	}
```
desyncr, line 45:
```suggestion
```

src/freenet/support/JVMVersion.java
ArneBab: There is no good reason this is public, no. The real reason is that extract-method made it public by default and I didn’t correct that.

The reason why I used a method is to preserve the structure of the isEOL method, since all other calls used in isEOL() are methods, too.

src/freenet/support/JVMVersion.java
ArneBab: I’m using the static import now.

PR #704, opened 2020-08-12
Removed unused imports

JanGerritsen:
Removed various unused imports.
Why: Better maintainability through the removal of false dependencies.

_Im am digging through fred's code to get familiar with it and noticed a lot of unused imports. 
I thought removing them might make a nice small PR to exercise building and testing fred and try out the development cycle._


Bombe:
Imports are not really dependencies, and unused imports are not false dependencies. However, cleaning that up a bit can’t hurt.

PR #706, opened 2020-08-13
Removed several usages of the deprecated Closer

JanGerritsen:
Closer is deprecated and can often be replaced by try with closeable statements.
In this PR several such places have been refactored. 
A few cases also fix potential leaks where resources are not closed probably in all cases.

### DIFF
--- a/src/freenet/client/Metadata.java
+++ b/src/freenet/client/Metadata.java
@@ -37,7 +37,6 @@ import freenet.support.api.Bucket;
 import freenet.support.api.BucketFactory;
 import freenet.support.api.RandomAccessBucket;
 import freenet.support.compress.Compressor.COMPRESSOR_TYPE;
-import freenet.support.io.Closer;
 import freenet.support.io.CountedOutputStream;
 import freenet.support.io.NullOutputStream;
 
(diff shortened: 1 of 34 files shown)


desyncr:
@JanGerritsen Thanks for the work! It could take a while to review it and be sure we don't break anything :)

I haven't been able to carefully review the code so far, but there's a mix of tabs and spaces causing some indentation issues.

JanGerritsen:
@desyncr: I am Sorry, but the files use mixed tabs and spaces and I opted to use tabs, because that's what seems to be used mostly.

Has there been a try to define a standard code style for fred?

desyncr:
I think the default standard is tabs. [Here](https://github.com/freenet/fred/pull/706/files#diff-064ded8a866a9193072429b727bed45bR1230)'s a mix of tabs and spaces.

JanGerritsen:
Yes I know. I tried to preserve Identation as much as I could, to make the diff not bigger than it needs to be, but well, the whole file already uses a mix.

I would suggest that there should be a defined format, and then to create one merge request, were someone just does the reformat of the project.
Afterwards it would be much easier to create pull requests, which show what has actually changed, not just reformatted.

But this would need to be a coordinated effort and one needs to a agree on a standard formater...

desyncr:
Sure, one or multiple PRs just fixing standard issues.

ArneBab:
Looks good and the tests pass. Merged.

PR #707, opened 2020-08-13
Fixed several unused variables issues

JanGerritsen:
Removed several unused local variables


JanGerritsen:
Might there be a problem in the CI Build? All Java Versions except Java 9 ran successfully, and the test for Java 9 just has stopped.
Is there a way to retrigger the Travis CI build?

desyncr:
@JanGerritsen Hey! Thanks for the work! Regarding the build failure I think you can force a re-build by pushing an empty commit: https://stackoverflow.com/a/31694534

Otherwise someone with Write access to the repo could trigger the build using Travis UI, @ArneBab ?

src/freenet/node/PeerManager.java
```diff
@@ -1793,7 +1791,7 @@ public void maybeLogPeerNodeStatusSummary(long now) {
...
+			Logger.normal(this, "Connected: " + numberOfConnected + "  Routing Backed Off: " + numberOfRoutingBackedOff + "  Too New: " + numberOfTooNew + "  Too Old: " + numberOfTooOld + "  Disconnected: " + numberOfDisconnected + "  Never Connected: " + numberOfNeverConnected + "  Disabled: " + numberOfDisabled + "  Bursting: " + numberOfBursting + "  Listening: " + numberOfListening + "  Listen Only: " + numberOfListenOnly + "  Clock Problem: " + numberOfClockProblem + "  Connection Problem: " + numberOfConnError + "  Disconnecting: " + numberOfDisconnecting + " Disconnecting " + numberOfRoutingDisabled + " No load stats: "+numberOfNoLoadStats);
```
Bombe, line 1794: The second `Disconnecting` seems to be mislabeled.

Bombe (changes requested):

src/freenet/node/PeerManager.java
JanGerritsen: Yes, you are right. Fixed it.

ArneBab:
@Bombe is this good to go?

Bombe:
As soon as somebody confirms that all removed method calls do not have any side effects, sure. :)

desyncr:
@Bombe Do you mean these ones:

- https://github.com/freenet/fred/pull/707/files#diff-5ee75719a7e50239b28aa9afc54916bccf13373d2b659220627834d07911dfaaL103
- https://github.com/freenet/fred/pull/707/files#diff-65af551d910019a886b79d1a718d2086f20f69c9b63a4290a9fc4b74f7472af5L2814

I believe the rest are unused variables being removed.

ArneBab:
Does that mean it’s good to go?

desyncr:
@ArneBab If the problem is with the removed methods specifically then we can leave them until we're certain there are no side effect, while removing the unused variables alone. Does that sound good to you @Bombe @JanGerritsen ?

Bombe:
I checked the methods and them seem to be free enough of side-effects to simply remove them.

desyncr:
Right on! Anything else left to do here?

desyncr:
@Bombe Are we good to go?

desyncr (approved):

Bombe:
Sure.

desyncr:
@Bombe Great! I guess we'll need an approval from you @Bombe before it can be merged.

Bombe (approved):

Bombe:
Oops, sorry for the delay, didn’t realize I had to push the button. 😄

desyncr:
@ArneBab Should we be ready to go!

ArneBab:
Thank you for your changes and thank you for the review!

PR #708, opened 2020-08-14
Added definition that source code is encoded in UTF-8  to gradl build definition

JanGerritsen:
The source code is encoded in utf-8, which is currently not specified.

This causes several issues:

- The following tests fail (at least under windows)
  - Base64Test.testDecodeStandard 
  - Base64Test.testEncodeStandard  

- The compiler generates several parsing errors during the compilation:
  - error: unmappable character (0x9D) for encoding windows-1252
  -...

Defining the source code encoding fixes these issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,14 @@ task buildInfo {
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
+tasks.withType(JavaCompile) {
+    options.encoding = "UTF-8"
+}
+
+tasks.withType(Javadoc) {
+    options.encoding = "UTF-8"
+}
+
 task compileVersion (type: JavaCompile) {
     copy {
         from sourceSets.main.java.srcDirs


desyncr:
Great find! Thank you!

build.gradle
```diff
@@ -69,6 +69,8 @@ task compileVersion (type: JavaCompile) {
     }
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
+    compileJava.options.encoding = "UTF-8"
```
Bombe, line 72: I’d prefer this to be more global than just for this task, i.e.
```
tasks.withType(JavaCompile) {
  options.encoding = "UTF-8"
}
```
at top-level somewhere.

JanGerritsen: I moved the definition to the same level where the sourceCompatibility level is defined globally, to keep them the same.

Bombe: Can you please use the way I suggested? This way it will always be set correctly for all java compilation tasks.

JanGerritsen: tasks.withType(JavaCompile) {
    options.encoding = "UTF-8"
}

tasks.withType(Javadoc) {
    options.encoding = "UTF-8"
}

Bombe: I personally wouldn’t have bothered with the javadoc but that’s fine. :)

PR #709, opened 2020-08-14
Replaces deprecated Constructors like Byte() or Integer()  with Byte.…

JanGerritsen:
…valueOf() or autoboxing

As of Java 9 the constructors to wrap a native type within the corresponding Object Class have been deprecated.
Replaces those calls with the long existing valueOf static methods or through autoboxing.


ArneBab:
looks good. Merged.

PR #710, opened 2020-08-14
Removed deprecated and now unused SortedIntSet

JanGerritsen:
The SortedIntSet had been deprecated and is not used anymore. So it was removed.


ArneBab:
Looks like Library currently uses SortedIntSet

PR #712, opened 2020-08-16
Fixed unreliable store store test cases

JanGerritsen:
Several store tests fail when run (on windows system).

**Update:**
Now each test is executed in its own directory, so tests should be able to be executed in parallel without conflicting.

There were several problems.
**Stores are not closed properly in tests** 
One problem was that in many places the stores were not closed correctly at the end of each test case, which kept their file handlers open.
On windows machines you cannot move or delete a file which has an open file handle pointing to it.

The file pointers in the stores were often only closed after they were garbage collected. Since the garbage collection run is unpredictable, the success of the test was unpredictable.

You could see this in the messages that the system complained that it could not move or delete certain files (because they were still open by stores of previous tests)

**SaltedHashFreenetStore can lose entries through collisions**
Another problem was that inserting keys into a SaltedHashFreenetStore can create a collision, so it might replace an existing key,
but the test cases relied on the fact that all keys are present. (Maybe an artifact from the old LRU Store?)

**Changes:**
* Changed FreenetStore to implement closable, which allows to use try-with-resources syntax, to make sure all stores are closed probably at the end of a block.
* SaltedHashFreenetStore and CachingFreenetStore also double check in their close() method, that they do nothing if called twice, to prevent it from writing their data again on a consecutive call. (I used atomic boolean to make the check thread save, don't know if it is required)
* Tests now use try-with-recource syntax to create stores, to make sure that the store is closed at the end of the test. Independent of, if the test was successful or not.
* Also added a check to SaltedHashFreenetStore.java.openStoreFiles() , to check if a lock could be acquired and generate a exception with a helpful message if not.
* The test cases have been refactored heavily, to take into account that a key can be lost.

### DIFF
--- a/src/freenet/store/caching/CachingFreenetStore.java
+++ b/src/freenet/store/caching/CachingFreenetStore.java
@@ -33,6 +33,7 @@ public class CachingFreenetStore<T extends StorableBlock> extends ProxyFreenetSt
 	 * True if close() has been called
 	 */
 	private AtomicBoolean closeCalled = new AtomicBoolean(false);
+
 	private final LRUMap<ByteArrayWrapper, Block<T>> blocksByRoutingKey;
 	private final StoreCallback<T> callback;
 	private final boolean collisionPossible;
--- a/test/freenet/store/RAMSaltMigrationTest.java
+++ b/test/freenet/store/RAMSaltMigrationTest.java
@@ -1,13 +1,19 @@
 package freenet.store;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
-import java.util.UUID;
 
-import junit.framework.TestCase;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
 
 import freenet.crypt.DummyRandomSource;
 import freenet.crypt.RandomSource;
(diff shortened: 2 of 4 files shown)


nextgens:
> The test cases have been refactored heavily, to take into account that a key can be lost.

Hmm, this should happen deterministically (and on all platforms) if that was the case. We should ensure we pick a deterministic PRNG with a fixed seed for testing (iirc we already do)... and pick the seed in a way where the is no collision (and test separately the case of collision, by creating one on purpose).

JanGerritsen:
> > The test cases have been refactored heavily, to take into account that a key can be lost.
> 
> Hmm, this should happen deterministically (and on all platforms) if that was the case. We should ensure we pick a deterministic PRNG with a fixed seed for testing (iirc we already do)... and pick the seed in a way where the is no collision (and test separately the case of collision, by creating one on purpose).

Yes, that's what I thought. but I could not find where. The Test-Case instantiates a a random number generator with a fixed seed.
But I am not too familiar with the code yet, there is some magic with global static constants set at random places, so maybe I just missed it.

Instead I concentrated to test what we can guarantee, how the Store really behaves in the wild with its randomness, since the store also does not have a fixed random initialisier in real use cases.

But if you have a good idea to improve the test, I will happily have a look how to implement it.

nextgens:
We're chasing what is presumably a race condition and should really use different files/stores for each test... please replace the various new File(tempDir, "saltstore") by a call to Files.createTempFile()...

And try setting something like this in your gradle.properties:
org.gradle.parallel = true
tasks.withType(Test)  {
  maxParallelForks = Runtime.runtime.availableProcessors()
}

Odds are the various CI environments are doing it by default

JanGerritsen:
> We're chasing what is presumably a race condition and should really use different files/stores for each test... please replace the various new File(tempDir, "saltstore") by a call to Files.createTempFile()...
> 
> And try setting something like this in your gradle.properties:
> org.gradle.parallel = true
> tasks.withType(Test) {
> maxParallelForks = Runtime.runtime.availableProcessors()
> }
> 
> Odds are the various CI environments are doing it by default

I implemented 
```java
private File getTempStoreFile() {
	return new File(tempDir, "saltstore_" + UUID.randomUUID());
}
```
for __test/freenet/store/RAMSaltMigrationTest.java__ to make sure tests use a unique storage.

I could extend this for all test cases, that it works on the folder, so parallel deletes would also not get into conflicts. (Each test case currently cleans its whole working directory at the end of the test.
Would this help?

_I implemented while I was debugging the mysterious failing tests, when I still thought the error were conflicting tests. But I didn't run them in parallel, only consecutively_

nextgens:
Use the testname from junit rather than a random GUID please... that should make it much easier when we will dig through it

ArneBab:
@nextgens is this good to go?

ArneBab:
Is this good to go?

nextgens:
Yes I think so

PR #717, opened 2020-09-07
Reduce max backoff times.

TheSeeker:
The only part of the 'performance' patches given any sort of go-ahead. forgotten about years ago.


ArneBab:
I think we should risk this; it could reduce the fraction of backed off connections which would increase the average bandwidth available to Freenet as network.

ArneBab:
merged — thank you!

PR #719, opened 2020-11-29
provide default implementation for the previous readFilter interface

ArneBab:
This keeps compatibility with plugins using the readFilter.

### DIFF
--- a/src/freenet/client/filter/ContentDataFilter.java
+++ b/src/freenet/client/filter/ContentDataFilter.java
@@ -7,6 +7,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Data filter for a specific MIME type.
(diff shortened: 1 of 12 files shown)


src/freenet/client/filter/ContentDataFilter.java
```diff
@@ -35,7 +35,17 @@
...
 	 */
-	public void readFilter(
+	void readFilter(
 			InputStream input, OutputStream output, String charset, HashMap<String, String> otherParams,
 			String schemeHostAndPort, FilterCallback cb) throws DataFilterException, IOException;
+
+	/**
+	 * Compatibility for readFilter without schemeHostAndPort. Please use readFilter with schemeHostAndPort.
+	 */
+	default void readFilter(
```
Bombe, line 45: Please add `@Deprecated` to this method so the compiler can tell people not to use this.

src/freenet/client/filter/ContentDataFilter.java
```diff
@@ -35,7 +35,17 @@
...
+			InputStream input, OutputStream output, String charset, HashMap<String, String> otherParams,
```
Bombe, line 46: Can you use the less restrictive `Map<String, String>` as parameter type here? It shouldn’t break existing code and is clearer to read.

Bombe (changes requested):

src/freenet/client/filter/ContentDataFilter.java
ArneBab: that sounds good, yes. And thankfully it’s just a “migrate”-option away to move the implementing classes, too.

src/freenet/client/filter/ContentDataFilter.java
ArneBab: done :-)

ArneBab:
I asked xor to test the change against WoT — github linked that.

ArneBab:
@Bombe is it good to go?

ArneBab:
WoT is now shown to work.

Bombe:
> @Bombe is it good to go?

Looks good!

PR #723, opened 2021-03-13
triple max size for update file transfers to avoid errors due to the …

ArneBab:
…larger windows installer


desyncr (approved):
LGTM 👍 Only thing I'd consider is using constants for seedrefs and installers.

Similar idea to the MAX_MAIN_JAR_LENGTH constant but for these:

- MAX_SEEDREF_LENGTH
- MAX_INSTALLER_LEGTH
- MAX_IPV4_TO_COUNTRY_LENGTH

ArneBab:
Ah, yes — I’ll do that. Thank you!

PR #726, opened 2021-04-03
start pitch black defense 10 minutes after startup, not 1 minute afte…

ArneBab:
…r startup


desyncr (approved):

ArneBab:
Thank you for the review!

PR #727, opened 2021-04-05
Add Diagnostics component

desyncr:
This PR updates the /diagnostics section (DiagnosticToadlet) to display thread information, including CPU usage, to the user. The intention is to show information to determine the cause of CPU usage (ex. plugins, Log, UPnP etc).

```
Threads (155):
 Thread ID          Job ID Name                                                                                       Prio.      Group Status               % CPU
       482      -600434685 NodeDiagnostics: thread monitor(61)                                                            5       main RUNNABLE             16.20
       337     -1605458390 RequestSender for UID -4819448343849908691 on 29966(41)                                        7       main TIMED_WAITING        4.17
       293       820065101 RequestSender for UID -3007138528378398579 on 29966(14)                                        7       main TIMED_WAITING        3.20
       510       283081595 RequestSender for UID 611930334131416017 on 29966(122)                                         7       main TIMED_WAITING        1.42
       613     -1700317102 RequestSender for UID -3867706999542896026 on 29966(175)                                       7       main TIMED_WAITING        1.29
       608      -101505222 RequestSender for UID 3554502223033864374 on 29966(173)                                        7       main TIMED_WAITING        1.27
        28              28 PacketSender thread for 29966                                                                 10       main TIMED_WAITING        1.27
       343      -486051315 RequestSender for UID 7371023460396097846 on 29966(48)                                         7       main TIMED_WAITING        1.15
        80      -971461974 UdpSocketHandler for port 30081(1)                                                            10       main RUNNABLE             0.84
       647     -1318572914 RequestSender for UID -7623304779548324689 on 29966(208)                                       7       main TIMED_WAITING        0.73
        78     -1374273574 <noname>(1)                                                                                    6       main WAITING              0.28
        29              29 Ticker thread for 29966                                                                       10       main TIMED_WAITING        0.10
       550      2024818125 Slow callback for freenet.io.xfer.BlockReceiver$3@49a11870(83)                                 5       main WAITING              0.04
        15              15 Wrapper-Control-Event-Monitor                                                                  5       main TIMED_WAITING        0.03
        24      1587637561 Network Interface Acceptor for ServerSocket[addr=/127.0.0.1,localport=8888](0)                 5       main RUNNABLE             0.02
        40      -960067964 SSK Request starter (29966) (bulk)(1)                                                          5       main WAITING              0.02
       770      1988340316 CHKInsertSender for UID -899544705615949250 on 29966 at 1628945874470(290)                     7       main TIMED_WAITING        0.01
        25              25 SimpleToadletServer                                                                            5       main WAITING              0.01
         3               3 Finalizer                                                                                      8     system WAITING              0.01
```


ArneBab:
the code looks good. Can you check whether there is something sensitive in the thread name when you insert a file? (any info about the target key)

can you check that you don’t change the indentation? The current format isn’t ideal, but better strange but consistent indentation than inconsistent indentation.

desyncr:
Some more comments over IRC:
> we usually attach instrumentation to the JVM's socket.. jconsole, yourkit, eclipse. That being said, #727 looks cool. You may want to make the sleep() in the middle configurable (profiling over 1s may not always make sense) using a get parameter or something...

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -409,15 +414,60 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
+			Map<Long, Long> threadCPU = new HashMap<Long, Long>();
+			for (ThreadInfo info : threadMxBean.dumpAllThreads(false, false)) {
+			    threadMxBean.getThreadInfo(info.getThreadId());
+				threadCPU.put(info.getThreadId(), threadMxBean.getThreadCpuTime(info.getThreadId()));
+			}
+
+			RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
+			long initialUptime = runtimeMxBean.getUptime();
+
+			try {Thread.sleep(1000);} catch (InterruptedException e) {}
```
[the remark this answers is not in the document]
Bombe: At least put this in its own method and format it correctly…

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -409,15 +414,60 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
+
+			for (ThreadInfo info : threadMxBean.dumpAllThreads(false, false)) {
+				Long prev = threadCPU.get(info.getThreadId());
+				if (prev == null)
+					continue;
+				threadCPU.put(info.getThreadId(), threadMxBean.getThreadCpuTime(info.getThreadId()) - prev);
+			}
+
+			Float totalCPU = (runtimeMxBean.getUptime() - initialUptime) * 1000000F * 1;
```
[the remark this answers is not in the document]
Bombe: Oh, and… `* 1`? What? :)

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -409,15 +414,60 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
+					"Status",
+					"% CPU"
+				)
+			);
+
+			for (Thread t : stats.getThreads()) {
+				if (t == null)
+					continue;
+
+				Float cpuUsage = threadCPU.getOrDefault(t.getId(), 0l) / totalCPU;
```
Bombe, line 457: Same as above, no reason to box this to a `Float` (and again no reason for `float` at all).

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
desyncr: Alternatively I was thinking launching a thread to keep this metrics up to date and not having to wait for this on every request. Definitely needs some more work.

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -409,15 +414,60 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
-			textBuilder.append(l10n("running")).append(": ").append(String.valueOf(activeThreadsByPriority[i])).append(" (").append(String.valueOf(i+1)).append(")\n");
-			textBuilder.append(l10n("waiting")).append(": ").append(String.valueOf(waitingThreadsByPriority[i])).append(" (").append(String.valueOf(i+1)).append(")\n");
-		}
-		textBuilder.append("\n");
+			// drawThreadPriorityStatsBox
+			textBuilder.append("Threads:\n");
+			ThreadMXBean threadMxBean = ManagementFactory.getThreadMXBean();
+			Map<Long, Long> threadCPU = new HashMap<Long, Long>();
+			for (ThreadInfo info : threadMxBean.dumpAllThreads(false, false)) {
+			    threadMxBean.getThreadInfo(info.getThreadId());
```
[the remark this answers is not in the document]
desyncr: Ah, a left over I think.

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
desyncr: haha, 1 = nroCPUs, I'll have to revise this code as the logic behind it is not clear to me either.

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
Bombe: Yeah, doing that as its own component, independent of this page, that would be cool, a lot of stuff from this could go into that and not clutter up this code. :)

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
Bombe: Yeah, I think the uptime is in milliseconds but the CPU time is in nanoseconds, hence multiplying with a million. Hmm, and if you’re using nanoseconds you can store this value in a `long` and only convert to double on the division below.

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
desyncr: I'll look into that. It'd avoid any cluttering and make it easier to remove / adapt if we need to. Any example components to look at for inspiration?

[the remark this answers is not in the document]
Bombe: I don’t think we have something like that, or, well, we probably have a lot of stuff that does something like that and is reachable from `Node` somehow but we never thought of it as components. Maybe something like `ThreadStats` in `freenet.node.stats`? That’s where I would put it unless I find something better while doing that. :)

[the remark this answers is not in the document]
desyncr: Nice one. I'll start from that and I'll see if there's a better spot. 👍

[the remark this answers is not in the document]
desyncr: Done. Created `NodeDiagnostics` as a module to add future metrics. For now it only carries `ThreadMonitor`.

- For now I choose to leave the sleep/queuedTimedJob intervals hardcoded but they could easily be moved into a configuration.
- There's some syncronization/thread safety going on with `nodeThreadInfos` that I'm not sure I'm doing right.
- I corrected the % CPU calculation, here's the reference I used: https://gist.github.com/lpsandaruwan/f2cef0aa91ae68cb041c7ecda04a0724#file-jvmcpuusage-java-L78

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
desyncr: Fixed. Here's the reference with some comments: https://gist.github.com/lpsandaruwan/f2cef0aa91ae68cb041c7ecda04a0724#file-jvmcpuusage-java-L78

desyncr:
> the code looks good. Can you check whether there is something sensitive in the thread name when you insert a file? (any info about the target key)

From what I can see using the thread info there's no sensitive information but I'll look closely at the different places this is being used.

> can you check that you don’t change the indentation? The current format isn’t ideal, but better strange but consistent indentation than inconsistent indentation.

I corrected the code styles and imports, should be following the convention now :+1:

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
```diff
@@ -0,0 +1,8 @@
+/* This code is part of Freenet. It is distributed under the GNU General
+ * Public License, version 2 (or at your option any later version). See
+ * http://www.gnu.org/ for further details of the GPL. */
+package freenet.node.diagnostics.diagnostics;
+
+public interface Diagnostics {
```
Bombe, line 6: I feel this name is somehow wrong because a `start` method and diagnostics have very little to do with each other…

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+    private final NodeStats nodeStats;
+    private final Ticker ticker;
+
+    /** Interval in ms the ThreadMonitor is launched */
+    private static final int MONITOR_INTERVAL = 10000;
+    private static final String MONITOR_THREAD_NAME = "NodeDiagnostics: thread monitor";
+
+    /** Sleep interval to calculate % CPU used by each thread */
+    private static final int CPU_SLEEP_INTERVAL = 1000;
+    List<NodeThreadInfo> nodeThreadInfos    = Collections.synchronizedList(new ArrayList<>());
```
Bombe, line 30: All these member fields should be `private`, nobody else should ever need to access them.

Bombe, line 30: `nodeThreadInfos` (the list, not the reference) is never modified so it doesn’t need to be a synchronized list. The way it is used here I would recommend using an `AtomicRef` to store the latest list; that way we can also get rid of the `synchronized` keyword on the getter below (which we should get rid of either way 🙂).

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+    public void start() {
+        scheduleNext(0);
+    }
+
+    private void scheduleNext() {
+        scheduleNext(monitorInterval);
+    }
+
+    @Override
+    public void run() {
```
Bombe, line 76: I would suggest something different here, in order to get rid of the `sleep` in here which I still find very distressing. 😀

1. Get snapshot T of all threads and their CPU time.
2. If run for the first time (some `lastRun` is `-1` or whatever), skip until 6.
3. Calculate total CPU time since last run (i.e. time ⨉ CPU count).
4. For all threads in T that we have a previous time for, calculate percentage of runtime.
5. Store all those times in some thread info object.
6. Store T so we can access it in the next invocation.
7. Store current time in `lastRun`.

(From the top of my head, might include mistakes, use with discretion.)

This way the thread running this wouldn’t block unproductively for a whole second (an eternity for a computer 😁).

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
```diff
@@ -0,0 +1,8 @@
+/* This code is part of Freenet. It is distributed under the GNU General
+ * Public License, version 2 (or at your option any later version). See
+ * http://www.gnu.org/ for further details of the GPL. */
+package freenet.node.diagnostics.diagnostics;
```
Bombe, line 4: Woah, wait a minute, that’s at least one `diagnostics` too many! 😉

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
Bombe: I don’t like this interface here at all.

I’d much rather have an interface for the `ThreadDiagnostics` which exposes the single `getThreads()` method. This way we are completely free to implement this however we see fit; exposing a concrete class will only increase the public API of Fred.

Actually, I would prefer to even expand this to the `NodeDiagnostics`  as well. Turn that into an interface with a single `getThreadDiagnostics()` method. During startup,`Node` knows how to create a, say, `DefaultNodeDiagnostics` object (with `Ticker` ’n stuff), and `DefaultNodeDiagnostics` knows how to create the `DefaultThreadDiagnostics` and also knows that it needs to be started. (Actually, both should maybe also be stoppable, now that I think of it? Can be left for a later PR, though, as long as a ticket for it is created. 🙂)

Yes, this increases the amount of seemingly useless code but at this point I really prefer to add as little as possible to Fred’s public API, and interfaces (together with some default implementation that should never be used without their respective interfaces) fit that bill perfectly.

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            threadCPU.put(
+                info.getThreadId(),
+                threadMxBean.getThreadCpuTime(info.getThreadId()) - prev
+            );
+        }
+
+        long elapsedUptime = runtimeMxBean.getUptime() - initialUptime;
+        double totalElapsedUptime = (elapsedUptime * operatingSystemMXBean.getAvailableProcessors()) * 10000F;
```
Bombe, line 108: I know where the `10000` comes from (`10000f` is still a float, btw, `10000d`/`10000.0` is where it’s at! 😁) but I would really like to be more explicit here that the value we calulate below is a *percentage* (0–100) and not a *ratio* (0–1), so the the `100` needed for that should be moved to the calculation below.

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+
+        List<NodeThreadInfo> threads = new ArrayList<>();
+        for (Thread t : nodeStats.getThreads()) {
```
Bombe, line 111: (Silly nitpicking, free free to ignore: single-char variable names should be constrained to when you, for whatsoever reason, need to run something 5 times or so; if you iterate over something that has a clearer purpose, I would appreciate a more appropriate name.)

Bombe (commented):
I’m sorry that my reviews always turn into these please-change-everything requests… 😕

src/freenet/node/Node.java
```diff
@@ -723,6 +724,10 @@ public void set(String val) throws InvalidConfigValueException {
 
 	public final SecurityLevels securityLevels;
 
+	/** Diagnostics */
+	public final NodeDiagnostics nodeDiagnostics;
```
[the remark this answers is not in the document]
desyncr: Perfect, yes, I made it public following the rest of the code. I'll make it private.

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
desyncr: Seems reasonable. I actually don't understand the DefaultNodeDiagnostics part but I get the idea of having a simpler interface with only NodeDiagnostics providing access to the (possible) various diagnostics we can put into it.

Re: being able to stop them is something I was considering, so yeah, makes sense.

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
desyncr: I'll take a look at AtomicRef 👍

src/freenet/node/diagnostics/diagnostics/threads/ThreadsDiagnostics.java
desyncr: I see where are you going. I'll see if I can make it work.

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
desyncr: Do you want to have the diagnostics themselves be at the NodeDiagnostics level?

Bombe: Yeah, why not? `f.n.diagnostics.NodeDiagnostics` as main entry point, `f.n.diagnostics.ThreadDiagnostics` for the thread diagnostics interface (as per suggestion), and everything else that may be added in the future. (An argument could be made to keep implementations in a different package but I don’t think adding more package structure here will keep things clearer.)

src/freenet/node/diagnostics/diagnostics/Diagnostics.java
Bombe: If `NodeDiagnostics` is turned into an interface you need an implementation somewhere. The real use of the interface is that we are free to change implementations without having to change Fred’s public API. However, you still *need* at least one concrete implementation (which is created during the Node’s startup) and I simply suggest naming that `DefaultNodeDiagnostics` because it’s our default implementation. Any other name (that is *not* `NodeDiagnosticsImpl`) would be fine, too. 🙂

src/freenet/node/diagnostics/NodeDiagnostics.java
```diff
@@ -0,0 +1,53 @@
+/* This code is part of Freenet. It is distributed under the GNU General
+ * Public License, version 2 (or at your option any later version). See
```
Bombe, line 2: Oh, one other thing, paging @ArneBab for this as well: should we put new code under GPLv3 or do we stay at GPLv2 for everything? I vaguely remember pieces of discussions about this but have completely forgotten any outcome. 😁

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -411,12 +414,41 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
 
 		// drawThreadPriorityStatsBox
 		textBuilder.append("Threads:\n");
-		int[] activeThreadsByPriority = stats.getActiveThreadsByPriority();
-		int[] waitingThreadsByPriority = stats.getWaitingThreadsByPriority();
-		for(int i=0; i<activeThreadsByPriority.length; i++) {
-			textBuilder.append(l10n("running")).append(": ").append(String.valueOf(activeThreadsByPriority[i])).append(" (").append(String.valueOf(i+1)).append(")\n");
-			textBuilder.append(l10n("waiting")).append(": ").append(String.valueOf(waitingThreadsByPriority[i])).append(" (").append(String.valueOf(i+1)).append(")\n");
+		// ID, Name, Priority, Group (system, main), Status, % CPU
```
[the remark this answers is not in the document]
desyncr: Done!

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -411,18 +413,54 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
+			)
+		);
+
+		ThreadDiagnostics threadDiagnostics = node
+			.getNodeDiagnostics()
+			.getThreadDiagnostics();
+
+		List<NodeThreadInfo> threads = threadDiagnostics.getThreads();
+		threads.sort(
+			(o1, o2) -> Double.compare(o2.getCpuTime(), o1.getCpuTime())
```
Bombe, line 445: Is sorting in increasing order a good idea for a default? I’d argue that you want to see thread that are using lots of CPU as the first ones…

src/freenet/node/Node.java
```diff
@@ -3166,6 +3172,8 @@ public void start(boolean noSwaps) throws NodeInitException {
 		// Process any data in the extra peer data directory
 		peers.readExtraPeerData();
 
+		((DefaultNodeDiagnostics)nodeDiagnostics).start();
```
Bombe, line 3175: If you declare the field as `DefaultNodeDiagnostics` you can remove this cast, exactly like you did in `DefaultNodeDiagnostics` with the `DefaultThreadDiagnostics`.

src/freenet/node/diagnostics/DefaultNodeDiagnostics.java
```diff
@@ -0,0 +1,46 @@
...
+ *  Some data pointers are obtained from NodeStats object.
+ */
+public class DefaultNodeDiagnostics implements NodeDiagnostics {
+    private final DefaultThreadDiagnostics defaultThreadDiagnostics;
+
+   /**
+     * @param nodeStats Used to retrieve data points.
+     * @param ticker Used to queue timed jobs.
+     */
+    public DefaultNodeDiagnostics(final NodeStats nodeStats, final Ticker ticker) {
```
Bombe, line 30: These `final`s here are only cluttering up the view and don’t do anything, please remove.

src/freenet/node/diagnostics/DefaultNodeDiagnostics.java
```diff
@@ -0,0 +1,46 @@
...
+        defaultThreadDiagnostics = new DefaultThreadDiagnostics(nodeStats, ticker);
+    }
+
+    public void start() throws NodeInitException {
```
Bombe, line 34: What is this exception and where does it come from and why?! 🙂

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     *
+     * @return
+     */
+    private List<NodeThreadInfo> buildThreadList() {
+        List<NodeThreadInfo> threads = new ArrayList<>();
+        long elapsedUptime = runtimeMxBean.getUptime() - initialUptime;
+        double totalElapsedUptime = (elapsedUptime * operatingSystemMXBean.getAvailableProcessors()) * 10000d;
```
Bombe, line 122: Now you’re missing a couple of zeroes in that number. 🙂

src/freenet/node/diagnostics/DefaultNodeDiagnostics.java
desyncr: I think it was a left over when I refactored it!

src/freenet/clients/http/DiagnosticToadlet.java
desyncr: That's the case yes, higher CPU % are listed at the top.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,151 @@
...
+    private void scheduleNext() {
+        scheduleNext(monitorInterval);
+    }
+
+    /** Map<ThreadId, ThreadCpuTimeNs> */
+    private final Map<Long, Long> threadCPU = new HashMap<>();
+    private long initialUptime = -1;
+
+    @Override
+    public void run() {
```
Bombe, line 79: Hmm, no, this isn’t what I meant. This needs two intervals to get the data; I’d prefer to have each run create new data by diffing it with the last run, except on the first run where no data is created.

desyncr: I don't follow.

Bombe: First run gathers initial data, then schedules the next run. Second run then creates the final data and resets `initialUptime` and `threadCPU`. This means that with an interval of 10 seconds the data only changes every 20 seconds.

All runs should do the same thing; get the current threads, get their previous CPU times, store the diff in the result. This way the data will never be older than the interval (ignoring processing time).

desyncr: Oh, I got you now. :+1: I'll adapt it to do that.

src/freenet/clients/http/DiagnosticToadlet.java
Bombe: What about replacing it with this? It makes the intent clearer and has less code written by us. 🙂 
```suggest
threads.sort(Comparator.comparing(NodeThreadInfo::getCpuTime).reversed());
```

desyncr: Less code is always better :)

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Should be done now! Looks way better :)

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,130 @@
...
+        Map<Long, Long> delta = new HashMap<>(threadCPU.get());
+        Map<Long, Long> threads = new HashMap<>();
+        for (ThreadInfo info : threadMxBean.dumpAllThreads(false, false)) {
+            long threadId = info.getThreadId();
+            long cpuTime = threadMxBean.getThreadCpuTime(threadId);
+
+            delta.put(threadId, cpuTime - delta.getOrDefault(threadId, 0L));
```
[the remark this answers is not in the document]
desyncr: I think I'm over-complicating things with these methods. The objective is to show threads (e.g. from plugins, data store checker, updater among others) that maybe consuming a lot of CPU time. For this showing the % of CPU time between the process threads is enough. Best of all we don't need to read with deltas or storing values to calculate the difference.

I implemented these changes and lot fairly easy to maintain and best of all useful. Let me know what you think.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
Bombe: Okay, the basic loop is nailed down, this looks very good. 😄

I’m not too sure about the `totalCpuTime`, though… if enough threads die between one iteration and the next the percentages might get distorted, I think. Also, one thread that has e.g. a constant 5% of CPU usage can be shown as using any amount of CPU because now the percentage doesn’t depend on wall time anymore but on all other threads.

Thynix:
This is a useful feature!

I'm wondering if the CPU time metrics could be misleading when combined with the pooled executor? If I understand correctly it [changes jobs](https://github.com/freenet/fred/blob/ffe2268f5d9c42aae58476c1559cdeb6a4ca8e2d/src/freenet/support/PooledExecutor.java#L248) within a given thread, so could it report a high CPU utilization for a thread that just got done doing something intensive, and misleadingly attribute that to an idle thread or something that actually doesn't use much CPU? Would it be worth it to modify the pooled executor to set a zero point when switching jobs to avoid this?

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Good point. I simplified the logic to avoid having "copies" of threads, it just uses one source of truth: nodeStats. It takes a snapshot of that, and creates the "NodeThreadInfo" list with the information, including CPU time.

Now I also separated the "CPU %" logic and moved it up to the client code to deal with the presentation logic (e.g. show as actual values or percentage).

I also changed the way we compute the % CPU by calculating the wall (all threads, not only node's). That should be more accurate.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,100 @@
...
+    private final NodeStats nodeStats;
+    private final Ticker ticker;
+
+    /** Sleep interval to calculate % CPU used by each thread */
+    private static final int MONITOR_INTERVAL = 1000;
+    private static final String MONITOR_THREAD_NAME = "NodeDiagnostics: thread monitor";
+
+    private final AtomicReference<List<NodeThreadInfo>> nodeThreadInfo = new AtomicReference<>(new ArrayList<>());
+
+    private final ThreadMXBean threadMxBean               = ManagementFactory.getThreadMXBean();
```
desyncr, line 32: TODO: Remove this blank space.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
Bombe: Maybe I’m misunderstanding what you are doing but now the client (i.e. the toadlet) has no way of knowing *when* a thread has used *any* CPU time because now *all* it knows is how much time a thread has used *in total*. And again, the only way to compare a thread’s runtime here is against all other threads but that doesn’t seem to be helpful, escpecially with longer-running threads (like e.g. the HTTP serving threads, or datastore-related threads) which at some point will probably have more runtime collected than any other thread and will completely obscure any useful information that can be taken from this.

I think that what you did two iterations ago (i.e. calculating the delta here and just exposing that) in combination with the current loop would be best.

desyncr: Ah, I don't know what I was thinking. You're right. I re-introduced the "delta" calculation although I tried to simplify it. Now DefaultThreadDiagnostics will calculate the delta CPU Time for each thread and if no delta can be calculated it will return 0.
The percentage calculation I'd leave it to the client code to deal with it.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -70,18 +70,46 @@ private void scheduleNext() {
...
+     * If there's no previous snapshot of CPU Time for the thread this method
+     * will return 0.
+     *
+     * @param threadId Thread ID to get the CPU usage
+     * @return Delta CPU time.
+     */
+    private double getCpuTimeDelta(long threadId) {
+        double current = threadMxBean.getThreadCpuTime(threadId);
+
+        double cpuUsage = current - threadCpu.getOrDefault(threadId, current);
```
Bombe, line 89: > `.getOrDefault(…, current)`

Nice trick! 👍

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -441,15 +441,13 @@ private StringBuilder threadsStats() {
...
-
-		ThreadMXBean threadMxBean = ManagementFactory.getThreadMXBean();
-		double totalCpuTime = Arrays.stream(threadMxBean.dumpAllThreads(false, false))
-			.mapToDouble(
-				(thread) -> threadMxBean.getThreadCpuTime(thread.getThreadId())
-			).sum();
-
 		threads.sort(Comparator.comparing(NodeThreadInfo::getCpuTime).reversed());
 
+		double totalCpuTime = threads
```
Bombe: You know what I just realized as I wondered why you would do this again? At this point we don’t actually know how much time was spent between two measurements used to create the delta. And as I don’t want this toadlet to fiddle around with the configuration I think we should put the list of thread infos _and_ the timing information into a new object… maybe a `ThreadDeltaSnapshot`? Hmm, sounds horrible. What do you think?

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public void start() {
+        scheduleNext(0);
+    }
+
+    private void scheduleNext() {
+        scheduleNext(monitorInterval);
+    }
+
+    private final HashMap<Long, Double> threadCpu = new HashMap<>();
```
Bombe, line 73: I think we might need a cleanup for this map at the end of the loop, just something simply like remove all thread IDs that are not in the current threads anymore.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,122 @@
...
+     * snapshot and the current CPU Time.
+     *
+     * If there's no previous snapshot of CPU Time for the thread this method
+     * will return 0.
+     *
+     * @param threadId Thread ID to get the CPU usage
+     * @return Delta CPU time.
+     */
+    private double getCpuTimeDelta(long threadId) {
+        double current = threadMxBean.getThreadCpuTime(threadId);
```
Bombe, line 87: The CPU time is a `long` so we shouldn’t coerce it into a `double`. (This also touches `NodeThreadInfo.cpuTime` below.)

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Yeah, I was wondering how much could this keep on growing.

src/freenet/clients/http/DiagnosticToadlet.java
desyncr: I considered this myself but I wanted to keep it as simple as possible. ThreadsSnapshot is not too bad with ThreadsSnapshot#getDelta, ThreadsSnapshot#getThreads and it can grow to contain more metadata.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
Bombe: Well, it would probably be fine for a couple of days, maybe even weeks but it very much depends on how much thread churn there is which without further investigation is hard to determine so I think it’s better to just clean it up. Put in a nice little method called “removeOldThreadDeltas” or something like that and it’s even self-documenting! \o/

desyncr:
Few things missing:
- Configuration toadlet to support: Activate/Deactivate monitoring, delta/interval
- Review comment regarding pooled executor https://github.com/freenet/fred/pull/727#issuecomment-828666181

src/freenet/clients/http/DiagnosticToadlet.java
Bombe: Heh, yes, there’s “simple” and there’s “simple.” 😁 

As long as we thought we’d only need the list of threads the list of threads was enough but now that we see that we need more it’s totally fine to add a new class into this that contains exactly what we need.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Well, I could as well do `threadCpu.keySet().removeIf(key -> threads.contains(key) )`. I haven't tested it though.

Bombe: Yes, if that works it would probably be clear enough that a separate method wouldn’t be warranted.

Bombe:
> Would it be worth it to modify the pooled executor to set a zero point when switching jobs to avoid this?

I don’t think the effort would be worth it. If you split up a thread’s runtime into A (the previous job, large runtime) and B (current job, small runtime) then you still wouldn’t see where the larger amount of time has been spent; it would just be a little less misleading but that only until the next measurement is complete because by then either we get the real runtime for the job or the thread has already been replaced by something else. Plus, if a thread can have two jobs running in a measurement interval it can also have 1000 jobs running in a measurement interval and I don’t think anybody wants to keep track of that. :)

We would have to modify both our own executor (which I think we should get rid off in favour of one of the JDK implementations) and the diagnostics for something that has questionable gain to begin with…

Thynix:
Ooh, okay. Yep. Very good points. Never mind then.

Thynix:
Is it worth adding a notice to the UI about this limitation? It doesn't seem evident from the UI, and seems only especially meaningful if a given job regularly shows up with high CPU time. Maybe something like:

> This CPU utilization information mostly consists of threads that work on many tasks. High utilization only becomes attributable to a specific job, as opposed to a specific worker thread, if it regularly has high utilization attributed to it.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Done, avoiding any coercion now.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Finally added a new private method for it, using streams though.

src/freenet/clients/http/DiagnosticToadlet.java
desyncr: Done, created NodeThreadSnapshot class to support threads list, total cpu time and interval. Interval for now is the configured for the Runner, could be calculated at each executing for further precision but I would argue such precision is not required now.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
```diff
@@ -0,0 +1,151 @@
...
+    private final int monitorInterval;
+
+    private final NodeStats nodeStats;
+    private final Ticker ticker;
+
+    /** Sleep interval to calculate % CPU used by each thread */
+    private static final int DEFAULT_MONITOR_INTERVAL = 1000;
+    private static final String DEFAULT_MONITOR_THREAD_NAME = "NodeDiagnostics: thread monitor";
+
+    private final AtomicReference<NodeThreadSnapshot> nodeThreadSnapshot = new AtomicReference<>();
```
Bombe, line 30: Totally got thrown off on a tangent here but do we have any nullability annotations somewhere in our dependencies? Because if we had we could generate an empty snapshot here when initializing the AtomicRef and could then annotate the get method below with `@Nonnull`. Just for kicks. (Ignore if we don't have nullability annotations in our dependencies.)

Bombe:
> Is it worth adding a notice to the UI about this limitation? It doesn't seem evident from the UI, and seems only especially meaningful if a given job regularly shows up with high CPU time. Maybe something like:
> 
> > This CPU utilization information mostly consists of threads that work on many tasks. High utilization only becomes attributable to a specific job, as opposed to a specific worker thread, if it regularly has high utilization attributed to it.

Hmm, no, I honestly wouldn't bother with that at all. A single measurement is not of big use and in a series of measurements any misattribution would disappear on its own.

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -410,19 +412,59 @@ public int compare(PeerNodeStatus firstNode, PeerNodeStatus secondNode) {
...
 
+	private StringBuilder threadsStats() {
+		StringBuilder sb = new StringBuilder();
+
+		ThreadDiagnostics threadDiagnostics = node
+			.getNodeDiagnostics()
+			.getThreadDiagnostics();
+
+		NodeThreadSnapshot threadSnapshot = threadDiagnostics.getThreadSnapshot();
+		double totalCpuTime = threadSnapshot.getTotalCpuTime();
```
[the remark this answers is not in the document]
desyncr: What meaningful value are you looking for here?

src/freenet/node/Node.java
```diff
@@ -2536,7 +2542,30 @@ public void set(Boolean val) throws InvalidConfigValueException,
...
+					return enableNodeDiagnostics;
+				}
+			}
+
+			@Override
+			public void set(Boolean val) throws InvalidConfigValueException,
+				NodeNeedRestartException {
+				synchronized(Node.this) {
+					enableNodeDiagnostics = val;
+					nodeDiagnostics.stop();
```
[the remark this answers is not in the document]
desyncr: I'm not sure, I'll take a look.

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
Bombe: The time a thread was running as a percentage of wall time and not as a percentage of all time spent running threads. The latter doesn’t give me anything; for threads A, B, and C with a runtime of 0.1s, 0.4s, and 0s and a measuring interval of 1s I want to see 10%, 40%, 0% and _not_ 20%, 80%, 0%.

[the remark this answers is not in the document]
desyncr: Alright, I got you now. Makes total sense :+1:

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: I've no clue about this one. I'll check the dependencies though.

src/freenet/node/Node.java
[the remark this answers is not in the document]
desyncr: It actually is smart enough to only apply the callback only when the option is modified.

src/freenet/node/Node.java
```diff
@@ -2536,7 +2542,30 @@ public void set(Boolean val) throws InvalidConfigValueException,
...
+			public Boolean get() {
+				synchronized(Node.this) {
+					return enableNodeDiagnostics;
+				}
+			}
+
+			@Override
+			public void set(Boolean val) throws InvalidConfigValueException,
+				NodeNeedRestartException {
+				synchronized(Node.this) {
```
Bombe, line 2557: Minor formatting nitpicking: as `synchronized` is a keyword there should be a space between it and the opening parens.

desyncr: Did a bit of clean up here. Let me know if there's a formatting issue.

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1105,6 +1105,8 @@ Node.enablePerNodeFailureTables=Enable per-node failure tables?
 Node.enablePerNodeFailureTablesLong=Enable automatically rerouting around nodes that failed a request within the last 10 minutes?
 Node.enableRoutedPing=Enable FNPRoutedPing?
 Node.enableRoutedPingLong=Enable FNPRoutedPing? Only useful in simulations, not on the real network. Turn it off.
+Node.enableDiagnostics=Enable diagnostics?
+Node.enableDiagnosticsLong=Enable diagnostics? Diagnostics keep track of thread CPU usage. It can be see in /diagnostics page.
```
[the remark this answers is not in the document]
desyncr: Good suggestion. Updated it. Let me know what you think!

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Apparently there's something like `com.sun.istack.internal`, although it fails to compile. Other than that there's no single dependency pulling the notation.

src/freenet/l10n/freenet.l10n.en.properties
```diff
@@ -1105,6 +1105,8 @@ Node.enablePerNodeFailureTables=Enable per-node failure tables?
 Node.enablePerNodeFailureTablesLong=Enable automatically rerouting around nodes that failed a request within the last 10 minutes?
 Node.enableRoutedPing=Enable FNPRoutedPing?
 Node.enableRoutedPingLong=Enable FNPRoutedPing? Only useful in simulations, not on the real network. Turn it off.
+Node.enableDiagnostics=Enable Diagnostics?
+Node.enableDiagnosticsLong=By enabling Diagnostics the node will keep detailed information of it's inner-workings (such as CPU usage per thread) which can help to troubleshoot problems. The collected data is kept in memory (i.e. not persisted on disk) and it's not sent to anybody over the network.
```
Bombe, line 1109: The first "it's" needs to be an "its", otherwise perfect. 👍

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
Bombe: Yes, `com.sun` is generally best avoided and it's also not what I was looking for. Okay, not important. :)

src/freenet/node/diagnostics/NodeDiagnostics.java
ArneBab: "GPLv2 or later" sounds good. I wouldn’t change that without need (need to check every part of the code whether it is "or later"). If we make even one file GPLv3 or later, then the whole project must be GPLv3 or later, which would pull more checks and changes.

When we include code under Apache 2.0, we’ll have to move the project to GPLv3 or later, then we could think about moving individual files to GPLv3 or later.

ArneBab:
> > the code looks good. Can you check whether there is something sensitive in the thread name when you insert a file? (any info about the target key)

> Edit: There's this one maybe: `Network Interface Acceptor for ServerSocket[addr=/127.0.0.1`, not sure if it's too sensitive.

This could actually be sensitive if people use the most secure proxy-method where the bindTo-Address is 127.x.y.z, not 127.0.0.1 — getting this address would allow remote websites to search for the locally running service much more efficiently. But this could be solvable with a simple special-casing.

src/freenet/clients/http/DiagnosticToadlet.java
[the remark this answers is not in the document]
desyncr: Done, now calculating cpu time as % of wall time.

src/freenet/node/diagnostics/NodeDiagnostics.java
Bombe: Alright, thanks!

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
Bombe: Hmm, you know, on the other hand, we could still initialize this here with an empty snapshot which would enable us (haha, you, really 😁) to remove the null check in the render method above. Yes, I think I would prefer that. Null checks are a pest.

src/freenet/node/diagnostics/threads/NodeThreadSnapshot.java
```diff
@@ -0,0 +1,48 @@
...
+    private final double totalCpuTime;
+    private final int interval;
+
+    /**
+     * @param threads List of threads for this snapshot.
+     */
+    public NodeThreadSnapshot(List<NodeThreadInfo> threads, int interval) {
+        this.threads = new ArrayList<>(threads);
+
+        totalCpuTime = Math.max(1, threads
```
Bombe, line 20: I think we don’t need this anymore. Burn it with 🔥!

Bombe (commented):
Almost done. Thank you for your awesome work!

Bombe:
> > Edit: There's this one maybe: `Network Interface Acceptor for ServerSocket[addr=/127.0.0.1`, not sure if it's too sensitive.
> This could actually be sensitive if people use the most secure proxy-method where the bindTo-Address is 127.x.y.z, not 127.0.0.1 — getting this address would allow remote websites to search for the locally running service much more efficiently. But this could be solvable with simple special-casing for the specific thread.

Shouldn't that be more sensibly be handled by making the diagnostics toadlet available to connections with full access only?

ArneBab:
> > > Edit: There's this one maybe: `Network Interface Acceptor for ServerSocket[addr=/127.0.0.1`, not sure if it's too sensitive.
> 
> > This could actually be sensitive if people use the most secure proxy-method where the bindTo-Address is 127.x.y.z, not 127.0.0.1 — getting this address would allow remote websites to search for the locally running service much more efficiently. But this could be solvable with simple special-casing for the specific thread.
> 
> Shouldn't that more sensibly be handled by making the diagnostics toadlet available to connections with full access only?

Yes, you‘re right. Also if the website manages to connect to the toadlet, it must already have the correct IP.

Bombe:
> 
> 
> > > > Edit: There's this one maybe: `Network Interface Acceptor for ServerSocket[addr=/127.0.0.1`, not sure if it's too sensitive.
> > 
> > 
> > > This could actually be sensitive if people use the most secure proxy-method where the bindTo-Address is 127.x.y.z, not 127.0.0.1 — getting this address would allow remote websites to search for the locally running service much more efficiently. But this could be solvable with simple special-casing for the specific thread.
> > 
> > 
> > Shouldn't that more sensibly be handled by making the diagnostics toadlet available to connections with full access only?
> 
> Yes, you‘re right. Also if the website manages to connect to the toadlet, it must already have the correct IP.

Good news! According to [FProxyToadlet#1245](https://github.com/freenet/fred/blob/next/src/freenet/clients/http/FProxyToadlet.java#L1245) this already happens so we're good here. :)

src/freenet/node/diagnostics/threads/NodeThreadSnapshot.java
desyncr: Added default empty snapshot 👍 Removed unused code.

src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
desyncr: Great idea! Added the empty snapshot on initialisation. No need for null check and no race conditions can happen now :)

src/freenet/clients/http/DiagnosticToadlet.java
```diff
@@ -2,12 +2,13 @@
...
 import java.io.File;
 import java.io.IOException;
+import java.lang.management.*;
 import java.net.URI;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.Map;
+import java.util.*;
```
desyncr, line 9: Damn you IntelliJ!

desyncr:
I think in terms of functionality all is done. I'll do some clean up on syntax though.

Bombe (approved):
You know what? I think this is it. 🙂

Thank you for your efforts!

ArneBab:
Looking at the frequency of name changes shows that the majority of threads do not report the name of the actual runnable. How viable would it be to reduce the collection timeframe? Would there be enough info in 100ms or in 30ms?

desyncr:
> Looking at the frequency of name changes shows that the majority of threads do not report the name of the actual runnable. How viable would it be to reduce the collection timeframe? Would there be enough info in 100ms or in 30ms?

Such slow intervals to be seems like a lot of work. Mostly because I imagine the user only watching at this a couple of seconds to figure out something and then leaving it alone. So a lot of wasted CPU cycles.

Maybe this value can be configured, like turning this functionality on/off is now.

ArneBab:
desyncr ***@***.***> writes:

>> Looking at the frequency of name changes shows that the majority of threads do not report the name of the actual runnable. How viable would it be to reduce the collection timeframe? Would there be enough info in 100ms or in 30ms?
>
> Such slow intervals to be seems like a lot of work.

I meant milliseconds. Can we make it so short that only a fraction of
the threads switch runnables?

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein
ohne es zu merken

Bombe:
> desyncr ***@***.***> writes:
> > Looking at the frequency of name changes shows that the majority of threads do not report the name of the actual runnable. How viable would it be to reduce the collection timeframe? Would there be enough info in 100ms or in 30ms? Such slow intervals to be seems like a lot of work.
> I meant milliseconds. Can we make it so short that only a fraction of the threads switch runnables?

I think you’re still trying to fix a problem that is an edge case. We can of course take another look at it when we see how it performs in the wild but for now I think both the current code and the default interval of 1s is fine.

ArneBab:
Could you check this patch and adjust the measurement period to minimize the switching threads?

```
diff --git a/src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java b/src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
index c58795291..b65019807 100644
--- a/src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
+++ b/src/freenet/node/diagnostics/threads/DefaultThreadDiagnostics.java
@@ -37,6 +37,8 @@ public class DefaultThreadDiagnostics implements Runnable, ThreadDiagnostics {
 
     /** Map to track thread's CPU differences between intervals of time */
     private final Map<Long, Long> threadCpu = new HashMap<>();
+    /** Key = Thread ID; Value = System.currentTimeMillis() at which threadCpu was updated. */
+    private final Map<Long, Long> lastThreadCpuMeasurementTime = new HashMap<>();
 
     /**
      * @param nodeStats Used to retrieve data points
@@ -108,10 +110,12 @@ public class DefaultThreadDiagnostics implements Runnable, ThreadDiagnostics {
      * @return Delta CPU time (nanoseconds)
      */
     private long getCpuTimeDelta(long threadId) {
+        long time = System.currentTimeMillis();
         long current = threadMxBean.getThreadCpuTime(threadId);
 
         long cpuUsage = current - threadCpu.getOrDefault(threadId, current);
         threadCpu.put(threadId, current);
+        lastThreadCpuMeasurementTime.put(threadId, time);
 
         return cpuUsage;
     }
@@ -128,6 +132,9 @@ public class DefaultThreadDiagnostics implements Runnable, ThreadDiagnostics {
 
         threadCpu.keySet()
                 .removeIf(key -> !activeThreads.contains(key));
+        
+        lastThreadCpuMeasurementTime.keySet()
+                .removeIf(key -> !activeThreads.contains(key));
     }
 
     @Override
@@ -137,10 +144,11 @@ public class DefaultThreadDiagnostics implements Runnable, ThreadDiagnostics {
             .filter(thread -> thread.getThreadGroup() != null)
             .map(thread -> new NodeThreadInfo(
                     thread.getId(),
-                    thread.getName(),
+                    getThreadNameIfAccurate(thread),
                     thread.getPriority(),
                     thread.getThreadGroup().getName(),
                     thread.getState().toString(),
+                    // Must be called after getThreadNameIfAccurate() to not destroy its input!
                     getCpuTimeDelta(thread.getId())
                 )
             )
@@ -153,4 +161,30 @@ public class DefaultThreadDiagnostics implements Runnable, ThreadDiagnostics {
         purgeInactiveThreads(threads);
         scheduleNext();
     }
+
+    /**
+     * Must be called before {@link #getCpuTimeDelta(long)} to ensure
+     * {@link #lastThreadCpuMeasurementTime} has not been changed yet for the given thread.
+     */
+    private String getThreadNameIfAccurate(Thread thread) {
+        if(!(thread instanceof PooledExecutor.MyThread))
+            return thread.getName();
+        
+        PooledExecutor.MyThread pooledThread = (PooledExecutor.MyThread)thread;
+        String allegedName = pooledThread.getName();
+        long timeOfLastNameChange = pooledThread.lastDisruptiveNameChangeMillis.get();
+        long timeOfBeginningOfMeasurement
+            = lastThreadCpuMeasurementTime.getOrDefault(thread.getId(), Long.MIN_VALUE);
+        
+        if(timeOfBeginningOfMeasurement > timeOfLastNameChange)
+            return allegedName;
+        else {
+            return "Name cannot be determined: It changed during measurement!";
+        }
+    }
 }
diff --git a/src/freenet/support/PooledExecutor.java b/src/freenet/support/PooledExecutor.java
index 4891136b2..e099cdc96 100644
--- a/src/freenet/support/PooledExecutor.java
+++ b/src/freenet/support/PooledExecutor.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.concurrent.atomic.AtomicLong;
 
 import freenet.node.PrioRunnable;
+import freenet.node.diagnostics.threads.DefaultThreadDiagnostics;
 import freenet.support.Logger.LogLevel;
 import freenet.support.io.NativeThread;
 
@@ -168,12 +169,18 @@ public class PooledExecutor implements Executor {
 		}
 	}
 
-	private class MyThread extends NativeThread {
+	public class MyThread extends NativeThread {
 		final String defaultName;
 		volatile boolean alive = true;
 		Job nextJob;
 		final long threadNo;
 		private boolean removed = false;
+		
+		/**
+		 * Allows {@link DefaultThreadDiagnostics} to determine if the Runnable this thread executes
+		 * changed while trying to measure CPU usage percentage of the thread.
+		 */
+		public final AtomicLong lastDisruptiveNameChangeMillis = new AtomicLong(0);
 
 		public MyThread(String defaultName, Job firstJob, long threadCounter, int prio, boolean dontCheckRenice) {
 			super(defaultName, prio, dontCheckRenice);
@@ -213,7 +220,14 @@ public class PooledExecutor implements Executor {
 					}
 					synchronized(this) {
 						if(nextJob == null) {
-							this.setName(defaultName);
+							String oldName = this.getName();
+							String newName = defaultName;
+							this.setName(newName);
+							long changeTimeMillis = System.currentTimeMillis();
+							if(!newName.equals(oldName)) {
+								lastDisruptiveNameChangeMillis.set(changeTimeMillis);
+							}
+							
 							try {
 								wait(TIMEOUT);
 							} catch(InterruptedException e) {
@@ -245,7 +259,24 @@ public class PooledExecutor implements Executor {
 
 				// Run the job
 				try {
-					setName(job.name + "(" + threadNo + ")");
+					String oldName = getName();
+					String newName = job.name + "(" + threadNo + ")";
+					setName(newName);
+					long changeTimeMillis = System.currentTimeMillis();
+					if(!newName.equals(oldName)) {
+						if(!defaultName.equals(oldName)) {
+							lastDisruptiveNameChangeMillis.set(changeTimeMillis);
+						}
+					}
+					
 					job.runnable.run();
 				} catch(Throwable t) {
 					Logger.error(this, "Caught " + t + " running job " + job, t);
```

desyncr:
@Bombe @ArneBab I think that by synchronizing with the PoolerExecutor's thread we can avoid it switching jobs during measurement. This way we'll have the correct job's name for the measure. This should only be a penalty in performance if the Diagnostics module is on.

ArneBab:
> @Bombe @ArneBab I think that by synchronizing with the PoolerExecutor's thread we can avoid it switching jobs during measurement. This way we'll have the correct job's name for the measure. This should only be a penalty in performance if the Diagnostics module is on.

Does this prevent a name change between two measurements. If I understand it correctly, this gets the threads CPU-time and subtracts the previous CPU-time of the thread — whichever name it had then.

If I do get this right, then the only way to make this exact would be to take a snapshot of the CPU-time for each *name* in a secondary structure (i.e. hashmap) and change the executor to record a delta everytime a name changes. The overhead of that could be significant, though, that’s why I’d rather opt to keep it simple and reduce the measuring time enough that name changes during measurement are rare. Maybe avoid wrong output by aggregating all the names the thread had during measurement (on every name change also add the new name to a list of names per thread).

desyncr:
> > @Bombe @ArneBab I think that by synchronizing with the PoolerExecutor's thread we can avoid it switching jobs during measurement. This way we'll have the correct job's name for the measure. This should only be a penalty in performance if the Diagnostics module is on.
> 
> Does this prevent a name change between two measurements. If I understand it correctly, this gets the threads CPU-time and subtracts the previous CPU-time of the thread — whichever name it had then.
> 
> If I do get this right, then the only way to make this exact would be to take a snapshot of the CPU-time for each _name_ in a secondary structure (i.e. hashmap) and change the executor to record a delta everytime a name changes. The overhead of that could be significant, though, that’s why I’d rather opt to keep it simple and reduce the measuring time enough that name changes during measurement are rare. Maybe avoid wrong output by aggregating all the names the thread had during measurement (on every name change also add the new name to a list of names per thread).

I see your point. I think if `PoolerExecutor` implemented some sort of Job id (some pseudo random int), the ThreadDiagnostics could track the CPU per job id, rather than per thread id.

desyncr:
Applied the logic to keep track of CPU usage per job id (rather than thread id). Synchronizing and keeping a copy of the jobs name at the time of measurement.

PR #728, opened 2021-04-13
Pitch black defense only if swapping enabled

ArneBab:
Missing this might be causing stray defensive location changes right now.

### DIFF
--- a/src/freenet/node/LocationManager.java
+++ b/src/freenet/node/LocationManager.java
@@ -190,6 +190,9 @@ public class LocationManager implements ByteCounter {
             @Override
             public void run() {
                 node.ticker.queueTimedJob(this, DAYS.toMillis(1));
+                if (swappingDisabled()) {
+                    return;
+                }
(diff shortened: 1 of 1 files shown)


desyncr (approved):

src/freenet/node/LocationManager.java
```diff
@@ -190,6 +190,9 @@ public void run() {
             @Override
             public void run() {
                 node.ticker.queueTimedJob(this, DAYS.toMillis(1));
+                if (swappingDisabled()) {
```
desyncr, line 193: Should not the job be queued _only_ if opennet is enabled?, if so we'll not would have to worry about this check (it will not hurt though). Then on the OpennetManager start up "notify" the LocationManager to start the job.

ArneBab: In that case the job would be missing, if someone switches off opennet.

ArneBab:
Thank you for your review!

PR #729, opened 2021-04-17
WIP: Node metrics

desyncr:
Node Metrics component to keep track of data points (such as Peers, DataStore, Bandwidth, Activity, etc) and store them in a timeseries database or similar in order to be queries for data inspection (graphing).

To goal is to provide a way to explore the node's activity over a period of time, being able to graph it with Grafana, d3 or similar web-based tools.
Screenshot from 2021-04-18 01-39-05


desyncr:
Created this plugin: https://github.com/desyncr/plugin-NodeMetrics

PR #730, opened 2021-04-21
Ignore output in build-old

ArneBab:


Bombe:
Which process creates this directory? If it’s something not created by the current version of our build scripts it shouldn’t be in the project’s `.gitignore` but in one of your local ones.

ArneBab:
Grepping the whole codebase for build-old, I don’t find any. I wonder how it it could appear in on January this year … 

Closing the pull-request and deleting the stuff locally. Thank you!

PR #734, opened 2021-05-05
m3u-player: prevent flickering on track-transitions

ArneBab:
prevent flickering on track-transitions by keeping the size fixed and overlaying an image of the video

### DIFF
--- a/src/freenet/clients/http/ToadletContextImpl.java
+++ b/src/freenet/clients/http/ToadletContextImpl.java
@@ -465,7 +465,7 @@ public class ToadletContextImpl implements ToadletContext {
 	private static String generateRestrictedScriptSrc() {
 		// TODO: auto-generate these hashes from the path to the source file
 		String[] allowedScriptHashes = new String[] {
-				"sha256-uBohlLWVKw+CT6aoh/dTlBfKXU7QWzLXnomOhe7JxdQ=" // freenet/clients/http/staticfiles/js/m3u-player.js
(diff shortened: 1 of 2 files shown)


desyncr:
🎉  I'll try this tonight! 👍

ArneBab:
Yay!

desyncr:
It's actually a very clever idea! :)

desyncr:
Got to test it finally. Although there's the sound flickering that you mention the video is smoother now. :+1:

ArneBab:
Yay! Thank you for testing!

ArneBab:
> @ArneBab can you please pay a bit attention to what you commit on which branch? Both the bookmark updates _and_ the version number increase do not belong into this pull request.

arg, sorry. I will fix this here. And there’s still a bit of work to be done: adjusting the canvas to the video size does not yet work in all cases.

ArneBab:
@Bombe the branch is cleaned up from stuff not belonging here now.

ArneBab:
@desyncr the latest commit should kill every flickering during transition. The only side-effect should be that the video stops dynamic size changes from resizing of the browser window while it is showing a loading spinner. It restarts them once the next part is loaded

ArneBab:
@desyncr so much for „last“: I found simplifications. Now it should be more robust and work well if the video element size is fixed to something that does not match the video aspect ratio.

desyncr:
I'll take a look tonight! :) The dynamic resizing sounds look a very specific use-case so I'm not worried about that!

ArneBab:
Did you get to look into it?

desyncr:
@ArneBab I could test it just now and looks way better! Both mobile and desktop.

ArneBab:
Thank you!

PR #735, opened 2021-06-05
Fix iframe exclusion logic.

TheSeeker:
sendReplyHeadersFproxy() is for filtered content on freenet.  sendReplyHeaders() is for toadlets like plugins.  we were denying iframes for local toadlets while potentially allowing them for freesites - depending which overloaded method was called.


ArneBab:
I am not sure about this one. We used to have plugins that Freesites could embed as iframes. Currently we do not have them, but they are a valid usecase for interactivity. The HTML-filter ensures that only Freenet- and fproxy-content can be in an iframe.

ArneBab:
I think currently not, but that is a usecase.

PR #737, opened 2021-06-29
Bugfix: skip in support.io.SkipShieldingInputStream must return non-negative

dennisnez:
read() in skip(n) sometimes returns -1, but skip cannot return a negative number. This caused an infinite loop in recent versions of commons-compress (IOUtils.java) which introduced another while loop in it's skip function that assumed the overrided skip() function returned a non-negative value.

### DIFF
--- a/src/freenet/support/io/SkipShieldingInputStream.java
+++ b/src/freenet/support/io/SkipShieldingInputStream.java
@@ -46,6 +46,14 @@ public class SkipShieldingInputStream extends FilterInputStream {
 
     @Override
     public long skip(long n) throws IOException {
-        return n < 0 ? 0 : read(SKIP_BUFFER, 0, (int) Math.min(n, SKIP_BUFFER_SIZE));
+        int retval;
+        if (n < 0)
+            retval = 0;
+        else {
+            retval = read(SKIP_BUFFER, 0, (int) Math.min(n, SKIP_BUFFER_SIZE));
+            if (retval < 0)
+                retval = 0;
+        }
+        return retval;
     }
 }


src/freenet/support/io/SkipShieldingInputStream.java
```diff
@@ -46,6 +46,14 @@ public SkipShieldingInputStream(InputStream in) {
 
     @Override
     public long skip(long n) throws IOException {
-        return n < 0 ? 0 : read(SKIP_BUFFER, 0, (int) Math.min(n, SKIP_BUFFER_SIZE));
+        int retval;
```
Bombe, line 49: Please don't add a variable just for the return value; returning the final value from where it's found is preferred.

src/freenet/support/io/SkipShieldingInputStream.java
```diff
@@ -46,6 +46,14 @@ public SkipShieldingInputStream(InputStream in) {
...
+        if (n < 0)
```
Bombe, line 50: Always enclose code blocks in curly braces, please.

Bombe (changes requested):
Apart from those two things, nice catch!

src/freenet/support/io/SkipShieldingInputStream.java
dennisnez: Well the whole point of this patch is to check the return value of that read() call to make sure it's not negative. How can that be done without introducing a return variable?

Bombe:
```
if (n < 0) {
  return 0;
}

Bombe: Of course, if you _need_ to store a return value, go right ahead but just for the sake of “only have a single exit point” (which is quite stupid)? Please don’t.

nextgens (approved):

PR #739, opened 2021-08-14
NodeConfig was reading a particular config from default on every start up

desyncr:
This particular config value is currently retrieved from the class default value and not from the config as expected.

### DIFF
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -486,7 +486,8 @@ public class Node implements TimeSkewDetectorCallback {
 	private String storeType;
 	private boolean storeUseSlotFilters;
 	private boolean storeSaltHashResizeOnStart;
-	
+	private int storeSaltHashSlotFilterPersistenceTime;
+
 	/** Minimum total datastore size */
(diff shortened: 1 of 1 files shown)


src/freenet/node/Node.java
```diff
@@ -2027,21 +2028,20 @@ public void set(Boolean val) throws InvalidConfigValueException,
 				"Node.storeSaltHashSlotFilterPersistenceTime", "Node.storeSaltHashSlotFilterPersistenceTimeLong", new IntCallback() {
 
 					@Override
-					public Integer get() {
-						return ResizablePersistentIntBuffer.getPersistenceTime();
-					}
+					public Integer get() { return storeSaltHashSlotFilterPersistenceTime;}
```
Bombe, line 2031: Please don’t collapse this into a single line; otherwise this looks good.

Bombe (approved):

src/freenet/node/Node.java
desyncr: Done! Yeah that looked weird!

desyncr:
🎉

PR #740, opened 2021-09-21
OpenJDK 16+ compat: add wrapper.conf argument to allow access to internal openjdk modules

ArneBab:
Needs very careful review because a broken wrapper.conf breaks installs.


ArneBab:
Thanks to xor for an important hint: Need to check with **Java 8**.

Maybe we can only do this on Java 9 upwards → **one more check missing.**

PR #742, opened 2021-10-01
Correct "current size" option in Wizard => datastore size

AC4BB21B:
If you happen to run the installation wizard after you've already configured your node, when reaching the "Datastore size" step, the wizard will load your current datastore size parameter.

However, when you choose a value as datastore size, the wizard will then apply the following operations:
- compute a `clientCacheSize`, of 10% or 200 MiB (whichever is smaller)
- compute a `slashdotCacheSize` of up to 10%
- compute `storeSize` = `[your chosen size] - clientCacheSize - slashdotCacheSize`, and use this as the datastore size.

So, the actual value for datastore size will be different from the chosen value. In particular, if you choose "current size", in the end the datastore will get reduced, leading to a datastore resize (which may take a long time and will lead to its partial purge).

## Example

For instance, if I choose 10 GiB, then in the logs I'll see:

> Setting datastore size to 9016MiB
> Setting client cache size to 200MiB
> Setting slashdot/ULPR/recent requests cache size to 1024MiB

And if I go back to the datastore size wizard step, "current size" will then show 8.8 GiB (the equivalent of 9016 MiB), instead of 10 GiB (what I chose).

## Proposed fix

In this PR, I compute `storeSize + clientCacheSize + slashdotCacheSize`, and I the use this to fill in the "current size" option, instead of just `storeSize`.

Although this is a bit inaccurate as far as what the current size of the datastore really is, it seems to me that it leads to a more coherent behavior.

### DIFF
--- a/src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
+++ b/src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
@@ -44,11 +44,19 @@ public class DATASTORE_SIZE implements Step {
 
 		@SuppressWarnings("unchecked")
 		Option<Long> sizeOption = (Option<Long>) config.get("node").getOption("storeSize");
+		@SuppressWarnings("unchecked")
+		Option<Long> clientCacheSizeOption = (Option<Long>) config.get("node").getOption("clientCacheSize");
+		@SuppressWarnings("unchecked")
+		Option<Long> slashdotCacheSizeOption = (Option<Long>) config.get("node").getOption("slashdotCacheSize");
 		if(!sizeOption.isDefault()) {
 			long current = sizeOption.getValue();
+			long currentTotal = current;
+			if (!clientCacheSizeOption.isDefault() && !slashdotCacheSizeOption.isDefault()) {
+				currentTotal += clientCacheSizeOption.getValue() + slashdotCacheSizeOption.getValue();
+			}
 			result.addChild("option",
 			        new String[] { "value", "selected" },
-			        new String[] { SizeUtil.formatSize(current), "on" }, WizardL10n.l10n("currentPrefix")+" "+SizeUtil.formatSize(current));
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
```diff
@@ -44,11 +44,19 @@ public void getStep(HTTPRequest request, PageHelper helper) {
...
+		@SuppressWarnings("unchecked")
+		Option<Long> clientCacheSizeOption = (Option<Long>) config.get("node").getOption("clientCacheSize");
+		@SuppressWarnings("unchecked")
+		Option<Long> slashdotCacheSizeOption = (Option<Long>) config.get("node").getOption("slashdotCacheSize");
 		if(!sizeOption.isDefault()) {
 			long current = sizeOption.getValue();
+			long currentTotal = current;
+			if (!clientCacheSizeOption.isDefault() && !slashdotCacheSizeOption.isDefault()) {
+				currentTotal += clientCacheSizeOption.getValue() + slashdotCacheSizeOption.getValue();
+			}
```
ArneBab, line 56: Does this mean that the clientCacheSizeOption value is only added if it is not the default?

Is the logic correct if it *is* the default?

ArneBab:
Thank you for your pull request! That looks like a solid improvement (see the comment for polish).

src/freenet/clients/http/wizardsteps/DATASTORE_SIZE.java
AC4BB21B: Yeah I wasn't really sure what to do if those are default...
And we only arrive here if `!sizeOption.isDefault()`, so I don't know if they can even be default at this point 🤔

Perhaps it would make more sense to just put this as a replacement of line 52:
`long current = sizeOption.getValue() + clientCacheSizeOption.getValue() + slashdotCacheSizeOption.getValue();`

AC4BB21B: After thinking it over, I do think it makes more sense to do it this way 9fa12bbf63fb6106738ebb4addb46dceb4a58568

PR #743, opened 2021-10-02
Remove some unused l10n strings

AC4BB21B:
First commit addresses this Mantis ticket https://freenet.mantishub.io/view.php?id=6742

Then I looked for more unused strings.

The 3rd commit removes strings related to removed default bookmarks.

The other 2 remove strings in various areas. A bit hard to make sure a string is unused without having any idea where it used to be used in the first place though, since some strings have pretty short IDs and may be loaded with something else than the usual `l10n()` or `NodeL10n.getBase().getString()`.
I'm currently running a build with all these removed and trying to see if I notice anything missing.


ArneBab:
Thank you for your cleanup! Did you check whether they could be used in plugins?

PR #747, opened 2021-11-19
randomize pitch black defense times

ArneBab:
wait up to two hours before starting the defense, scheduling the next check for a random time on the next day.


Thynix:
Are you hoping only for a review of that the implementation does well what it intends to, or are you also looking for review of if doing so is a good idea?

Thynix (approved):
This certainly seems to do what it intends to in a straightforward way, sure. Whether it makes sense to do that, I don't think I can evaluate.

src/freenet/node/LocationManager.java
```diff
@@ -193,11 +195,12 @@ public void run() {
 
             @Override
             public void run() {
-                node.ticker.queueTimedJob(this, PITCH_BLACK_MITIGATION_FREQUENCY_ONE_DAY);
+                LocalDateTime now = LocalDateTime.now(clockForTesting);
```
Thynix, line 198: Is this clock not just for running (unit?) tests? Why use it at runtime?

src/freenet/node/LocationManager.java
```diff
@@ -246,7 +249,21 @@ public void run() {
...
+        }, (int) (node.fastWeakRandom.nextFloat() * PITCH_BLACK_MITIGATION_STARTUP_DELAY));
+    }
+
+    /** @return the millis to wait until the next pitch black check: tomorrow and at least 12 hours in the future. */
+    private long getNextPitchBlackMitigationDelayMillisecondsTomorrow(LocalDateTime now) {
+        return Math.max(HOURS.toMillis(12), getMillisUntilRandomTimeTomorrow(now));
+    }
+
+    private long getMillisUntilRandomTimeTomorrow(LocalDateTime now) {
+        LocalDateTime tomorrowHourUTC = now
```
Thynix, line 261: Why call it hour? That makes it sound like other components will be discarded. What about something like `tomorrowTime`?

ArneBab:
> Are you hoping only for a review of that the implementation does well what it intends to, or are you also looking for review of if doing so is a good idea?

I’m hoping for a review of that the implementation does well what it intends. This is a result of a discussion with nextgens, so I’m pretty sure that it is a good idea to do this.

src/freenet/node/LocationManager.java
ArneBab: It’s actually a system-clock running at UTC that can be replaced for testing, so the naming was not that great. I renamed it to systemClockUTC now.

Thank you!

src/freenet/node/LocationManager.java
ArneBab: `tomorrowTime` sounds much better — thank you!

ArneBab:
@Thynix I fixed the naming — can you have one more look at the change? (the last commit: https://github.com/freenet/fred/pull/747/commits/91e34c91d539e31969ac2368587f4e2f80c37207 )

PR #749, opened 2022-01-03
fix peer scaling for very fast peers:

ArneBab:
slow peers have insufficient total bandwidth to support the inbound
bandwidth expectation of a fast peer due to square-root peer count
scaling.

### DIFF
--- a/src/freenet/node/OpennetManager.java
+++ b/src/freenet/node/OpennetManager.java
@@ -977,10 +977,12 @@ public class OpennetManager {
 		int max = node.getMaxOpennetPeers();
 		if(ENABLE_PEERS_PER_KB_OUTPUT) {
 			int obwLimit = node.getOutputBandwidthLimit();
-			int targetPeers = (int)Math.round(Math.min(MAX_PEERS_FOR_SCALING, Math.sqrt(obwLimit * SCALING_CONSTANT / 1000.0)));
+			int targetPeers = (int)Math.round(Math.sqrt(obwLimit * SCALING_CONSTANT / 1000.0));
 			if(targetPeers < MIN_PEERS_FOR_SCALING)
 				targetPeers = MIN_PEERS_FOR_SCALING;
 			targetPeers = addMorePeersIfSlowPeersCannotSupplyEnoughBandwidthPerConnection(targetPeers);
+			// limit to max peers
+			targetPeers = Math.min(MAX_PEERS_FOR_SCALING, targetPeers);
 			if(max > targetPeers) {
 				max = targetPeers; // Allow user to reduce it.
 			}
--- a/src/freenet/node/OpennetManager.java
+++ b/src/freenet/node/OpennetManager.java
@@ -980,9 +980,46 @@ public class OpennetManager {
 			int targetPeers = (int)Math.round(Math.min(MAX_PEERS_FOR_SCALING, Math.sqrt(obwLimit * SCALING_CONSTANT / 1000.0)));
 			if(targetPeers < MIN_PEERS_FOR_SCALING)
 				targetPeers = MIN_PEERS_FOR_SCALING;
-			if(max > targetPeers) max = targetPeers; // Allow user to reduce it.
+			targetPeers = addMorePeersIfSlowPeersCannotSupplyEnoughBandwidthPerConnection(targetPeers);
+			if(max > targetPeers) {
+				max = targetPeers; // Allow user to reduce it.
+			}
 		}
 		return max;
+	}
+
+	/**
+	 * A fast peer requires inbound bandwidth per connection proportional to the number of target peers.
+	 * But a peer with fewer than sqrt(targetPeers) connections does not have enough bandwidth
+	 * to support a connection to such a fast peer.
+	 *
+	 * Therefore a fast peer needs more peers than given by the square-root scaling.
+	 *
+	 * To calculate the missing bandwidth, we have to make an assumption: How many peers are too slow?
+	 * We assume that 50% of the peers have the lowest possible peer count.
+	 *
+	 * We need targetPeers packets per peer. We assume that half the peers have the minimum peer count.
+	 * To stay in the 50% slow nodes model, we assume that half the peers of the slow peer are slow,
+	 * the other half are as fast as we are. We receive targetPeers / (minPeers + targetPeers)
+	 * of the packages the slow peer receives.
+	 *
+	 * For each additional peer, we receive one additional packet from fast peers
+	 * and receivedFraction * MIN_PEERS_FOR_SCALING * MIN_PEERS_FOR_SCALING from slow peers.
+	 *
+	 * @param targetPeers the target peers from square root scaling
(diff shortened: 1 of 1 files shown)


ArneBab:
received reviews in FMS, Sone and IRC.

- SSK@nVFNpm3-gyYdF2mIB67JrdQm2vChuAaWjLTztuTLSsM,dQwhU1oouOvlCtIngaqqCkA0mJgUSmmT9coQbFO88ek,AQACAAE/fms|2022-01-03|Message-0
- http://127.0.0.1:8080/forumviewthread.htm?messageuuid=B0D2BE96-258D-4594-94DD-CDC0F58961A6@nVFNpm3gyYdF2mIB67JrdQm2vChuAaWjLTztuTLSsM#B0D2BE96-258D-4594-94DD-CDC0F58961A6@nVFNpm3gyYdF2mIB67JrdQm2vChuAaWjLTztuTLSsM

ArneBab@t-YmshUm43nS16QFieoZMfBdfieb9Opy0a2aMWsoH8o wrote :
> CloudID@nVFNpm3-gyYdF2mIB67JrdQm2vChuAaWjLTztuTLSsM wrote :
>> ArneBab@t-YmshUm43nS16QFieoZMfBdfieb9Opy0a2aMWsoH8o wrote :
>>> During a mostly sleepless night between the years I went over our peer scaling again and discovered a flaw in reasoning: fast peers expect inbound traffic scaled by the number of their peers, but slow peers cannot supply as much. This might explain why we consistently fell short of the allocated bandwidth for fast peers. The following pull-request is a first stab at fixing that:  https://github.com/freenet/fred/pull/749  
>>> 
>>> If you have 30 peers with the current scaling, it will give you 38 now. If you had 60, it will give you 90. So it seems part of the speed patches did have a point. This change should address that much more effectively (because the other peers will then actually be able to handle the added traffic they receive instead of being forced into backoff one step away from the inserter).
>> 
>> It looks like we might need to worry about the MAX_PEERS_FOR_SCALING being overrun with this change (up the max allowed peers).  Last thing we need is for fast peers to get disconnected from because they have more connections than allowed.
> 
> Yes, that was actually a problem in this version of the patch — it’s fixed locally (I plan to push it this evening). The max peers cap has to be applied after the upscaling.
> 
> Thank you!
> 
>>> patch:
>>> 
>>> fix peer scaling for very fast peers:
>>> 
>>> slow peers have insufficient total bandwidth to support the inbound
>>> bandwidth expectation of a fast peer due to square-root peer count
>>> scaling.
>>> 
>>> 1 file changed, 38 insertions(+), 1 deletion(-)
>>> src/freenet/node/OpennetManager.java | 39 +++++++++++++++++++++++++++++++++++-
>>> 
>>> modified   src/freenet/node/OpennetManager.java
>>> @@ -980,11 +980,48 @@ public class OpennetManager {
>>>  			int targetPeers = (int)Math.round(Math.min(MAX_PEERS_FOR_SCALING, Math.sqrt(obwLimit * SCALING_CONSTANT / 1000.0)));
>>>  			if(targetPeers < MIN_PEERS_FOR_SCALING)
>>>  				targetPeers = MIN_PEERS_FOR_SCALING;
>>> -			if(max > targetPeers) max = targetPeers; // Allow user to reduce it.
>>> +			targetPeers = addMorePeersIfSlowPeersCannotSupplyEnoughBandwidthPerConnection(targetPeers);
>>> +			if(max > targetPeers) {
>>> +				max = targetPeers; // Allow user to reduce it.
>>> +			}
>>>  		}
>>>  		return max;
>>>  	}
>>>  
>>> +	/**
>>> +	 * A fast peer requires inbound bandwidth per connection proportional to the number of target peers.
>>> +	 * But a peer with fewer than sqrt(targetPeers) connections does not have enough bandwidth
>>> +	 * to support a connection to such a fast peer.
>>> +	 *
>>> +	 * Therefore a fast peer needs more peers than given by the square-root scaling.
>>> +	 *
>>> +	 * To calculate the missing bandwidth, we have to make an assumption: How many peers are too slow?
>>> +	 * We assume that 50% of the peers have the lowest possible peer count.
>>> +	 *
>>> +	 * We need targetPeers packets per peer. We assume that half the peers have the minimum peer count.
>>> +	 * To stay in the 50% slow nodes model, we assume that half the peers of the slow peer are slow,
>>> +	 * the other half are as fast as we are. We receive targetPeers / (minPeers + targetPeers)
>>> +	 * of the packages the slow peer receives.
>> 
>> I feel like this needs some kinds of blinking red //TODO -- WARNING! ALCHEMY! -- because that 50% *needs* to be measured and not just assumed.  We could just tell our peers what our OBWLimit is and save ourselves a LOT of trouble.
> 
> That’s why there is this huge comment for the method :-)
> 
> We cannot trust the OBWLimit reported by peers, we can somewhat trust the FOAFs (since faking these causes lots of garbage traffic to their node that only costs them bandwidth).
> 
> The more precise fix would be to count the FOAFs we see and adjust the targetPeers based on that, but that would couple the real network with the peer count calculation.
> 
> This change can cause a 50% increase of peers for very fast nodes (for which it underestimates the number of nodes that are too slow).
> 
> But it should have a constant called ASSUMPTION_SLOW_NODE_FRACTION=0.5, right next to the last network size estimate.
> 
> ⇒ will do that — thank you!
> 
>>> +	 * For each additional peer, we receive one additional packet from fast peers
>>> +	 * and receivedFraction * MIN_PEERS_FOR_SCALING * MIN_PEERS_FOR_SCALING from slow peers.
>>> +	 *
>>> +	 * @param targetPeers the target peers from square root scaling
>>> +	 * @return increased estimate to provide enough bandwidth for the fast peer.
>>> +	 */
>>> +	private int addMorePeersIfSlowPeersCannotSupplyEnoughBandwidthPerConnection(int targetPeers) {
>>> +		double receivedFraction = ((double) targetPeers) / (MIN_PEERS_FOR_SCALING + targetPeers);
>>> +		double packetsFromSlowPeer = receivedFraction * MIN_PEERS_FOR_SCALING * MIN_PEERS_FOR_SCALING;
>>> +		if (targetPeers <= packetsFromSlowPeer) {
>>> +			return targetPeers;
>>> +		}
>>> +		double missingPacketsPerSlowPeer = targetPeers - packetsFromSlowPeer;
>>> +		double missingPackets = 0.5 * targetPeers * missingPacketsPerSlowPeer;
>>> +		double additionalPacketsPerAddedPeer = 0.5 * targetPeers + packetsFromSlowPeer;
>>> +		// always compensate for the missing packets. The worst nodes to be underused are the fast ones.
>>> +		return targetPeers + 1 + (int) (missingPackets / additionalPacketsPerAddedPeer);
>>> +	}
>>> +
>>>  	/** Get the target number of opennet peers. Do not call while holding locks. */
>>>  	public int getNumberOfConnectedPeersToAim(LinkLengthClass distance) {
>>>  	    if(distance == null) throw new IllegalArgumentException();
>>> 
>>> 
>> 
>

PR #750, opened 2022-01-04
add voice recognition warning

ArneBab:
see https://www.youtube-nocookie.com/embed/aVo0EJKaEQo?start=1345


desyncr:
LGTM. It would be nice to add the sources as well, don't you think?

ArneBab:
I’d need to track down the original source for that, not the youtube-video of someone talking about it … (I don’t have that source)

PR #752, opened 2022-01-08
try replacing spaces by newlines in noderefs when parsing fails.

ArneBab:
This is now possible (since 2013 where encoding by key==<base64> was added)


desyncr:
Code-wise looks good to me. I'll give it a try tonight!

desyncr (approved):

ArneBab:
Thank you!

PR #753, opened 2022-01-08
New peer via freenet uri

ArneBab:


desyncr (approved):
Looks good! Great addition by the way!

ArneBab:
Thank you! This came from actually trying to add someone as peer and realizing that the simplest way is just uploading the noderef into Freenet — but that we could not actually put that key into the field before.

I had that on my semi-forgotten list of papercuts for years, and now that it hit me I decided to finally fix it :-)

PR #757, opened 2022-01-11
Increase default max threads to 1000

ArneBab:
We have far more peers than 10 years ago to support the larger spread
in node bandwidth around the world.

This is accompanied by a change in wininstaller_innosetup and
java_installer to add -Xss256k to wrapper.conf which decreases the
minimum memory consumption per thread (for its stack) by 75%.

### DIFF
--- a/src/freenet/node/NodeStats.java
+++ b/src/freenet/node/NodeStats.java
@@ -385,8 +385,8 @@ public class NodeStats implements Persistable, BlockTimeCallback {
 			defaultThreadLimit = 500;
 			Logger.minor(this, "Setting 500 thread limit due to <=512MB memory limit. This should be enough but more memory is better.");
 		} else {
-			Logger.minor(this, "Setting standard 1000 thread limit. This should be enough for most nodes.");
 			defaultThreadLimit = 1000;
+			Logger.minor(this, "Setting standard 1000 thread limit. This should be enough for most nodes.");
 		}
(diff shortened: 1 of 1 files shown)


src/freenet/node/NodeStats.java
```diff
@@ -381,9 +381,12 @@ public void shouldUpdate(){
 		} else if(memoryLimit > 0 && memoryLimit < 192) {
 			defaultThreadLimit = 400;
 			System.out.println("Setting 400 thread limit due to <=192MB memory limit. This should be enough but more memory is better.");
-		} else {
-			System.out.println("Setting standard 500 thread limit. This should be enough for most nodes but more memory is usually a good thing.");
+		} else if(memoryLimit > 0 && memoryLimit < 512) {
 			defaultThreadLimit = 500;
+			System.out.println("Setting 500 thread limit due to <=512MB memory limit. This should be enough but more memory is better.");
+		} else {
+			System.out.println("Setting standard 1000 thread limit. This should be enough for most nodes.");
```
desyncr, line 388: I'd put the s.o.p after the defaultThreadLimit is assigned, just for the sake of consistency.

desyncr:
Small comment but it looks good!

src/freenet/node/NodeStats.java
Bombe: I’d rather use the logger and stop writing on `stdout`. :)

hernic: Excellent ! Should avoir lots of freezes

PR #758, opened 2022-02-05
Fix FreenetURI intern() to not forget edition of USK

debbiedub:
I noticed that the suggested edition was not preserved when calling intern() in FreenetURI. Here is a unit test to prove this. This test currently fails.

I assume a fix would be simple enough, I haven't provided any. The consequences of a fix in fred and in the modules, I guess could be confusing so maybe fixing this method is not the best option.

### DIFF
--- a/src/freenet/keys/FreenetURI.java
+++ b/src/freenet/keys/FreenetURI.java
@@ -249,7 +249,7 @@ public class FreenetURI implements Cloneable, Comparable<FreenetURI>, Serializab
 			noCacheURI = true;
 			return this;
 		}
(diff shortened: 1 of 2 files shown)


ArneBab:
Thank you! We’ll have to check for the consequences — I’m not sure whether this might actually already be causing bugs.

To be honest, I think that you found a bug. Interning changes the suggested edition to -1, so this will not be able to find edition zero (0), and requesting an interned USK will be *much* slower than requesting a non-interned one, because it will wait for all checks for future editions to finish — starting at edition 1.

debbiedub:
A fix is added. The newly added unit test succeeds.

There are unit tests failing but those are the same as before this fix.

ArneBab:
Looks good! Thank you!

PR #760, opened 2022-03-05
Update default bookmarks for 1493

ArneBab:


AC4BB21B:
Might be a good idea to remove l10n lines related to removed bookmarks, like `Bookmarks.Defaults.ShortDescription.Enzo`. Otherwise they end up like this #743 and we're never sure what we can or can't remove 👀

ArneBab:
It would be great if you could have a look again, then I’d merge.

AC4BB21B (approved):

ArneBab:
Thank you for checking!

PR #762, opened 2022-03-12
fix: this threw an exception if the checksum was exactly [0,0,0,0].

ArneBab:
CRC32 can legally be 0000, so this had the chance to fail falsely.

It was implemented 2014 in one go for storage for splitfile inserts,
so I expect this to have been just a bug.


Thynix (approved):
I found this added in a947164487ffd9879cd1ff315be4e9f2a3a20430 - seems legit that it's just incorrect cautiousness.

ArneBab:
Thank you for checking!

PR #765, opened 2022-04-01
fix CHK for wrapper-linux-ppc-64 to actually match upstream 3.5.30

ArneBab:
this wad the key for x86-32 — copy-paste error :-(


TheSeeker:
Keyutils Splitfile Explorer says Hashes:
  SHA256: 51a1d1e30b8e050cfe56c8c07996c0629b674986693ededd3de4f303cb48f7e2

and the file downloads successfully, so the hashes must match.

PR #767, opened 2022-04-20
Propagate translation changes

ArneBab:


ArneBab:
@sanity thank you for the review!

PR #768, opened 2022-04-20
m3u-player: more robust sizes, do not use overlay for audio.

ArneBab:


Thynix:
I'm not qualified to review JavaScript, but I'm glad to see such a change! :)

ArneBab:
@Thynix thank you for checking it nontheless!

It might be possible to get reduced stuttering on switching the track, but everything I tested to date just made it worse.

Going full-way and accumulating in the blob like many of the other players do it would make the action of the player non-explicit (you could not just skip to the next fragment) and would increase the complexity and dependence on implementation details of browsers. So I decided to keep this simpler.

PR #769, opened 2022-04-20
add opens jvmargs when building on java 17

ArneBab:


Thynix (approved):

ArneBab:
@Thynix thank you for your review!

PR #778, opened 2022-06-04
Trivuele batch 3

ArneBab:
This is for reviewing the local request routing patches by Trivuele.

I really like the idea of limiting who receives local requests. At the same time I’m worried, though. I see **two dangers**:

- We have less peers we send those requests to, so high-speed nodes that actually respond to requests get better statistics. They will get a bit less than twice as many local requests than they would get otherwise.

- New nodes in the network will not receive any local requests, so they will only route half as many HTL18 requests. A new node will therefore have not only half the anonymity set against an attacker, but also only half the cover traffic.

Also this further centralizes routing on a core of very fast nodes.

The advantage is that this change disarms the attack of regularly changing the opennet location to do superficial surveillance of many nodes. You then need to provide actual long-lived high bandwidth nodes to do any kind of surveillance against local requests.

Please comment what you think. Do we need additional protections for new nodes?


amiga-500:
I love your 3 batch posts @ArneBab .  Great work!  Hope the devs take them seriously.

ArneBab:
@amiga-500 thanks :-) 

These are not by me, though. They are by Trivuele, contributed via FMS. I am reviewing them, and I am one of the devs taking them seriously.

ArneBab:
I see a way to use this to find out exactly whether a given node is the
originator: When you see suspicious requests with a long-lived node,
connect to the target with a short-lived node. If the short-lived node
receives none of the suspicious requests, you know **without a doubt** that
the target is the originator.

PR #780, opened 2022-06-16
Announcement fixes cherry picked from pr 746

ArneBab:
*Changes that look like they should be included even if we do not merge PR #746 , but they need more review.*


vwoodzell (commented):
The new logic seems to be correct. Nice to enable a feature that's never worked. :) I can't comment on the effect that the feature might have.

ArneBab:
Thank you for checking!

We might have to roll this out on one or two seednodes first to check whether it throttles announcements too much. I’ve seen a seednode have all peers in backoff, so this could be important to have (and an overall improvement).

PR #781, opened 2022-06-25
Fix translations and UpdatedVersionAvailableUserAlert.updateNowButton

ArneBab:
It used the FALLBACK_LANG to get the fallback lang, but
iterator().next() skipped to the next value, which is the raw key.

This had also broken our whole translation system.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 next:
 
 - Show a user alert (once every Freenet update) if the datastore is
-  below 10% of available space   thanks to Trivuele
+  below 10% of available space with a link to the store size wizard page
+  to make it easy to increase the store ― thanks to Trivuele
 - Do not store blocks in the cache, if they are eligible for the store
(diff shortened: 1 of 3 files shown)


vwoodzell:
I agree with the change to MainJarUpdater. I can't imagine why we'd want to show the fallback string there instead of the localized one.

I don't think that BaseL10n.getDefaultString() should be changed, though. It doesn't return the raw key unless the key is missing from the fallback translation. I've added [tests](https://github.com/vwoodzell/fred-staging/commit/a6f6e2e7a2ab88b39936e22b808da92e11e5362f) that show it working correctly.

ArneBab:
Thank you for checking! I merged your tests and reverted the change to getDefaultString(). Also I changed the return value documentation to match the implementation.

Good to go? :-)

vwoodzell (approved):
Haha, good idea on the documentation change.

PR #785, opened 2022-07-16
Document and increase max running healing inserts: heal up to 250MiB …

ArneBab:
…per download.


ArneBab:
As a test I downloaded a 300MiB file with the new code. It triggered 2700 healing inserts, so the old limit would have only permitted healing 20% of the dead blocks.

ArneBab:
@desyncr does your :+1: mean you reviewed this?

desyncr:
@ArneBab Yeah, code looks OK 👍

PR #788, opened 2022-08-05
avoid breaking on getting a stray null

ArneBab:


Bombe (commented):
I am skeptical about this. Why should `message` ever be `null`? That code is only used within Fred, right? There’s no user input involved? Because if there isn’t it should be made sure that `null` is never handed into this.

ArneBab:
> I am skeptical about this. Why should `message` ever be `null`? That code is only used within Fred, right? There’s no user input involved? Because if there isn’t it should be made sure that `null` is never handed into this.

Does my reply on FLIP answer your question?

Bombe:
> Does my reply on FLIP answer your question?

Not really. 😁 I traced back every call of that method and I couldn’t see how anything given in to this method can ever be `null`.

ArneBab:
> I traced back every call of that method and I couldn’t see how anything given in to this method can ever be `null`.

Damn :-) 

I also found no way how this could happen, I only saw that it did. When we get there, we’re deep in callback hell with timeouts, so I’m not sure whether I found every possible path.

ArneBab:
@Bombe do you see any way how this change could cause problems? If not, I’d like to merge it, because it fixes a problem that was actually observed.

Bombe:
The only problem I have with it is that I see no way for this to be necessary or even useful. As soon as you (or anybody, really 🙂) can show me how a `null` can be given into that method, I’m all for fixing it. Well, actually, I’d rather fix the root cause for there being a `null` in the first place… 😁

ArneBab:
I checked the reported problem again and found that the NPE was most likely caused by changes done to the code by the reporter. This bug can be closed: Not a Bug.

@Bombe thank you for being sceptical! :heart:

ArneBab:
not a bug

PR #790, opened 2022-08-05
Redesign firsttime wizard (updating the PR from redwerk)

ArneBab:
fixing merge errors and packaging dependencies of https://github.com/freenet/fred/pull/669

TODO: 

- [x] ensure that the new dependencies are available everywhere: pebble, unbescabe, slf4j-api:
  - [x] gradle
  - [x] ant
  - [X] Gentoo (no answer yet, but since Guix works, Gentoo should be easy)
  - [x] Debian
  - [x] Guix
- [x] Check for more transitive dependencies, add in gradle-witness and in ant as needed. 
- [x] Ensure that the packages build with the new dependencies.


ArneBab:
TODO: 

[ ] ensure that the new dependencies are available everywhere. 
[ ] Check all transitive dependencies in gradle-witness and in ant. 
[ ] Ensure that the packages build with the new dependencies.

ArneBab:
I added and updated build-clean.xml and build.properties from the Gentoo ebuild to make it easier to update existing Gentoo, Debian and Guix packages and provide the required build files from a shared place again. All the packages copied it from the Gentoo package, because none of them could use gradle.

ArneBab:
The packagers say they can easily package the deps.

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
```diff
@@ -0,0 +1,329 @@
...
+                    }
+                } catch (NumberFormatException e) {
+                    errors.put("bandwidthMonthlyLimitError",
+                            l10n("valid.number.prefix.bandwidthMonthlyLimit") + " " + e.getMessage());
+                }
+            }
+
+            try {
+                long storageLimit = this.storageLimit.isEmpty() ? 0 : Fields.parseLong(this.storageLimit + "GiB");
+                if (storageLimit < Node.MIN_STORE_SIZE * 5 / 4) { // min store size + 10% for client cache + 10% for slashdot cache
```
Thynix, line 211: This 5/4 business is fair enough because it's explained in the comment, but it seems strange to bury logic like this in form validation code. What do you think about putting it in `DatastoreUtil` as `minStorageLimit` where it'll be alongside `autodetectDatastoreSize()` and `maxDatastoreSize()`?

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
```diff
@@ -0,0 +1,329 @@
...
+                    config.get("node").set("inputBandwidthLimit", Long.toString(bandwidth.downBytes));
+                    config.get("node").set("outputBandwidthLimit", Long.toString(bandwidth.upBytes));
+                }
+            } catch (ConfigException e) {
+                Logger.error(this, "Should not happen, please report! " + e, e);
+            }
+
+            DATASTORE_SIZE.setDatastoreSize(storageLimit + "GiB", config, this);
+
+            // TODO: not sure
```
Thynix, line 311: Not sure about what?

This block looks like it might be incorrect? `isPasswordEmpty` - which looks like it's whether the physical threat level results in a password being set - is checked first, but then there's a branch that sets the master password. Does this only call `core.node.setMasterPassword(password, true)` when it's not supposed to?

Thoughts on renaming `isPasswordEmpty` to something like `passwordRequested`? I was confused by checking both `isPasswordEmpty` and `setPassword.isEmpty()`. Am I missing something or is `setPassword` non-empty if and only if `isPasswordEmpty` is true? Do we need both variables to exist? Are they separate because `isPasswordEmpty` is only set on GET? What's the use of the value being set only on GET? Is it an initial setting for password-ness within the page?

Would it make sense for this block to be something like:

```
                try {
                    if (passwordRequested) {
                        core.node.securityLevels.setThreatLevel(SecurityLevels.PHYSICAL_THREAT_LEVEL.HIGH);
                        core.node.setMasterPassword(password, true);
                    } else {
                        core.node.securityLevels.setThreatLevel(SecurityLevels.PHYSICAL_THREAT_LEVEL.NORMAL);
                        core.node.setMasterPassword("", true);
                    }
                } catch (Node.AlreadySetPasswordException | MasterKeysWrongPasswordException | MasterKeysFileSizeException | IOException e) {
                    Logger.error(this, "Should not happen, please report! " + e, e);
                }
```

src/freenet/clients/http/templates/first-time-wizard.html
```diff
@@ -0,0 +1,115 @@
...
+        </fieldset>
+        <fieldset>
+            <legend>{{ l10n("storage") }}</legend>
+            <p>
+                {% if errors.storageLimitError is not null %}
+                    <div class="error">{{ errors.storageLimitError }}</div>
+                {% endif %}
+                Give
+                <!-- TODO: Populated by the node with some percentage of free space. -->
+                <!-- TODO: Possibly attach this number with a slider of percentage of free space. -->
```
Thynix, line 85: Agreed that this would be nice, but definitely not necessary for an initial implementation.

Thynix (commented):
This is exciting to see! Thanks to redwerk for for implementing this, and to you for rescuing it! I have a few comments, but nothing major. I didn't closely review the template or test out the wizard yet, but wanted to post what feedback I had already.

src/freenet/clients/http/templates/first-time-wizard.html
ArneBab: I agree — I’ll leave in the TODO as a reminder.

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
ArneBab: That’s a good point … your suggestion sounds good (and I guess that was the "not sure" in the TODO), but I’ll have to test this. Thank you!

ArneBab:
One more note: If javascript is disabled, this falls back on the old wizard (or rather: if JS is available, the wizard is redirected to /wiz — the new wizard).

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
ArneBab: I adjusted this and also added a few additional fixes.

ArneBab:
@Thynix I added some more fixes. It would be great if you could have a look at the added commits.

src/freenet/clients/http/templates/first-time-wizard.html
```diff
@@ -0,0 +1,116 @@
...
+                <input id="storage" value="{{ storageLimit }}" name="storage" type="number" min="0.04" max="999999999" step=".01">
```
Thynix, line 87: Would it be practical to have the minimum be a template value that uses the computed minimum? It'd avoid hardcoding the value and duplicating it.

Thynix (approved):
That password variable makes so much more sense now, thank you. I'll verify that I see the same checksums as well as run it myself for testing later today, but the code looks good to me!

src/freenet/clients/http/templates/first-time-wizard.html
ArneBab: Yes, we can! :-) — see the hopefully final commit.

src/freenet/clients/http/FirstTimeWizardNewToadlet.java
ArneBab: damn, I removed the comment when putting it into a constant. I recovered the comment in a hopefully final commit :-)

I would keep it in a constant of the toadlet for the time being. If we need it at another place, we can still pull it into the storage util.

dependencies.properties
```diff
@@ -30,6 +30,30 @@ jna-platform.key=CHK@G0F322jl0AxJqk-yfuRes5vPcpR~eb3pXjckqjr0qmU,mEDDjQHrPDt3Kt9
 jna-platform.size=2327597
 jna-platform.type=CLASSPATH
 jna-platform.order=3
+pebble.version=3.1.5
+pebble.filename=pebble-3.1.5.jar
```
Thynix, line 34: Is it worth moving to Pebble 3.1.6? https://github.com/PebbleTemplates/pebble/releases/tag/3.1.6

dependencies.properties
```diff
@@ -30,6 +30,30 @@ jna-platform.key=CHK@G0F322jl0AxJqk-yfuRes5vPcpR~eb3pXjckqjr0qmU,mEDDjQHrPDt3Kt9
...
+pebble.filename-regex=pebble-*.jar
+pebble.key=CHK@y~p8HMUVXmVgfSnrmUyu2UNXMO9uMDHS5nwo2YuOKvw,yzwLFP0GXa8RjwRpicQCPFKNggDXLkTQKH8nISe0qUY,AAMC--8/pebble-3.1.5.jar
+pebble.size=318169
+pebble.sha256=85e77f9fd64c0a1f85569db8f95c1fb8e6ef8b296f4d6206440dc6306140c1a1
```
Thynix, line 38:
```
steve@steve-ubuntu ~/Downloads> sha256sum pebble-3.1.5.jar 
d253a6dde59e138698aaaaee546461d2f1f6c8bd2aa38ecdd347df17cf90d6f0  pebble-3.1.5.jar
```

Hm, I'm seeing a different hash. I got mine from https://mvnrepository.com/artifact/io.pebbletemplates/pebble/3.1.5 - did you get yours from the build fetching it? Do you know where from?

dependencies.properties
```diff
@@ -30,6 +30,30 @@ jna-platform.key=CHK@G0F322jl0AxJqk-yfuRes5vPcpR~eb3pXjckqjr0qmU,mEDDjQHrPDt3Kt9
...
+pebble.type=CLASSPATH
+pebble.order=4
+unbescape.version=1.1.6
+unbescape.filename=unbescape-1.1.6.RELEASE.jar
+unbescape.filename-regex=unbescape-*.jar
+unbescape.size=173935
+unbescape.sha256=597cf87d5b1a4f385b9d1cec974b7b483abb3ee85fc5b3f8b62af8e4bec95c2c
```
Thynix, line 45:
```
steve@steve-ubuntu ~/Downloads> sha256sum unbescape-1.1.6.RELEASE.jar 
597cf87d5b1a4f385b9d1cec974b7b483abb3ee85fc5b3f8b62af8e4bec95c2c  unbescape-1.1.6.RELEASE.jar
```

This one matches.

dependencies.properties
```diff
@@ -30,6 +30,30 @@ jna-platform.key=CHK@G0F322jl0AxJqk-yfuRes5vPcpR~eb3pXjckqjr0qmU,mEDDjQHrPDt3Kt9
...
+unbescape.key=CHK@jHd93tSbhdRt99JWteWCUbUGVyzwOMH1D8NmH-y7YpU,2-SBelkCK7iT3IGuwAWkGEi2YkEXKtmEYjYpNXkijg0,AAMC--8/unbescape-1.1.6.RELEASE.jar
+unbescape.type=CLASSPATH
+unbescape.order=5
+slf4j-api.version=1.7.25
+slf4j-api.filename=slf4j-api-1.7.25.jar
+slf4j-api.filename-regex=slf4j-api-*.jar
+slf4j-api.size=41203
+slf4j-api.sha256=18c4a0095d5c1da6b817592e767bb23d29dd2f560ad74df75ff3961dbde25b79
```
Thynix, line 53:
```
steve@steve-ubuntu ~/Downloads> sha256sum slf4j-api-1.7.25.jar 
18c4a0095d5c1da6b817592e767bb23d29dd2f560ad74df75ff3961dbde25b79  slf4j-api-1.7.25.jar
```

As does this one.

dependencies.properties
```diff
@@ -30,6 +30,30 @@ jna-platform.key=CHK@G0F322jl0AxJqk-yfuRes5vPcpR~eb3pXjckqjr0qmU,mEDDjQHrPDt3Kt9
 jna-platform.size=2327597
 jna-platform.type=CLASSPATH
 jna-platform.order=3
+pebble.version=3.1.5
+pebble.filename=pebble-3.1.5.jar
+pebble.filename-regex=pebble-*.jar
+pebble.key=CHK@y~p8HMUVXmVgfSnrmUyu2UNXMO9uMDHS5nwo2YuOKvw,yzwLFP0GXa8RjwRpicQCPFKNggDXLkTQKH8nISe0qUY,AAMC--8/pebble-3.1.5.jar
+pebble.size=318169
```
Thynix, line 37: Strangely enough, the size matches.

Thynix (changes requested):
I think we'll want to figure out this hash mismatch for `pebble-3.1.5.jar`, and possibly upgrade to `pebble-3.1.6.jar` to address some (seemingly irrelevant, but still) CVEs.

dependencies.properties
ArneBab: Arg, that would have been an awful failure mode. Checking the sha256sum of the jar, I also get d253a6dde59e138698aaaaee546461d2f1f6c8bd2aa38ecdd347df17cf90d6f0 — but there is another entry with the same hash:

`wrapper-update-mac.files.libwrapper-macosx-universal-64.sha256=85e77f9fd64c0a1f85569db8f95c1fb8e6ef8b296f4d6206440dc6306140c1a1`

which is where I copied the keys from. So  this is a copy-paste error.

Thank you for checking! :heart:

Thynix: Oh! I didn't think to check for that. It seems like it'd make sense to test that the dependencies are properly fetched, then! Would that be doable with a testing update USK?

ArneBab: It should be fixed now.

dependencies.properties
ArneBab: "fix for older android" sounds like we have to …

ArneBab: Actually no — the android build brings its own dependencies, to we can skip that for now.

Thynix (approved):

PR #791, opened 2022-08-24
bookmarks: Add Sharesite, new jFniki url, add FFS, move Software to second position, rename first section to Starting Points

ArneBab:
The goal is to have the directly actionable information earlier. To
show more prominently which cool stuff is possible with Freenet.


vwoodzell (commented):
Looks good. I swapped out my bookmarks file and verified the links.

ArneBab:
Thank you for checking!

PR #797, opened 2022-09-30
Add css transition and word-breaking support by naejadu

ArneBab:
FMS-message:

Word breaking is not fully supported right now.

What I'd need for Freenet-Quickstart is <wbr> as a html tag and "transition-duration" & "transition-delay" as css tags.

<wbr> indicates possible places for the browser to do a word break, which is very much needed for long USK that usually break the page layout because the browser does not break them at all.

This is a very easy fix, you can just add it to freenet.client.filter.HTMLFilter.getAllowedTagVerifiers().group2

----------
I need the css to show a popup with a help tooltip when hovering over a link. The popup should be visible for a second after the cursor leaves the link, so you can actually move your mouse to the popup and copy text from there. Right now the popup vanishes instantly - see the purple links I have on Freenet-Quickstart.

Adding css is more complicated, but I took the time to figure it out and wrote a patch.

Please copy the explanation below to the issue tracker along with the patch.

----------
This patch adds the html <wbr> tag as well as the transition-* css tags to the FRED filter.

Transition tags support multiple entries as a comma separated list:

transition-duration: 5s, 1s;
transition-delay: 1s, 3s;
transition-property: width, transform;
transition-timing-function: ease, ease-out;
Currently not supported: transition-timing-function with the function values: steps(...) & cubic-bezier(...).

Further information: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions

What this patch does:
1) Fix some tab/space issues.
2) Increase array size of "auxilaryVerifiers" because we add new ones.

3) Add three new auxilaryVerifiers (145, 146 & 147) for the transition effects. 3.1) 145: "ti" allows for time values (1s, ...) to be present. 3.2) 146: "id" allows for identifier values (width, height, transform, ...) to be present. Both 145 & 146 are the same as the existing "auxilaryVerifiers[143]", except for ti/id instead of se). 3.3) 147: Allows for a fixed set of keywords (ease, ease-in, ...) to be present. This is the same as the existing "auxilaryVerifiers[70]", except with other keywords.

4) Add each auxilaryVerifier to the corresponding css tag. 4.0) The last 'true' means we can have comma separated values, because transition tags support multiple values. 4.1) "Arrays.asList("145<1,65535>")" means we allow from 1 to 65535 values in the list and apply auxilaryVerifier 145 to each element, so we test that each element is a time value. Same for the other tags. This is similar to the existing "background-origin" tag. 65535 is the value used by existing verifiers.

5) Add <wbr> to HTMLFilter.java.

I tested the new tags with "Filesharing" -> "Filter a file" and they seem to work.

### DIFF
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -180,18 +180,18 @@ class CSSTokenizerFilter {
 		allelementVerifiers.add("caret-color");
 		allelementVerifiers.add("clear");
 		allelementVerifiers.add("clip");
-                allelementVerifiers.add("break-before");
-                allelementVerifiers.add("break-after");
-                allelementVerifiers.add("break-inside");
-                allelementVerifiers.add("column-count");
-                allelementVerifiers.add("column-fill");
-                allelementVerifiers.add("column-gap");
-                allelementVerifiers.add("column-rule-color");
-                allelementVerifiers.add("column-rule-style");
-                allelementVerifiers.add("column-rule-width");
-                allelementVerifiers.add("column-span");
+		allelementVerifiers.add("break-before");
+		allelementVerifiers.add("break-after");
+		allelementVerifiers.add("break-inside");
+		allelementVerifiers.add("column-count");
+		allelementVerifiers.add("column-fill");
+		allelementVerifiers.add("column-gap");
+		allelementVerifiers.add("column-rule-color");
+		allelementVerifiers.add("column-rule-style");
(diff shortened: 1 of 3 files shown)


vwoodzell:
I can't really find my way through the CSSTokenizerFilter code, but the changes seem reasonable, and the tests pass.

There is one thing to fix, though: there are duplicate `if` cases for all four properties in `addVerifier`. The duplicates are identical; one set just needs to be removed.

ArneBab:
@vwoodzell I removed the duplicates and merged — thank you for your review!

PR #799, opened 2022-10-01
Fix prefix detection to support https, web+freenet: and ext+freenet:

TheSeeker:
Support web+freenet: and ext+freenet: as protocol handlers since it doesn't look like freenet: is getting whitelisted any time soon.

https seems to work without the change, perhaps the host part is currently always being stripped before calling the function?  Added the check to allow it in case that ever changes.


ArneBab:
That looks good. Can you check whether you can add a test that ensures that it still works for previously supported URLs but does not let any additional URLs through? This will require extending FreenetURITest.java

PR #803, opened 2022-11-17
Update license to GPLv3 only

freesoftwarerocks:
Effective binary is GPLv3 only

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,301 +1,200 @@
-		    GNU GENERAL PUBLIC LICENSE
-		       Version 2, June 1991
+GNU GENERAL PUBLIC LICENSE
+Version 3, 29 June 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
+Copyright © 2007 Free Software Foundation, Inc. <https://fsf.org/>
 
-			    Preamble
+Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
 
-  The licenses for most software are designed to take away your
-freedom to share and change it.  By contrast, the GNU General Public
-License is intended to guarantee your freedom to share and change free
-software--to make sure the software is free for all its users.  This
-General Public License applies to most of the Free Software
-Foundation's software and to any other program whose authors commit to
-using it.  (Some other Free Software Foundation software is covered by
-the GNU Lesser General Public License instead.)  You can apply it to
-your programs, too.
+Preamble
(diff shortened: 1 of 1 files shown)


freesoftwarerocks:
update license as GPL-3-only

ArneBab:
This needs a lot more explanation than just "Effective binary is …".

I prefer GPLv3 or later to GPLv2 or later, but switching to GPLv3 only reduces long-term compatibility and copyleft safety compared to GPLv2 or later. Also you don’t just change the license of an existing project.

freesoftwarerocks:
> This needs a lot more explanation than just "Effective binary is …".
> 
> I prefer GPLv3 or later to GPLv2 or later, but switching to GPLv3 only reduces long-term compatibility and copyleft safety compared to GPLv2 or later. Also you don’t just change the license of an existing project.

freenet includes code that is apache v2 licensed and it is only compatible with GPLv3 only. Since freenet is GPLv2 or later, mixing codes together produces a binary that is GPLv3 only

freesoftwarerocks:
> This needs a lot more explanation than just "Effective binary is …".
> 
> I prefer GPLv3 or later to GPLv2 or later, but switching to GPLv3 only reduces long-term compatibility and copyleft safety compared to GPLv2 or later. Also you don’t just change the license of an existing project.

https://gitlab.com/inkscape/inkscape/-/blob/master/COPYING

Inkscape is under GPLv2 or later, but includes an LGPLv3 or later library so the binary as a whole is under GPLv3 or later

freesoftwarerocks:
> This needs a lot more explanation than just "Effective binary is …".
> 
> I prefer GPLv3 or later to GPLv2 or later, but switching to GPLv3 only reduces long-term compatibility and copyleft safety compared to GPLv2 or later. Also you don’t just change the license of an existing project.

[Apache License, Version 2.0](https://directory.fsf.org/wiki/License:Apache2.0) ([#apache2](https://www.gnu.org/licenses/license-list.en.html#apache2))
This is a free software license, compatible with version 3 of the GNU GPL.

Please note that this license is not compatible with GPL version 2, because it has some requirements that are not in that GPL version. These include certain patent termination and indemnification provisions. The patent termination provision is a good thing, which is why we recommend the Apache 2.0 license for substantial programs over other lax permissive licenses.

ArneBab:
That’s what’s already written in the README. Our code, however — which is what you get from this repository — is released under GPLv2 *or later*. This allows releasing binaries as GPLv3.

I would merge a PR that only changes the license information of a released binary or providing multiple LICENSE files, but as project we should not merge a change to the license information of the GPLv2 or later code.

ArneBab:
@freesoftwarerocks I now closed the pull request after getting no reply for 9 days.

If you want to implement a version that limits the changes to what’s released as binaries or to the info in the web interface, then feel free to re-open the pull-request.

PR #804, opened 2022-11-17
fix: downloading one-letter or two-letter files failed

ArneBab:


Thynix (approved):
I see this doesn't include the `File.createTempFile()` usages in `MainJarUpdater`, `LegacyJarFetcher`, or `MainJarDependenciesChecker`, but that doesn't seem like a problem given that I can't see such short filenames being used there.

ArneBab:
Yes — I explicitly excluded those from the change because they have fixed filename prefixes, so the problem cannot occur there.

PR #805, opened 2022-11-27
fix test for now cleaner m3u player tag insertion

ArneBab:
the test still contained the m3u-player for a redirect, but this was fixed in https://github.com/freenet/fred/pull/802


Thynix (approved):
Is it worth adding something to check for the HTMLFilter adding the media player when expected?

ArneBab:
I think yes — will do that for the next release, though.

PR #806, opened 2022-11-27
add m3u-player insertion test: is added at end of body

ArneBab:


test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -184,13 +194,25 @@ public void testHTMLFilter() throws Exception {
...
-        assertEquals(DELETED_BASE_HREF, htmlFilter(BAD_BASE_HREF5));
-    }
+		assertEquals(BASE_HREF, HTMLFilter(BASE_HREF));
+		assertEquals(DELETED_BASE_HREF, HTMLFilter(BAD_BASE_HREF));
+		assertEquals(DELETED_BASE_HREF, HTMLFilter(BAD_BASE_HREF2));
+		assertEquals(DELETED_BASE_HREF, HTMLFilter(BAD_BASE_HREF3));
+		assertEquals(DELETED_BASE_HREF, HTMLFilter(BAD_BASE_HREF4));
+		assertEquals(DELETED_BASE_HREF, HTMLFilter(BAD_BASE_HREF5));
+
+		// m3u filter is added when there is a video or audio tag
```
Bombe, line 204: Hmm, this doesn’t sit quite right with me. First of all, I believe this warrants its own test method. 😄

And then I think that including a random String from somewhere (`HTMLFilter.m3uPlayerScriptTagContent()` in this case) will make the test more brittle, and it makes it harder for the test to actually fail if something is wrong with the player script.

Do we have an HTML parser that is not a completely atrocious abomination somewhere in Fred? (I believe in Sone I use Jsoup for these kinds of tests.) I would prefer that this test gets the HTML from the filter, parses it, and then verifies that a certain tag is present as the last element of the body.

ArneBab: I don’t think we have a parser for that, no. The point of this test is to check whether the conditional addition works (only if <video> or <audio> is present), not to test the content of the m3u script (that can change).

ArneBab: I now moved that into its own test and also added a check that just checks whether we have `<script` after content, so if the `m3uPlayerScriptTagContent()` should for some reason be empty or not be a script tag, this will fail.

Bombe (approved):

ArneBab:
Thank you for your review! Merged!

PR #808, opened 2022-11-28
add warnings about IME and voice recognition to the new wizard.

ArneBab:
The first is vitally important in countries with Asian writing systems! The second is important for people used to command their phones by voice; essential warning for the android app.


Thynix (approved):

ArneBab:
Thank you!

PR #809, opened 2022-12-10
Css improvements 2022: combinators, attribute-selectors, and :checked

ArneBab:
- Support CSS combinators, i.e. a ~ b
- support more attribute selectors
- support :checked

### DIFF
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -1727,7 +1727,7 @@ class CSSTokenizerFilter {
 			StringBuilder buf=new StringBuilder(elementString);
 			buf.delete(elementString.indexOf('['), elementString.indexOf(']')+1);
 			elementString=buf.toString();
-			if(logDEBUG) Logger.debug(CSSTokenizerFilter.class, "attSelection="+attSelection+"	elementString="+elementString);
+			if(logDEBUG) Logger.debug(CSSTokenizerFilter.class, "attSelection="+attSelection+"  elementString="+elementString);
 			if(attSelections == null) attSelections = new ArrayList<String>();
 			attSelections.add(attSelection);
 		}
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -3,9 +3,14 @@
  * http://www.gnu.org/ for further details of the GPL. */
 
 package freenet.client.filter;
+
+import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.Reader;
 import java.io.Writer;
+import java.net.URI;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -22,11 +27,6 @@ import freenet.support.Logger;
 import freenet.support.api.Bucket;
 import freenet.support.io.Closer;
 import freenet.support.io.FileBucket;
-
-import java.io.File;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.net.URI;
 
 /** Comprehensive CSS2.1 filter. The old jflex-based filter was very far
  * from comprehensive.
@@ -55,6 +55,8 @@ class CSSTokenizerFilter {
 	static {
 		Logger.registerClass(CSSTokenizerFilter.class);
 	}
+
+	private static final List<String> CSS_COMBINATORS = Arrays.asList(">", "+", "~");
 
 	CSSTokenizerFilter(){
 		passedCharset = "UTF-8";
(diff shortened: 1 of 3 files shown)


ArneBab:
Added tests and fixed a bug that caused only the first char of the key in a[foo$=bar] to be checked, so $= was accepted — falsely.

Found that bug because the test succeeded even after I removed the changes — unexpected success pointed to a bug ☺

Also simplified the implementation for ~, because > and + were already supported, I just hadn’t spotted that. Found while adding the test.

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1814,6 +1815,21 @@ else if(HTMLelement.indexOf('#')!=-1)
...
+		        } else if(attSelection.indexOf("$=")!=-1) {
+		            attSelectionParts=new String[2];
+		            attSelectionParts[0]=attSelection.substring(0,attSelection.indexOf("$="));
+		            attSelectionParts[1]=attSelection.substring(attSelection.indexOf("$=")+2,
+		                    attSelection.length());
+		        } else if(attSelection.indexOf("*=")!=-1) {
+		            attSelectionParts=new String[2];
+		            attSelectionParts[0]=attSelection.substring(0,attSelection.indexOf("*="));
+		            attSelectionParts[1]=attSelection.substring(attSelection.indexOf("*=")+2,
+		                    attSelection.length());
```
Bombe, line 1832: Whenever I see code like this I wonder if there’s a way to make it cleaner and less repetitive…

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1835,6 +1851,7 @@ else if(HTMLelement.indexOf('#')!=-1)
 		            if(!((c>='a' && c<='z') || (c>='A' && c<='Z')))
 		                return null;
 		            for(int i=1;i<attSelectionParts[0].length();i++) {
+		                c=attSelectionParts[0].charAt(i);
```
Bombe, line 1854: I’m not sure what this does but it sure looks like something a test should catch. 🙂

ArneBab: The  fixed test actually caught that ☺

`CSS2_BAD_SELECTOR.add("h1[foo,=bar] {}");` ← if we had had this one before, this would have been caught by the test.

src/freenet/client/filter/CSSTokenizerFilter.java
ArneBab: I removed the redundancy and fixed the scoping of the attSelectionParts on the side (I’m pretty sure garbage collection does not need a reused array as help anymore).

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1801,27 +1802,23 @@ else if(HTMLelement.indexOf('#')!=-1)
...
+				for (String comparisonOperator : operators) {
+						if (attSelection.contains(comparisonOperator)) {
+							attSelectionParts = new String[2];
+							attSelectionParts[0] = attSelection
+									.substring(0, attSelection.indexOf(comparisonOperator));
+							attSelectionParts[1] = attSelection
+									.substring(attSelection.indexOf(comparisonOperator) + comparisonOperator.length());
+							break;
+						}
+					}
```
Bombe, line 1821: The formatting and indentation of this whole thing is a bit wild, and there’s a stray `;` in line 1809. 🙂

ArneBab: The indentation there follows the rules, but other parts partly break that.

Since we have no other CSS PRs right now, I tabified the file now.

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1722,18 +1722,18 @@ public static String HTMLelementVerifier(String elementString, boolean isIDSelec
...
 		while(elementString.indexOf('[')!=-1 && elementString.indexOf(']')!=-1 && (elementString.indexOf('[')<elementString.indexOf(']')))
 		{
-		    if(isIDSelector) return null;
+			if(isIDSelector) return null;
 			String attSelection=elementString.substring(elementString.indexOf('[')+1,elementString.indexOf(']')).trim();
 			StringBuilder buf=new StringBuilder(elementString);
 			buf.delete(elementString.indexOf('['), elementString.indexOf(']')+1);
 			elementString=buf.toString();
-			if(logDEBUG) Logger.debug(CSSTokenizerFilter.class, "attSelection="+attSelection+"  elementString="+elementString);
+			if(logDEBUG) Logger.debug(CSSTokenizerFilter.class, "attSelection="+attSelection+"	elementString="+elementString);
```
Bombe, line 1730: I think you may have been a bit overzealous with the tabifying, usually only multiple _leading_ spaces are turned into tabs… 😄

ArneBab: should be fixed now — thank you for your mind for the detail!

ArneBab:
@Bombe I fixed the merge conflicts and force pushed the result — is this good to go?

Bombe (approved):

ArneBab:

PR #810, opened 2022-12-10
refactor the browser warning wizard page

ArneBab:


src/freenet/clients/http/wizardsteps/BROWSER_WARNING.java
```diff
@@ -104,6 +99,34 @@ public void getStep(HTTPRequest request, PageHelper helper) {
...
+	public String infoBoxHeaderText(boolean incognito, boolean isRelativelySafe) {
+		String infoBoxHeader;
+		if(incognito) {
+			infoBoxHeader = WizardL10n.l10n("browserWarningIncognitoShort");
+		} else if (isRelativelySafe) {
+			infoBoxHeader = WizardL10n.l10n("browserWarningShortRelativelySafe");
+		} else {
+			infoBoxHeader = WizardL10n.l10n("browserWarningShort");
+		}
+		return infoBoxHeader;
```
Bombe, line 127: I’m not a big fan of the one-return-per-method rule of thumb that was somehow created from the single-exit rule from assembly programming times (where it made sense!); just return the value as soon as you have it. It also makes the method easier to understand, IMO.

```
if (incognito) {
    return a;
}
if (isRelativelySafe) {
    return b;
}
return c;
```

Plus, the indentation after your `if`s is inconsistent. 🙂

ArneBab: That’s the "let Intellij do the refactoring"-single return rule :smile:  , but you’re right.

src/freenet/clients/http/wizardsteps/BROWSER_WARNING.java
```diff
@@ -104,6 +99,32 @@ public void getStep(HTTPRequest request, PageHelper helper) {
...
+			if (showTabWarning) {
+				oldBrowserWarnings.add(WizardL10n.l10n("browserWarningFirefoxMightHaveClobberedTabs"));
+			} else if(!incognito) {
+				oldBrowserWarnings.add(WizardL10n.l10n("browserWarningOldFirefoxNewerHasPrivacyMode"));
+			}
+		}
+		return oldBrowserWarnings;
+	}
+
+	public String infoBoxHeaderText(boolean incognito, boolean isRelativelySafe) {
```
Bombe, line 118: Indentation in this method is inconsistent.

src/freenet/clients/http/wizardsteps/BROWSER_WARNING.java
```diff
@@ -104,6 +99,32 @@ public void getStep(HTTPRequest request, PageHelper helper) {
 		        new String[] { "submit", "next", NodeL10n.getBase().getString("Toadlet.next")});
 	}
 
+	public List<String> oldBrowserWarnings(
```
Bombe, line 102: Argh. I really don’t know when to stop nitpicking, I’m sorry. 😢 Feel free to start ignoring comments at any convenient point.

* Inconsistent indentation.
* Parameter order should be modified so that more important parameters come first, in this case `isOldFirefox` seems to be more important than the other ones.
* And again I would try to exit this method as early as possible, i.e. if `!isOldFirefox` I’d return an empty list (convenience method `Collections.emptyList()` exists) and only then start creating the list and adding stuff to it. This also keeps indentation levels in the method itself down for larger parts.

src/freenet/clients/http/wizardsteps/BROWSER_WARNING.java
```diff
@@ -53,25 +56,17 @@ public void getStep(HTTPRequest request, PageHelper helper) {
...
+		if(!oldBrowserWarnings.isEmpty()) {
 			HTMLNode p = infoboxContent.addChild("p");
-			p.addChild("#", WizardL10n.l10n("browserWarningOldFirefox"));
-			if (showTabWarning) {
-				p.addChild("#", " " + WizardL10n.l10n("browserWarningFirefoxMightHaveClobberedTabs"));
-			} else if(!incognito) {
-				p.addChild("#", " " + WizardL10n.l10n("browserWarningOldFirefoxNewerHasPrivacyMode"));
-			}
+			p.addChild("#", oldBrowserWarnings.remove(0));
+			oldBrowserWarnings.forEach(s -> p.addChild("#", " " + s));
```
Bombe, line 69: This kind of obscurs what’s happening here.
```suggestion
			p.addChild("#", oldBrowserWarnings.stream().collect(Collectors.joining(" ")));
```
(Also, add a static import for the `joining` to keep the line shorter.)

PR #811, opened 2022-12-10
compare charset case insensitively (UTF-8 is equivalent to utf-8)

ArneBab:


Bombe:
This looks like it could use a test!

PR #812, opened 2022-12-10
Add meta Charset tests

ArneBab:


test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -258,6 +266,12 @@ private String headFilter(String data) throws Exception {
 		return s;
 	}
 
+	@Test
+	public void testMetaCharset() throws IOException {
+		assertEquals(META_CHARSET, META_CHARSET);
```
Bombe, line 271: I’m gonna need an explanation for these two lines, please… 🙂

ArneBab: Blindness! Utter Blindness! ← sadly that … thank you for your review! I’ll rework this.

ArneBab: It wasn’t WIP until someone detected the utter blindness that hat stricken me :)

ArneBab:
I un-WIP-ped this PR ☺

Now it actually does what it claims (though I’m unhappy about having to use two variables for each test, because we get spaces added inside comments).

Bombe:
> I’m unhappy about having to use two variables for each test, because we get spaces added inside comments

I have additional questions. 😁 

* Why does that happen? Does every comment simply get wrapped in spaces? (WTF?)
* Also, why do you have comments in there? Is it necessary for the test?
* Also, what exactly is it you are actually testing? The name of the test method is incredibly vague and non-descriptive, and the fact that there’s multiple asserts in there doesn’t make it easier to deduce what this test is about.

ArneBab:
> > I’m unhappy about having to use two variables for each test, because we get spaces added inside comments
> 
> I have additional questions. grin

I love your reviews! (when I have the time to address them *duck*)

>     * Why does that happen? Does every comment simply get wrapped in spaces? (WTF?)

WTF was exactly my thought … maybe a way to ensure that it’s impossible to break out of the comment after filtering: `<!-- foo -(filtered out)-> evil -->` could cause evil to escape from the comment.

>     * Also, why do you have comments in there? Is it necessary for the test?

I reproduced the breakage in the Freesite. It isn’t strictly needed for this test, so I removed it from one.

>     * Also, what exactly is it you are actually testing? The name of the test method is incredibly vague and non-descriptive, and the fact that there’s multiple asserts in there doesn’t make it easier to deduce what this test is about.

That’s a great complaint — thank you! The name should have been (and now is): `testThatHtml5MetaCharsetIsPreserved()`

Bombe:
> WTF was exactly my thought … maybe a way to ensure that it’s impossible to break out of the comment after filtering: `<!-- foo -(filtered out)-> evil -->` could cause evil to escape from the comment.

That would be an even bigger WTF.

> The name should have been (and now is): `testThatHtml5MetaCharsetIsPreserved()`

That is acceptable. 🙂 Go ahead and merge it!

PR #815, opened 2022-12-29
fix: first time wizard caused a startup loop with JS enabled (release blocker)

ArneBab:
with JS enabled, a new install redirected from /wizard to /wiz with Javascript and then redirected back to /wizard due to the check for the first time wizard.

### DIFF
--- a/src/freenet/clients/http/SimpleToadletServer.java
+++ b/src/freenet/clients/http/SimpleToadletServer.java
@@ -985,9 +985,10 @@ public final class SimpleToadletServer implements ToadletContainer, Runnable, Li
 			//If the user has not completed the wizard, only allow access to the wizard and static
 			//resources. Anything else redirects to the first page of the wizard.
 			if (!(path.startsWith(FirstTimeWizardToadlet.TOADLET_URL) ||
-				path.startsWith(StaticToadlet.ROOT_URL) ||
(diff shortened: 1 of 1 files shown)


Bombe:
Blocking this on the server side seems like the wrong approach. Why do we include JavaScript that redirects somewhere else, only to have the server redirect back? Is Freenet at odds with itself? 🙂

Thynix:
Do I understand correctly that redirect behavior is not intended behavior, and the JS redirect is more of JS detection?

ArneBab:
Steve Dougherty ***@***.***> writes:

> Do I understand correctly that redirect behavior is not intended behavior, and the JS redirect is more of JS detection?

Yes, the JS redirect is intended to ensure that you only ever get the
new /wiz, if you have working Javascript.

Thank you!

Best wishes,
Arne
-- 
Unpolitisch sein
heißt politisch sein,
ohne es zu merken.
draketo.de

ArneBab:
David Roden ***@***.***> writes:

> Blocking this on the server side seems like the wrong approach. Why do we include JavaScript that redirects somewhere else, only to have the server
> redirect back? Is Freenet at odds with itself? 🙂

Ah, I missed your review — sorry!

The Javascript is just JS detection: only use the new wizard, if you
have JS (because the new wizard depends on JS).

PR #818, opened 2023-01-03
Fix KeepAlive error on new keys

ThePlantEater:
- fix negative maxsize per new fetch override
- also override maxTempLength
- LowLevelGetException also return the throwable because the error doesnt help, the real gets hidden

see: https://github.com/freenet/plugin-KeepAlive/pull/2


ArneBab:
Looks good! Thank you!

PR #820, opened 2023-01-09
🐛 Fix JarClassLoader’s ability to work with ServiceLoader

Bombe:
ServiceLoader uses the `ClassLoader.findResources()` method to locate its service definition files. The default implementation of `findResources()` returns an empty `Enumeration`, making it impossible to use ServiceLoader from plugins.


ArneBab:
This looks good to me. Thank you!

PR #825, opened 2023-01-22
fix regression: default security level is normal

ArneBab:


Bombe (approved):

ArneBab:
The original wizard did something different, so this is a regression from introducing the new wizard :-)

PR #826, opened 2023-01-23
Improve tests in the `freenet.client` package

venfernand:
Improve tests in the `freenet.client` package
----
Depends on #822  and #824 
----

Adds improvements to the tests in the  `freenet.client` package:
* wraps input/output streams with try-with-resources
* removes duplication
* uses better assertion methods
* creates helper utility ResourceFileUtil
* adds better code formatting

### DIFF
--- a/test/freenet/client/FailureCodeTrackerTest.java
+++ b/test/freenet/client/FailureCodeTrackerTest.java
@@ -32,13 +32,11 @@ public class FailureCodeTrackerTest {
     }
 
     private int getStoredLength(FailureCodeTracker f) throws IOException {
-        try (
-            CountedOutputStream os = new CountedOutputStream(new NullOutputStream());
-            DataOutputStream dos = new DataOutputStream(os)
-        ) {
+        CountedOutputStream os = new CountedOutputStream(new NullOutputStream());
+        try (DataOutputStream dos = new DataOutputStream(os)) {
             f.writeFixedLengthTo(dos);
-            return (int) os.written();
         }
+        return (int) os.written();
     }
 
 }
--- a/test/freenet/client/FailureCodeTrackerTest.java
+++ b/test/freenet/client/FailureCodeTrackerTest.java
@@ -12,13 +12,15 @@ import freenet.support.io.NullOutputStream;
 
 public class FailureCodeTrackerTest {
 
-    /** Test that the fixed size representation really is fixed size */
+    /**
+     * Test that the fixed size representation really is fixed size
+     */
     @Test
     public void testSize() throws IOException {
         testSize(false);
         testSize(true);
     }
-    
+
     public void testSize(boolean insert) throws IOException {
         FailureCodeTracker f = new FailureCodeTracker(insert);
         int fixedLength = FailureCodeTracker.getFixedLength(insert);
@@ -30,11 +32,13 @@ public class FailureCodeTrackerTest {
     }
 
     private int getStoredLength(FailureCodeTracker f) throws IOException {
-        CountedOutputStream os = new CountedOutputStream(new NullOutputStream());
-        DataOutputStream dos = new DataOutputStream(os);
-        f.writeFixedLengthTo(dos);
-        dos.close();
-        return (int) os.written();
+        try (
+            CountedOutputStream os = new CountedOutputStream(new NullOutputStream());
+            DataOutputStream dos = new DataOutputStream(os)
+        ) {
+            f.writeFixedLengthTo(dos);
+            return (int) os.written();
+        }
     }
 
 }
--- a/test/freenet/client/FetchContextTest.java
+++ b/test/freenet/client/FetchContextTest.java
@@ -23,7 +23,7 @@ public class FetchContextTest {
             new ArrayBucketFactory(),
             new SimpleEventProducer()
         );
-        final ArrayBucket bucket = new ArrayBucket();
+        ArrayBucket bucket = new ArrayBucket();
         try {
             try (DataOutputStream dos = new DataOutputStream(bucket.getOutputStream())) {
                 context.writeTo(dos);
--- a/test/freenet/client/FetchContextTest.java
+++ b/test/freenet/client/FetchContextTest.java
@@ -14,22 +14,28 @@ import freenet.support.io.ArrayBucketFactory;
 import freenet.support.io.StorageFormatException;
 
 public class FetchContextTest {
-    
+
     @Test
     public void testPersistence() throws IOException, StorageFormatException {
-        FetchContext context = 
-            HighLevelSimpleClientImpl.makeDefaultFetchContext(Long.MAX_VALUE, Long.MAX_VALUE, 
-                    new ArrayBucketFactory(), new SimpleEventProducer());
-        ArrayBucket bucket = new ArrayBucket();
-        DataOutputStream dos = new DataOutputStream(bucket.getOutputStream());
-        context.writeTo(dos);
-        dos.close();
-        assert(bucket.size() != 0);
-        DataInputStream dis = new DataInputStream(bucket.getInputStream());
-        FetchContext ctx = new FetchContext(dis);
-        dis.close();
-        assertTrue(ctx.equals(context));
-        bucket.free();
+        FetchContext context = HighLevelSimpleClientImpl.makeDefaultFetchContext(
+            Long.MAX_VALUE,
+            Long.MAX_VALUE,
+            new ArrayBucketFactory(),
+            new SimpleEventProducer()
+        );
+        final ArrayBucket bucket = new ArrayBucket();
+        try {
+            try (DataOutputStream dos = new DataOutputStream(bucket.getOutputStream())) {
+                context.writeTo(dos);
+            }
+            assertNotEquals(0, bucket.size());
+            FetchContext ctx;
+            try (DataInputStream dis = new DataInputStream(bucket.getInputStream())) {
+                ctx = new FetchContext(dis);
+            }
+            assertEquals(ctx, context);
+        } finally {
+            bucket.free();
+        }
     }
-
 }
--- a/test/freenet/client/filter/CSSParserTest.java
+++ b/test/freenet/client/filter/CSSParserTest.java
@@ -1,5 +1,7 @@
 package freenet.client.filter;
 
+import static org.hamcrest.Matchers.anyOf;
+import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.junit.Assert.*;
 
 import java.io.IOException;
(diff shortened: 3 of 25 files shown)


test/freenet/client/filter/OggPageTest.java
```diff
@@ -1,59 +1,57 @@
...
+		) {
+			testResourceFile(
+				"./ogg/nonsensical_interruption_filtered.ogg",
+				(input) -> readPages(expectedDataStream, input)
+			);
+			testResourceFile(
+				"./ogg/nonsensical_interruption.ogg",
+				(input) -> readPages(actualDataStream, input)
+			);
+			assertArrayEquals(expectedDataStream.toByteArray(), actualDataStream.toByteArray());
```
Bombe, line 27: Oh, wow, this test is bad. 😄 No need to do anything, though, this test was bad before the refactoring, too… 🙂

test/freenet/client/filter/OggFilterTest.java
```diff
@@ -24,49 +22,60 @@ public void setUp() {
...
-		try {
-			filter.readFilter(input, output, null, null, null, null);
-		} catch(DataFilterException e) {}
-		assertTrue(Arrays.equals(new byte[]{}, output.toByteArray()));
-		input.close();
-		output.close();
+	public void testValidSubPageStripped() throws IOException {
+		testResourceFile("./ogg/contains_subpages.ogg", (input) -> {
+			try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
+				assertThrows(
```
Bombe, line 37: Hmm, I haven’t run the test but I don’t think the exception is the point of this test, otherwise there’d be a `fail()` after the `readFilter()`.

test/freenet/client/filter/OggFilterTest.java
```diff
@@ -24,49 +22,60 @@ public void setUp() {
...
+				(input) -> {
+					filter.readFilter(input, output, null, null, null, null);
+					writeToTestOutputFile(output);
+				}
+			);
+			assertArrayEquals(unchangedData.toByteArray(), output.toByteArray());
+		}
+	}
+
+	private void writeToTestOutputFile(ByteArrayOutputStream output) throws IOException {
```
Bombe, line 68: Oof, what is even this?! Looks like somebody committed debugging code! (Again, no action needed.)

test/freenet/client/filter/OggBitStreamFilterTest.java
```diff
@@ -1,55 +1,52 @@
...
-		try {
-			filter.parse(page);
-			fail("Expected exception not caught");
-		} catch(DataFilterException e) {}
-		input.close();
+		testResourceFile("./ogg/pages_out_of_order.ogg", (input) -> {
+			OggPage filterPage = OggPage.readPage(input);
+			OggBitstreamFilter filter = new OggBitstreamFilter(filterPage);
+			OggPage page = OggPage.readPage(input);
+			assertThrows(DataFilterException.class, ()-> filter.parse(page));
```
Bombe, line 41: Ha! Missed a space there after the `()`! 😄

test/freenet/client/filter/GIFFilterTest.java
```diff
@@ -117,17 +117,16 @@ public void testFilterPairs() throws IOException {
 
     @Test
     public void testReject() throws IOException {
+        ContentDataFilter filter = new GIFFilter();
```
Bombe, line 120: I think this declaration was totally fine where it was, you know, close to its use.

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -27,360 +32,363 @@
...
+                e.getCause().printStackTrace();
+            }
+            // Ok.
+        }
+
+        if (failed) {
+            try (FileOutputStream fos = new FileOutputStream("unfiltered")) {
+                fos.write(total);
+            }
+            throw failures.get(0);
```
Bombe, line 345: Gosh, at some point we need to take a fine-toothed comb and unsuck all of these tests…

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -9,13 +11,10 @@
 import java.io.StringWriter;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
+import java.util.*;
```
Bombe, line 14: I am confused. In another test a star import was expanded, and here you’re collapsing it again?

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -918,112 +905,107 @@ public void setUp() throws InvalidThresholdException {
...
+			String key = entry.getKey();
+			String value = entry.getValue();
+			assertTrue("key=\"" + key + "\" value=\"" + filter(key) + "\" should be \"" + value + "\"", filter(key).contains(value));
 		}
+	}
 
-		assertTrue("key=\""+CSS_DELETE_INVALID_SELECTOR+"\" value=\""+filter(CSS_DELETE_INVALID_SELECTOR)+"\" should be \""+CSS_DELETE_INVALID_SELECTORC+"\"", CSS_DELETE_INVALID_SELECTORC.equals(filter(CSS_DELETE_INVALID_SELECTOR)));
-		assertTrue("key=\""+CSS_INVALID_MEDIA_CASCADE+"\" value=\""+filter(CSS_INVALID_MEDIA_CASCADE)+"\"", "".equals(filter(CSS_INVALID_MEDIA_CASCADE)));
+	private void testBadSelectorFiltering(Set<String> badSelectorSet) throws IOException, URISyntaxException {
+		for(String key : badSelectorSet) {
```
Bombe, line 930: Inconsistent whitespace after `for`.

test/freenet/client/FailureCodeTrackerTest.java
```diff
@@ -30,11 +32,13 @@ public void testSize(boolean insert) throws IOException {
     }
 
     private int getStoredLength(FailureCodeTracker f) throws IOException {
-        CountedOutputStream os = new CountedOutputStream(new NullOutputStream());
-        DataOutputStream dos = new DataOutputStream(os);
-        f.writeFixedLengthTo(dos);
-        dos.close();
-        return (int) os.written();
+        try (
```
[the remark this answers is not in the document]
venfernand: Improved this a little.
`DataOutputStream` also closes an underlying stream,
and there is no need to close `CountedOutputStream` in this test.

test/freenet/client/filter/ResourceFileUtil.java
```diff
@@ -0,0 +1,41 @@
...
+                inputStreamConsumer.processDataInputStream(input);
+            }
+        }
+    }
+
+    static ArrayBucket resourceToBucket(String filename) throws IOException {
+        ArrayBucket ab;
+        try (InputStream is = ResourceFileUtil.class.getResourceAsStream(filename)) {
+            if (is == null) {
+                throw new java.io.FileNotFoundException(filename);
```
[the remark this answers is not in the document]
venfernand: `FileNotFoundException` is now thrown from both methods

test/freenet/client/FetchContextTest.java
```diff
@@ -14,22 +14,28 @@
...
-        dis.close();
-        assertTrue(ctx.equals(context));
-        bucket.free();
+        FetchContext context = HighLevelSimpleClientImpl.makeDefaultFetchContext(
+            Long.MAX_VALUE,
+            Long.MAX_VALUE,
+            new ArrayBucketFactory(),
+            new SimpleEventProducer()
+        );
+        final ArrayBucket bucket = new ArrayBucket();
```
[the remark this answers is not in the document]
venfernand: Removed `final` modifier.
I agree, that this is not really necessary here.

test/freenet/client/filter/ResourceFileUtil.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.InputStream;
+
+class ResourceFileUtil {
+
+    @FunctionalInterface
+    interface ResourceFileStreamConsumer {
+        void processDataInputStream(DataInputStream dataInputStream) throws IOException;
+    }
+
+    static void testResourceFile(String fileName, ResourceFileStreamConsumer inputStreamConsumer) throws IOException {
```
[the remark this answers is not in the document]
venfernand: Simplified the `ResourceFileUtil` and made that method to return `DataInputStream`.
This made usage places more easy to read and understand.

Now this package-local `ResourceFileUtil` class has a collection of methods,
which help to convert a resource file to various domain objects.

test/freenet/client/filter/OggFilterTest.java
venfernand: Current impl is more strict and expects an exception to be thrown.
Previous solution with empty `try-catch` may be restored easily, 
but that is not able to capture a moment, when `readFilter()` stops throwing exception for configured input.
There is a special branch in the `readFilter()` method, which reports if no bytes were written,
and `assertThrows` tests that branch as well.
A little bit overlaps with `testEmptyOutputRaisesException()`,
but, in my opinion, more checks are better and those help catch more regressions.

Feel free to reject this change if you do not like it.

test/freenet/client/filter/GIFFilterTest.java
venfernand: Moved back.
Initially tried remove the instantiation from the loop, but this is correct only if `GIFFilter()` has no any internal state.

test/freenet/client/filter/CSSParserTest.java
venfernand: I think it was some automatic effort of IDE.
Expanded it back.

test/freenet/client/filter/CSSParserTest.java
venfernand: Fixed, 
manually typed vs auto-formatted ;)

test/freenet/client/filter/ResourceFileUtil.java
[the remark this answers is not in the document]
Bombe: I really liked the `Consumer` stuff you had in there… are you perchance doing a lot of Kotlin development? 😁

test/freenet/client/filter/OggFilterTest.java
Bombe: My point is that a test should be for one specific thing. If throwing the exception is an expected outcome for `contains_subpages.ogg` then that should be its own test, and then you’d have a second test for the stripped subpages which does not expect the exception.

Then again, why is an exception thrown but valid stuff is still expected to be written to the output? That seems like a horrible design choice…

Bombe (approved):

PR #829, opened 2023-02-12
Improve MultiValueTable

venfernand:
MultiValueTable previously used HashTable class internally and all access methods required to be synchronized. This solution was replaced with ConcurrentHashMap and CopyOnWriteArrayList, which may be accessed concurrently and with less synchronization.

Also changed some methods, which used older collection APIs.

Improved usages of MultiValueTable.

Signed-off-by: Veniamin Fernandes <veniamin.fernandes@proton.me>


ArneBab:
@venfernand thank you for your pull request! This makes the code much nicer.

To ensure that this does not break plugins, it needs to preserve all existing API, so to be able to merge it, this needs shims for all previously existing public methods and constructors.

src/freenet/clients/fcp/RequestStatusCache.java
```diff
@@ -112,7 +113,7 @@ synchronized void removeByIdentifier(String identifier) {
 			downloads.remove(status);
 			FreenetURI uri = status.getURI();
 			assert(uri != null);
-			downloadsByURI.removeElement(uri, status);
+			downloadsByURI.removeElement(uri, (DownloadRequestStatus) status);
```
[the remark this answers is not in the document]
Bombe: Every time you do an `instanceof` check, somebody kills a kitten! 😢

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,128 @@
...
 
     public MultiValueTable(int initialSize) {
-        this(initialSize, 3);
+        table = new ConcurrentHashMap<>(initialSize);
     }
 
-    public MultiValueTable(int initialSize, int initialEntrySize) {
-        table = new Hashtable<K, Vector<V>>(initialSize);
-        ies = initialEntrySize;
-    }
```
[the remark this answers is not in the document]
Bombe: I’m pretty sure all of my plugins use it as well…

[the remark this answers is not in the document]
Bombe: (Not this specific constructor, though, but yeah, existing API should be preserved as much as possible, breaking APIs needs to be _very_ well thought out.)

ArneBab:
@venfernand will you add the shims to preserve the existing API?

venfernand:
> @venfernand will you add the shims to preserve the existing API?

Sorry for being too distructive.
I restored all previous public API in `MultiValueTable`  and marked some methods as `@Deprecated`  to speed up transition to newer methods.

test/freenet/support/MultiValueTableTest.java
Bombe, line 1: Oof, most of these tests are pretty horrible. 🙁 Coverage is pretty good, though, so they’ll do.

Bombe (changes requested):

src/freenet/clients/http/FProxyFetchTracker.java
```diff
@@ -37,7 +38,7 @@ public void shouldUpdate() {
 	private boolean requeue;
 
 	public FProxyFetchTracker(ClientContext context, FetchContext fctx, RequestClient rc) {
-		fetchers = new MultiValueTable<FreenetURI, FProxyFetchInProgress>();
+		this.fetchers = new MultiValueTable<>();
```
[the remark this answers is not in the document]
venfernand: I moved this where this field is declared.

src/freenet/clients/http/FProxyFetchTracker.java
```diff
@@ -92,18 +93,25 @@ public FProxyFetchWaiter makeWaiterForFetchInProgress(FreenetURI key,long maxSiz
...
+						if (fctx != null && !progress.fetchContextEquivalent(fctx)) {
+							continue;
+						}
+						if (logMINOR) {
+							Logger.minor(this, "Using " + progress);
+						}
 						return progress;
-					} else
-						if(logMINOR) Logger.minor(this, "Skipping "+progress);
+					} else if (logMINOR) {
```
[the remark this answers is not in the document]
venfernand: I tried to improve this method even more.

src/freenet/clients/http/ToadletContextImpl.java
```diff
@@ -12,16 +12,11 @@
...
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Locale;
-import java.util.StringJoiner;
-import java.util.TimeZone;
+import java.util.*;
```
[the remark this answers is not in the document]
venfernand: Expanded.

src/freenet/clients/http/ToadletContextImpl.java
```diff
@@ -366,28 +359,39 @@ public void setCookie(Cookie newCookie) {
...
-		boolean allowCaching; // For privacy reasons, only static
-							  // content may be cached
-		if (mTime == null) {
-			allowCaching = false;
-		} else {
-			allowCaching = true;
 		}
+
+		// content may be cached
+		// For privacy reasons, only static
```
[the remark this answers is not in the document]
venfernand: I think it was an outcome of some automatic refactoring or just regular blindness :)

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,269 @@
...
     }
 
     public void remove(K key) {
-		synchronized (table) {
-			table.remove(key);
-		}
+        removeAndGet(key);
+    }
+
+    public List<V> removeAndGet(K key) {
```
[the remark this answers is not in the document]
venfernand: I may be used when atomic remove and retrieve is required, but you are right, there are no real usages at the moment.
I may be added quite easily if this will be required.
I removed it.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,269 @@
...
-        public final void step() {
-            while (!current.hasMoreElements() && global.hasMoreElements())
-                current = global.nextElement().elements();
-        }
-
-        @Override
-        public final boolean hasMoreElements() {
-            return global.hasMoreElements(); // || current.hasMoreElements();
-        }
+    public Stream<V> values() {
```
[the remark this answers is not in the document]
venfernand: I tried to return a lazy view, which does not allocate that much.
But i agree that this breaks expectations of callers and it is better to return a simpler type from there.
I changed it to return `Collection<V>` and now it looks simpler.

test/freenet/support/MultiValueTableTest.java
venfernand: Do you mean that there is no need to check such trivial places?
Or those preparations of test collections  are not very clear?

venfernand:
I added some more fixes.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
+                list = new CopyOnWriteArrayList<>();
             }
-            v.addElement(value);
-        }
+            list.add(value);
+            return list;
+        });
+    }
+
+    public void putAll(K key, Collection<V> elements) {
```
bertm, line 79: `Collection<V>` put arguments should read `Collection<? extends V>` instead, so it becomes possible to put a `List<Foo>` into a `MultiValueTable<?, Object>`.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
-    	Vector<V> v;
-		synchronized (table) {
-			v = table.get(key);
-		}
-        return (v == null ?
-                new EmptyEnumeration<V>() :
-                v.elements());
+        List<V> elements = getAllAsList(key);
+        if (elements.isEmpty()) {
+            return Collections.emptyEnumeration();
```
bertm, line 132: This special case is unnecessary.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
-            return (v == null ?
-                    null :
-                    v.firstElement());
+        return getFirst(key);
+    }
+
+    /**
+     * Returns the first element for this key.
+     */
+    public V getFirst(K key) {
```
bertm, line 100: Now this is just a duplicate of `get`. Is there any real need for extending the public API with this method? If so, please consider deprecating the method it replaces.

src/freenet/support/MultiValueTable.java
venfernand: Pushed changes for `putAll()` and `from()` methods.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
 
     public void put(K key, V value) {
-        synchronized (table) {
-            Vector<V> v = table.get(key);
-            if (v == null) {
-                v = new Vector<V>(ies);
-                table.put(key, v);
+        this.table.compute(key, (k, list) -> {
+            if (list == null) {
+                list = new CopyOnWriteArrayList<>();
```
bertm, line 72: Why the CoW list? If we stick to unmodifiable lists instead (replace the list in the map, instead of altering its values), the implementation becomes a whole lot easier to reason about (once you've read a value lost, its contents remain stable forever) and you don't need to mess with wrapping in `Collections.unmodifiableList()` on every read.

src/freenet/support/MultiValueTable.java
venfernand: The only benefit of `Collections.emptyEnumeration()` is that it returns same object instance all the time.
Otherwise, yes, it can be simpler.
Will make it to be one-liner then.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
+            return Collections.enumeration(elements);
+        }
+    }
+
+    public List<V> getAllAsList(K key) {
+        List<V> list = this.table.get(key);
+        if (list == null || list.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return Collections.unmodifiableList(list);
```
bertm, line 143: This returns an unmodifiable view of the underlying list. This means that when a new entry is inserted or an existing entry deleted in the table for this key, the change is reflected in this returned list. That is highly unexpected (and if intended, should be thoroughly documented), most likely undesired, and probably  an unintended effect of choice of datastructure. See also my comment about the use of CoWList.

src/freenet/support/MultiValueTable.java
venfernand: New method looks more explicit for me.
Besides that there is no that big need and old method may be still used.
What variant do you prefer more?

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
+        } else {
+            return Collections.enumeration(elements);
+        }
+    }
+
+    public List<V> getAllAsList(K key) {
+        List<V> list = this.table.get(key);
+        if (list == null || list.isEmpty()) {
+            return Collections.emptyList();
+        }
```
bertm, line 142: This could be reduced to a simple oneliner:
```java
List<V> list = table.getOrDefault(key, Collections.emptyList());
```

src/freenet/clients/fcp/RequestStatusCache.java
```diff
@@ -192,17 +193,30 @@ public synchronized void updateStarted(String identifier, FreenetURI redirect) {
...
 
 	public synchronized CacheFetchResult getShadowBucket(FreenetURI key, boolean noFilter) {
-		Object[] downloads = downloadsByURI.getArray(key);
-		if(downloads == null) return null;
-		for(Object o : downloads) {
-			DownloadRequestStatus download = (DownloadRequestStatus) o;
+		List<DownloadRequestStatus> downloads = downloadsByURI.getAllAsList(key);
+		if (downloads == null) {
+			return null;
+		}
```
bertm, line 199: Unreachable code, `getAllAsList` never returns `null`

src/freenet/clients/http/FProxyFetchTracker.java
```diff
@@ -84,26 +83,44 @@ public FProxyFetchWaiter makeWaiterForFetchInProgress(FreenetURI key,long maxSiz
...
-						if(logMINOR) Logger.minor(this, "Found "+progress);
-						if(fctx != null && !progress.fetchContextEquivalent(fctx)) continue;
-						if(logMINOR) Logger.minor(this, "Using "+progress);
-						return progress;
-					} else
-						if(logMINOR) Logger.minor(this, "Skipping "+progress);
+			List<FProxyFetchInProgress> fetchList = fetchers.getAllAsList(key);
+			if (fetchList == null) {
+				return null;
+			}
```
bertm, line 103: Unreachable return; `fetchList` is never `null`.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
-            Vector<V> v = table.get(key);
-            if (v == null)
-                return null;
-            else {
-            	Object[] r = new Object[v.size()];
-                v.copyInto(r);
-                return r;
-            }
-        }
+        return getAllAsList(key).toArray();
```
bertm, line 180: This breaks the (undocumented) return value semantics of `getArray`: previously it would return `null` when no values are present, currently it returns an empty array.

Please consider simulating the old behaviour and documenting it in the Javadoc.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
-			return table.get(key);
-		}
+        return getAllAsList(key);
     }
 
+    /**
+     * Returns mapped value collection as array
+     *
+     * <b>Note:</b> this method is deprecated,
+     * please use other {@link #getAllAsList(Object)} method variant, which provides a typed {@link List} collection.
```
bertm, line 174: The `@deprecated` Javadoc tag is intended for this kind of deprecation notices, and provides better IDE integration. Suggestion:
```suggestion
     * @deprecated use {@link #getAllAsList(Object)} with {@link List#toArray()} to obtain the values as object array
```

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
+     * please use other {@link #getAllAsList(Object)})} method variant, which provides a typed {@link List} collection.
+     *
+     * @param key key mapping
+     */
+    @Deprecated
     public Object getSync(K key) {
-		synchronized (table) {
-			return table.get(key);
-		}
+        return getAllAsList(key);
```
bertm, line 167: This method used to return a `Vector` (or `null` if there was no value), now it always returns a `List`. There are many methods that are implemented by Vector but not by List, so this is a breaking change (even though this behaviour was undocumented, as usual).

Additionally, this now returns an unmodifiable list where it would previously allow modifications. We might as well remove this method right away - we cannot in any reasonable way simulate previous behaviour using this new implementation, and no sane plugin should use it anyways (given its entirely undocumented nature).

I checked the following plugins to be sure (listing by Plugin-Main-Class):
```
de.saces.fnplugins.Shoeshop.ShoeshopPlugin
de.saces.fnplugins.SiteToolPlugin.SiteToolPlugin
freenetsearch.Plugin
keepalive.Plugin
net.pterodactylus.sone.main.SonePlugin
org.freenetproject.freemail.FreemailPlugin
plugins.CENO.Client.CENOClient
plugins.floghelper.FlogHelper
plugins.flophelper.FlogHelper
plugins.FMS.FMSPlugin
plugins.Freereader.Freereader
plugins.JSTUN.JSTUN
plugins.KeyUtils.KeyUtilsPlugin
plugins.Library.Main
plugins.MDNSDiscovery.MDNSDiscovery
plugins.N2NChat.core.N2NChatPlugin
plugins.ShareLink.Plugin
plugins.Sharesite.Plugin
plugins.ShareWiki.Plugin
plugins.ThawIndexBrowser.ThawIndexBrowser
plugins.UPnP2.UPnP2
plugins.UPnP.UPnP
plugins.WebOfTrust.WebOfTrust
```

None of them use anything from MultiValueTable except:
```
freenet.support.MultiValueTable#<init>() // the no-args constructor
freenet.support.MultiValueTable#put(java.lang.Object,java.lang.Object)
```

@ArneBab @Bombe thoughts on removing this method?

src/freenet/support/MultiValueTable.java
bertm: `getFirst` make way more sense indeed (not that any of the other methods of this class make much sense, but that's all hindsight)

We should not have 2 non-deprecated methods doing the exact same thing, because then things get confusing for developers - which one should you use? I'm all for `getFirst`, but then let's deprecate `get` and refer to `getFirst` in the `@deprecated` notice.

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,260 @@
...
+    /**
+     * Returns table keys as {@link Enumeration}
+     *
+     * <b>Note:</b> this method is deprecated,
+     * please use other {@link #keySet()} method variant, which provides a {@link Set} of keys.
+     *
+     * @param key key mapping
+     */
+    @Deprecated
     public Enumeration<K> keys() {
```
bertm, line 221: This method does not have a `key` parameter even though one is specified in the Javadoc.

src/freenet/support/MultiValueTable.java
venfernand: `CopyOnWriteArrayList` was picked to allow concurrent iteration and access to be safe.
It also was a continuation of the previous implementation, where underlying data structure was modified in place.

Immutable lists are really better here, but these require some more code to write.
Stable collections may also prevent from various concurrent modification bus.

I pushed new changes containing immutable lists instead of `CopyOnWriteArrayList`.

src/freenet/support/MultiValueTable.java
venfernand: This was also true for the `Enumeration` objects, which were returned before.
I think that nobody ever used this like that and wanted any updates during iteration or later use.
I agree that it is really better to return "safe" and stable collections and iterators.

Now, when lists are truly immutable, then all returned values are also not modifiable.
Please see the new changes, it is probably better now.

src/freenet/support/MultiValueTable.java
venfernand: Added more deprecation comments there.

src/freenet/support/MultiValueTable.java
venfernand: Yes, the new implementation looks exactly like that.
When table holds immutable lists, then those may be returned as is.

src/freenet/support/MultiValueTable.java
venfernand: Restored the previous behavior and put notes into Javadoc

src/freenet/support/MultiValueTable.java
venfernand: I changed this back  to return `Vector`.
It may be removed at any time later.

src/freenet/support/MultiValueTable.java
venfernand: Fixed

src/freenet/clients/fcp/RequestStatusCache.java
venfernand: Removed

ArneBab:
@bertm is this ready to be merged, now?

@venfernand can you merge `next` into your branch and resolve the conlicts?

src/freenet/support/MultiValueTable.java
```diff
@@ -1,230 +1,308 @@
 package freenet.support;
 
-import java.util.Enumeration;
-import java.util.Hashtable;
-import java.util.NoSuchElementException;
-import java.util.Vector;
+import java.util.*;
```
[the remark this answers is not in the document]
ArneBab: I’ll create a separate PR for expanding star imports.

ArneBab:
merged after resolving conflicts. Waiting longer would only increase the amount of conflicts.

ArneBab:
@venfernand thank you for your PR!

PR #830, opened 2023-03-17
Remove usage of deprecated freenet.support.io.Closer

venfernand:
Removes usage of deprecated freenet.support.io.Closer in various places.
Replaces io streams handling with try-with-resources.


ArneBab:
This looks good to me. Thank you!

I’m sorry that it took so long to review: while the changes are all clear, reviewing such a huge change in github was pretty undoable. I got it done now by checking changes directly in Emacs (that makes it much more obvious what changed).

PR #833, opened 2023-04-19
UX  show plugins before source selection

ArneBab:


ArneBab:
Bombe will use a better solution.

PR #834, opened 2023-04-19
Use logging instead of backtrace for IPv6-addresses when unsupported

ArneBab:
Throwing backtraces is needlessly expensive and spams the log with no additional gain.


bertm (approved):
Seems reasonable.

ArneBab:
@bertm merged — thank you!

PR #835, opened 2023-04-19
fix: invalid max store size showed bytes with GiB suffix

ArneBab:


ArneBab:
reviewed face-to-face by @Bombe  on sunday

PR #838, opened 2023-05-23
Sort addresses for handshake by host→ipv6→ipv4

ArneBab:


Bombe:
Oof, there’s no tests for anything, there’s commented code checked in, there’s TODO comments added, this doesn’t really do anything to improve Fred’s code… 😞

PR #839, opened 2023-05-23
 Sort addresses for handshake by host→ipv6→ipv4

ArneBab:


gits7r:
This is good work, a much needed PR - the current behavior must be fixed, our network / peer connectivity code behaves in a way that is against the rfc's for dual stacks, possibly marking an unknown number of peers 'unreachable' as false positive. For example seednodes almost only see (98%) IPv4 seed traffic, even they are available on IPv6 as well, because their IPv6 address is not even tried. This will make it impossible for a new connecting client that is IPv6 only to bootstrap in OpenNet.

Thank you @ArneBab for this epic work. Since it needs review, I'll sum up here the intended behavior:

Since `physical.udp` can contain the following address *classes* (and each class can have one or more addresses per each class): hostname(s), raw IPv6 address(es), raw IPv4 address(es), we should have the following behavior:

- prefer HOSTNAMEs over raw IPv6 or raw IPv4 addresses. If multiple HOSTNAMEs are found, try them in order;
- further, resolve HOSTNAMEs and and if both AAAA (IPv6) and A (IPv4) records are found, prioritize IPv6 (AAAA), otherwise just use A records (IPv4). If more addresses are returned for both AAAA and A records, try them in order (first all IPv6, then all IPv4);
- prefer raw IPv6 addresses over raw IPv4 addresses , try them in order in case multiple from each address class giving priority to  IPv6 addresses.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -1,29 +1,71 @@
...
 		byte[] bytes0 = arg0.getAddress();
 		byte[] bytes1 = arg1.getAddress();
+        // prefer IPv6 over IPv4
 		if(bytes0.length > bytes1.length)
-			return -1; // IPv6 < IPv4. => prefer IPv6 over IPv4
+			return -1;
 		if(bytes1.length > bytes0.length)
-			return 1; // IPv4 < IPv6.
+			return 1;
+		// prefer reachable over unreachable. TODO: This actually pings all advertised ip addresses. Is that OK?
```
gits7r: prefer reachable over unreachable is OK.  What do you mean by pings all advertised IP addresses? ping as ping (ICMP) or freenet / fred specific UDP connection attempt?

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -21,26 +21,19 @@ public class InetAddressIpv6FirstComparator implements Comparator<InetAddress> {
...
 		}
-		byte[] bytes0 = arg0.getAddress();
-		byte[] bytes1 = arg1.getAddress();
-        // prefer IPv6 over IPv4
-		if(bytes0.length > bytes1.length)
-			return -1;
-		if(bytes1.length > bytes0.length)
-			return 1;
-		// prefer reachable over unreachable. TODO: This actually pings all advertised ip addresses. Is that OK?
+		// prefer reachable over unreachable addresses. This is usually a ping. TODO: This actually pings all advertised ip addresses. Is that OK? Do we need a maximum number of accepted addresses to prevent abuse as DDoS?
```
gits7r, line 44: I don't see this a big avenue for DDoS but we might but a global limit on the physical.udp record itself that is large enough to cover at least X addresses from each address class, but I am not sure what limit this should be as hostnames can be quite large strings and still be valid addresses.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -66,7 +59,22 @@ public int compare(InetAddress arg0, InetAddress arg1) {
 		} else if (!reachable0 && reachable1) {
 			return 1;
 		}
-		// By hash code first. Works really fast for IPv4.
+		// among routable addresses prefer LAN addresses over global addresses, because they should stay within VPNs
```
gits7r, line 70: This is OK and also good not only for VPN use case, also for LANs.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -40,15 +42,15 @@ public int compare(InetAddress arg0, InetAddress arg1) {
 		Boolean reachable1 = reachabilityCache.get(b);
 		if (reachable0 == null) {
 			try {
-				reachable0 = arg0.isReachable(300);
+				reachable0 = arg0.isReachable((int) DEFAULT_MAX_PING_TIME);
```
gits7r, line 51: about the default max ping time: maybe needs to be tweaked, for example let's say I am connected to a VPN very far from my location, and I have 10x times more latency for the peers in my VPN (which are preferred, local LAN addresses) than public addresses closer to my country or within my country. Won't this make fred backoff my VPN peers and prefer my other peers because of lower latency?

src/freenet/node/PeerNode.java
```diff
@@ -601,6 +602,9 @@ public PeerNode(SimpleFieldSet fs, Node node2, NodeCrypto crypto, boolean fromLo
 			Logger.normal(this, "No IP addresses found for identity '" + identityAsBase64String + "', possibly at location '" + location + ": " + userToString());
 			detectedPeer = null;
 		} else {
+			nominalPeer.sort(Peer.PEER_COMPARATOR);
+			// TODO	this throws away all valid addresses but the first, without checking whether they can connect. Need to try a later one if connection fails.
```
Thynix, line 606: Is my understanding correct that this isn't a change to behavior, just a note about it?

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -0,0 +1,92 @@
...
+
+	// need a cache for reachability to avoid doing NlogN issReachable checks in worst case.
+	public LRUCache<Integer, Boolean> reachabilityCache = new LRUCache<>(1000, 300000);
+
+	public static final InetAddressIpv6FirstComparator COMPARATOR =
+		new InetAddressIpv6FirstComparator();
+
+	@Override
+	public int compare(InetAddress arg0, InetAddress arg1) {
+		if(arg0 == arg1) return 0;
```
Thynix, line 25: Is this a reference comparison? Or it is implicitly calling Equals()?

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -0,0 +1,92 @@
...
+			return 1;
+		}
+		// prefer reachable over unreachable addresses. This is usually a ping. TODO: This actually pings all advertised ip addresses. Is that OK? Do we need a maximum number of accepted addresses to prevent abuse as DDoS? 
+		int a = arg0.hashCode();
+		int b = arg1.hashCode();
+		Boolean reachable0 = reachabilityCache.get(a);
+		Boolean reachable1 = reachabilityCache.get(b);
+		if (reachable0 == null) {
+			try {
+				reachable0 = arg0.isReachable((int) DEFAULT_MAX_PING_TIME);
```
Thynix, line 51: Is it reasonable to not allow configuring this ping time? I'm not sure about the usability and security implications here, but wanted to note that this is hardcoded.

Thynix (approved):

src/freenet/support/io/InetAddressIpv6FirstComparator.java
ArneBab: PING as ICMP.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
ArneBab: I’d push that to a later step if it is needed. Usually the default max ping time should be plenty for a real ping.

src/freenet/io/comm/Peer.java
```diff
@@ -279,4 +283,28 @@ public boolean isIPv6(boolean defaultValue) {
...
+					return 1;
+				} else if (hasHostnameP0) {
+					return 0;
+				}
+				if (isIpv6P0 && !isIpv6P1) {
+					return -1;
+				} else if (!isIpv6P0 && isIpv6P1) {
+					return 1;
+				}
+				return InetAddressIpv6FirstComparator.COMPARATOR
```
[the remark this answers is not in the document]
ArneBab: Yes: this passes on the addresses, not the peers, to the specialized Ipv6 first comparator.

src/freenet/node/PeerNode.java
ArneBab: Not completely: this sorts the addresses, so it gets the best address instead of the first. Usually that will be the hostname.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
ArneBab: it looks like a reference comparison, yes … damn … I’ll fix that during merging. Thank you!

src/freenet/support/io/InetAddressIpv6FirstComparator.java
ArneBab: I think it’s reasonable. I hope there won’t be a need to configure it, but if that arises, we can add that.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
ArneBab: I’ll have to look into that, but I don’t think we have to fix that before the next release.

PR #841, opened 2023-05-25
Change logging for "too many excluded sub-arrays" from normal to minor.

ArneBab:


src/freenet/support/SectoredRandomGrabArray.java
```diff
@@ -173,7 +173,8 @@ private RandomGrabArrayItem removeRandomLimited(
 				Logger.error(this, "Slot "+x+" is null for client "+grabClients[x]);
 				excluded++;
 				if(excluded > MAX_EXCLUDED) {
-					Logger.normal(this, "Too many sub-arrays are entirely excluded on "+this+" length = "+grabArrays.length, new Exception("error"));
+					if(logMINOR)
```
Bombe, line 176: Please stop doing this. Optimizations of this kind haven’t been necessary for over 10 years now.

src/freenet/support/SectoredRandomGrabArray.java
```diff
@@ -182,7 +183,8 @@ private RandomGrabArrayItem removeRandomLimited(
 			if(excludeTime > 0) {
 				excluded++;
 				if(excluded > MAX_EXCLUDED) {
-					Logger.normal(this, "Too many sub-arrays are entirely excluded on "+this+" length = "+grabArrays.length, new Exception("error"));
+					if (logMINOR)
+                        Logger.minor(this, "Too many sub-arrays are entirely excluded on "+this+" length = "+grabArrays.length, new Exception("error"));
```
Bombe, line 187: You’re using whitespace that does not match the surrounding area here.

src/freenet/support/SectoredRandomGrabArray.java
ArneBab: This was observed being run 100k times — while 100 times creating the backtrace is OK, 100k backtraces (`new Exception("error")`) actually cause massive load.

The reason why if(logMINOR) is useful here is not that it prevets running Logger.minor(…), but that it prevents the creation of the backtrace which will be run before Logger.minor(…) decides whether to log.

src/freenet/support/SectoredRandomGrabArray.java
ArneBab: yikes, thank you. That’s the tab vs. 4 spaces …

Bombe: Yeah, as usual… IDEA can auto-detect and -adapt to existing whitespace conditions! 😄

src/freenet/support/SectoredRandomGrabArray.java
Bombe: I know exactly what that line does. 🙂

Nevertheless, creating 100000 exceptions takes less than 36 ms on my MacBook. I’d much rather simply get rid of creating the exception than littering the source code with those `if` blocks.

And apparently that exception doesn’t even help because otherwise you’d simply fix the underlying issue, right? 😁

Bombe (approved):

PR #842, opened 2023-05-25
when using a USK with -1 as version, also find version 0.

ArneBab:


Bombe:
Again, I am very much disgusted by the lack of any tests. 🙂

ArneBab:
> Again, I am very much disgusted by the lack of any tests. 🙂

I’ve been trying to add a test, but this area is awfully ugly for that. I fear that adding a test will require quite some refactoring. How do you test a callback without mocking everything or doing a full system test?

Bombe:
That depends a bit on a lot of different factors… who creates the callback, who calls the callback, does the callback contain code that needs to be tested… 🙂

ArneBab:
I’m not happy with the test, but this is fixing an actual bug: when trying to access `USK@.../-1/` for a USK that is only available on `USK@.../0/` the access never finishes without this fix. This just hit ChampiZen for a link to a new Sharesite.

PR #843, opened 2023-06-23
Fix NPE When a SimpleFieldSet’s “subsets” Are Uninitialized

Bombe:
This fixes https://freenet.mantishub.io/view.php?id=7197.

(I also sneakily updated the Hamcrest dependency to 2.2 because it has a couple more cool matchers and it doesn't need the -core/-library split anymore.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -244,8 +244,7 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile "org.mockito:mockito-core:1.9.5"
-    testCompile "org.hamcrest:hamcrest-library:1.3"
-    testCompile "org.hamcrest:hamcrest-core:1.3"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -244,8 +244,7 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile "org.mockito:mockito-core:1.9.5"
-    testCompile "org.hamcrest:hamcrest-library:1.3"
-    testCompile "org.hamcrest:hamcrest-core:1.3"
+    testCompile "org.hamcrest:hamcrest:2.2"
```
ArneBab, line 247: since we do not ship that (not needed for runtime), this is uncritical. :+1:

ArneBab:
Looks good to me! :heart:

PR #849, opened 2023-08-24
Add generate media site and Opennet SeedNodes tats to the default bookmarks

ArneBab:


ArneBab:
Included in re-ordering

PR #851, opened 2023-08-29
allow 127.0.0.1 in bindTo

ArneBab:
For a Tor bridge that excludes all other IPs from the noderef to avoid spilling the real IP


ArneBab:
TODO: add config option `[ ] only allow operation as localhost, for example for tor or i2p`.

src/freenet/node/NodeIPPortDetector.java
```diff
@@ -58,7 +58,7 @@ public void shouldUpdate(){
 	 */
 	FreenetInetAddress[] detectPrimaryIPAddress() {
 		FreenetInetAddress addr = crypto.getBindTo();
-		if(addr.isRealInternetAddress(false, true, false)) {
+		if(addr.isRealInternetAddress(false, true, true)) {
```
ArneBab, line 61: TODO: add config option [ ] only allow operation as localhost, for example for tor or i2p.

ArneBab:
I now added a config option for this.

src/freenet/node/NodeIPDetector.java
```diff
@@ -526,6 +530,27 @@ public void set(String val) throws InvalidConfigValueException {
...
+		nodeConfig.register("allowBindToLocalhost", false, sortOrder++, true, false, "Node.allowBindToLocalhost", "Node.allowBindToLocalhostLong", new BooleanCallback() {
+
+			@Override
+			public Boolean get() {
+				return allowBindToLocalhost;
+			}
+
+			@Override
+			public void set(Boolean val) throws NodeNeedRestartException {
+				if (allowBindToLocalhost != val) {
```
Bombe, line 545: This will throw a `NullPointerException` if `val` is `null`.

src/freenet/node/NodeIPDetector.java
```diff
@@ -526,6 +530,27 @@ public void set(String val) throws InvalidConfigValueException {
...
+					synchronized(this) {
```
Bombe, line 546: This synchronization is broken, as long as there are places where `allowBindToLocalhost` is accessed without synchronization, such as in the `get()` method.

src/freenet/support/SortedIntSet.java
Bombe, line 1: This change belongs to a different pull request, doesn’t it?

Bombe (changes requested):

gits7r:
Binding to localhost already works for darknet node, at least I can decode 127.0.0.1:port from my base64 encoded physical.udp line, s**o this feature is only related to opennet mode**. In this case, shouldn't we register config as `node.opennet.allowBindToLocalhost` instead of `Node.allowBindToLocalhost` ?

And add comment that it should probably (not mandatory) be used along with `node.opennet.includeLocalAddressesInNoderefs` and `node.opennet.alwaysAllowLocalAddresses`, but those latter two config parameters will also work for other Local Addresses even if Bind to Localhost is set to false.

A sane description would be for this config param (of please someone make it shorter if possible, I'm not too good at shorting phrases):

"_In Opennet mode it would be impossible for someone else to connect to your node using the localhost loopback address. This feature is needed if you use other applications that provides a reachable address of any sort and proxies that to localhost. Currently this node has no way to guess those kind of addresses and add them to your physical.udp line, and there might be no other peers that can reach certain type of addresses. Set this to "true" only if you know what you are doing_".

It would be really neat if this would make it into the next. Covers at least few use cases reported on FMS.

src/freenet/config/ConfigCallback.java
```diff
@@ -3,7 +3,54 @@
...
 public abstract class ConfigCallback<T> {
+
+	/**
+	 * Create a config callback from lambdas. Needs to be specialized
+	 * in Subtypes of ConfigCallback to match the nodeConfig.register API.
+	 *
+	 * @param set accepts the new value and return an Exception or null (if no exception)
+	 */
+	public static <T> ConfigCallback<T> from(Supplier<T> get, Function<T, Exception> set) {
+		return new ConfigCallback<T>() {
```
Bombe, line 19: Did somebody forget to write a test for this? 🤣

src/freenet/config/ConfigCallback.java
```diff
@@ -3,7 +3,54 @@
...
+
 public abstract class ConfigCallback<T> {
+
+	/**
+	 * Create a config callback from lambdas. Needs to be specialized
+	 * in Subtypes of ConfigCallback to match the nodeConfig.register API.
+	 *
+	 * @param set accepts the new value and return an Exception or null (if no exception)
+	 */
+	public static <T> ConfigCallback<T> from(Supplier<T> get, Function<T, Exception> set) {
```
Bombe, line 18: I get why you’re doing this but functions returning exceptions rub me all kinds of wrong ways. I’d prefer something like this:

```java
public interface ConfigConsumer<T> {
    void accept(T value) throws InvalidConfigValueException, NodeNeedRestartException;
}
```

This way we don’t need casting in the `set()` method, and we don’t have to return anything from the `Consumer`.

src/freenet/node/NodeIPDetector.java
```diff
@@ -63,7 +65,7 @@ public void shouldUpdate(){
 	/** Explicit forced IP address in string form because we want to keep it even if it's invalid and therefore unused */
 	String overrideIPAddressString;
 	/** config: allow bindTo localhost? */
-	boolean allowBindToLocalhost;
+	Boolean allowBindToLocalhost;
```
Bombe, line 68: Is this actually necessary? The compiler should take care of all the boxing/unboxing…

src/freenet/support/api/BooleanCallback.java
```diff
@@ -3,11 +3,21 @@
...
+
 import freenet.config.ConfigCallback;
 
 /**
  * A callback to be called when a config value of integer type changes.
  * Also reports the current value.
  */
 public abstract class BooleanCallback extends ConfigCallback<Boolean> {
+  public static BooleanCallback fromBoolean(Supplier<Boolean> get, Function<Boolean, Exception> set) {
+    return (BooleanCallback) ConfigCallback.from(get, set);
```
Bombe, line 18: Uhm… this is not how casting works. `ConfigCallback.from` returns a `ConfigCallback` object, no amount of casting will magically turn it into a `BooleanCallback`.

You know, you would have caught that if you had written tests for this. 😠

Bombe (changes requested):

src/freenet/config/ConfigCallback.java
ArneBab: Yes. That was „I can do this quickly before going to bed“ :sweat:

src/freenet/config/ConfigCallback.java
ArneBab: that’s much nicer — thank you!

src/freenet/node/NodeIPDetector.java
ArneBab: No it isn’t — thank you!

ArneBab:
accidentally merged, then undid that merging

ArneBab:
I didn’t find a way to actually undo the closing here, so please check here: https://github.com/hyphanet/fred/pull/951

PR #852, opened 2023-08-31
Problem with maxDatastoreSize

ThePlantEater:
### Situation
- 500GB disk
- 100GB datastore
- 80GB free space

### Problem 1
If you know try to set your datastore to 130GB an error appears... because because `DatastoreUtil.maxDatastoreSize()` will return  only the 80GB left (https://github.com/hyphanet/fred/blob/fcc32a21cfb962a03b731edd3a14d6d0b2e0860c/src/freenet/support/io/DatastoreUtil.java#L44)
**Solution:**  use the current datastore size + the free space as max

### Problem 2
`DatastoreUtil.maxDatastoreSize()` only checks the free space where hyphanet is executed not where the datastore is. If the datastore is on another disk this lead also to a problem.
**Solution:** use the config value for the datastore

#### Information
- please squash
- please test this

### DIFF
--- a/src/freenet/clients/http/FirstTimeWizardNewToadlet.java
+++ b/src/freenet/clients/http/FirstTimeWizardNewToadlet.java
@@ -222,7 +222,7 @@ public class FirstTimeWizardNewToadlet extends WebTemplateToadlet {
                 if (storageLimit < MIN_STORAGE_LIMIT) { // min store size + 10% for client cache + 10% for slashdot cache
                     errors.put("storageLimitError", NodeL10n.getBase().getString("Node.invalidMinStoreSizeWithCaches"));
                 } else {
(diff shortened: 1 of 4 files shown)


ArneBab:
Thank you for tracking this down! The code looks good for me (but I have to test it).

ArneBab:
Upon discussion with @Bombe we found a problem: if the store has to be resized, it likely needs the current size *and* the new size during the resizing operation, so the maxSize is actually correct: it might otherwise run out of disk space during resizing.

Therefore I fear the only fix we can do is to add this as information in the configuration items description.

ArneBab:
@ThePlantEater thank you for closing! Would you like to change the description of the option so that people know what to expect?

PR #853, opened 2023-09-03
Switch swiss flag to civil and state ensign — thanks to Percept0r@NY

ArneBab:
With this change the swiss flag matches the other flags in proportions, so it’s size matches the sizes of other flags in the "Connections to Strangers" view.

thanks to Percept0r@NYZkOs7eQNLmiqWzm7YY8hFt-c7mUe3V~q6idbEsZjE for fixing the issue!


ArneBab:
Wikipedia states that while this isn’t the Swiss national flag, it is actually used as the "civil and state ensign", so I think for this it is matching: https://en.wikipedia.org/wiki/Flag_of_Switzerland

In the FMS thread there was also the flag of Nepal with white background. For that I asked for a version with transparent background, because it would be fully wrong with white bg (and look strange on a dark scheme).

Bombe:
Okay, okay, okay, go for it! 😄

PR #854, opened 2023-09-04
✨ Allow Registering of Ignored Options

Bombe:
The whole reason for this is that Fred logs messages at ERROR if a config file contains an option it does not know about. As the plan is to remove a number of options, not having these values show up in log files is a definite plus.


ArneBab:
Looks great! 🎆

ArneBab (approved):

PR #857, opened 2023-09-09
Remote HTTP(S) Download for Official Plugins

Bombe:
This should resolve https://freenet.mantishub.io/view.php?id=7198.

I think I found all places where we’re allowing the user to do this. If anybody knows of any other places, now would be a good time to say something! 😄


src/freenet/clients/http/PproxyToadlet.java
```diff
@@ -134,12 +134,11 @@ public void handleMethodPOST(URI uri, final HTTPRequest request, ToadletContext
 			
 			if (request.isPartSet("submit-official")) {
 				final String pluginName = request.getPartAsStringFailsafe("plugin-name", 40);
-				final String pluginSource = request.getPartAsStringFailsafe("pluginSource", 10);
-				
+
 				node.executor.execute(new Runnable() {
 					@Override
 					public void run() {
-						pm.startPluginOfficial(pluginName, true, true, "https".equals(pluginSource));
```
ArneBab, line 142: I fear that startPluginOfficial is (used as) public API — can you add a deprecated compatibility method that just ignores the two force options and calls the real one without them?

ArneBab:
Thank you for that great cleanup! The page is much nicer now and getting rid of the config option that effectively said "always connect a clearnet server" is a big step towards not suggesting that people shoot themselves in the foot.

2023-09-09-135219_943x913_scrot

ArneBab:
Looks good — thank you! :+1:

ArneBab (approved):

PR #858, opened 2023-09-09
Set the new load management option to ignored.

ArneBab:
If this was active, connecting was impossible, and NLM has been inoperable and unused for years.

Currently not removing the routing code, because that’s more risky and there are parts that might be useful to salvage (like tempered preference for nodes with better ping times).

### DIFF
--- a/src/freenet/node/NodeStats.java
+++ b/src/freenet/node/NodeStats.java
@@ -319,8 +319,8 @@ public class NodeStats implements Persistable, BlockTimeCallback {
 	public final DecayingKeyspaceAverage avgClientCacheSSKSuccess;
 	public final DecayingKeyspaceAverage avgStoreSSKSuccess;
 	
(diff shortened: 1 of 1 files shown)


src/freenet/node/NodeStats.java
```diff
@@ -319,8 +319,8 @@ public void shouldUpdate(){
 	public final DecayingKeyspaceAverage avgClientCacheSSKSuccess;
 	public final DecayingKeyspaceAverage avgStoreSSKSuccess;
 	
-	private volatile boolean enableNewLoadManagementRT;
-	private volatile boolean enableNewLoadManagementBulk;
+	private final boolean enableNewLoadManagementRT = false;
+	private final boolean enableNewLoadManagementBulk = false;
```
ArneBab, line 323: This is the minimally invasive change

Bombe (approved):
I like disabling stuff! 🥳

PR #860, opened 2023-09-10
Move Library to advanced plugins because new users tend to get lost

ArneBab:
Before moving it back to the default list of plugins, we need to figure out a good user experience for it. At the moment either it does not work or it floods users with bad results.


Bombe (approved):

ArneBab:
Thank you for your review!

PR #861, opened 2023-09-10
Move UPnP2 to normal plugins. It does not seem broken, but UPnP does

ArneBab:
There were several reports from users that UPnP did not start. We should give UPnP2 more testing so we can enable it by default.


Bombe:
Are we going to replace UPnP at some point?

ArneBab:
I think we should replace UPnP, yes. It’s just so far outside areas I’m working in at the moment that I did not get to checking and testing it well enough. It’s sadly one of the essential plugins for connectivity in a NAT.

Bombe (approved):

PR #862, opened 2023-09-10
Fix merge error.

ArneBab:
This is an embarrassing one … it looks like our debian package build doesn’t run the tests yet.


Bombe (approved):

Bombe:
> This is an embarrassing one … it looks like our debian package build doesn’t run the tests yet.

That’s probably because our `jar` task does not depend on the `test` task.

PR #863, opened 2023-09-10
Fix gzip compressor instead of test

ArneBab:


src/freenet/support/compress/GzipCompressor.java
```diff
@@ -33,6 +37,17 @@ public Bucket compress(Bucket data, BucketFactory bf, long maxReadLength, long m
 			Closer.close(is);
 			Closer.close(os);
 		}
+		// force OS byte to 0 regardless of Java version (java 16 changed to setting 255 which would break hashes)
+		if (output instanceof ArrayBucket) {
```
Bombe, line 41: I’d very much prefer something that will never fail, for whatever reasons… how about wrapping the `OutputStream` in another `OutputStream` which will override the 10th byte only? 🤔

ArneBab: I didn’t see how to wrap the stream, so changed the fallback to copy the bucket, fixing the OS byte.

Bombe: I imagined something like this: https://github.com/Bombe/fred/commit/5250b03abad4db7c536da3286cf167b1b7015370.

ArneBab: @Bombe I now switched to your implementation and removed the meddling at the wrong abstraction. Thank you!

Is is good to go with this?

Bombe (approved):

ArneBab:
@Bombe thank you! This was the fix that still needed to go in, because the bug would have otherwise caused more and more inserts that would not be reproducible on lower Java versions or fixed hyphanet. Though I think that the effect is limited, because few files are inserted as gzip nowadays (LZMA usually wins).

PR #864, opened 2023-09-10
Use assertArrayEquals and add proper error messages to asserts

ArneBab:


test/freenet/crypt/CTRBlockCipherTest.java
```diff
@@ -226,7 +226,7 @@ private void checkNISTRandomLength(int bits, byte[] key, byte[] iv,
 				}
 				c.doFinal(plaintext, 0, plaintext.length - inputPtr, output,
 						outputPtr);
-				assertTrue(Arrays.equals(output, ciphertext));
+				assertArrayEquals(output, ciphertext);
```
Bombe, line 229: I believe in all those assert methods the first parameter is the _expected_ one, and the second one is the actual. For the actual comparison this does not matter, but the error message will have the compared objects switched which _will_ confuse people.

However, if you’re already fiddling with the asserts, I’d much rather you’d switch them to using the Hamcrest matchers, because that will immediately remove any confusion that might possibly arise from the parameter order, like so:

```suggestion
				assertThat(output, equalTo(ciphertext));
```

(And yes, the matcher returned by `equalTo` knows how to compare arrays correctly. 🙂)

ArneBab: Switched to hamcrest — thank you!

test/freenet/crypt/CryptByteBufferTest.java
```diff
@@ -185,7 +190,7 @@ public void testSuccessfulRoundTripInPlaceOffset() throws GeneralSecurityExcepti
             byte[] copyBuffer = buffer.clone();
             System.arraycopy(originalPlaintext, 0, buffer, header, originalPlaintext.length);
             crypt.encrypt(buffer, footer, originalPlaintext.length);
-            assertTrue(!Arrays.equals(buffer, copyBuffer));
+            assertThat(buffer, Matchers.not(equalTo(copyBuffer)));
```
Bombe, line 193: The matchers look best when they are all statically imported, so it becomes:
```suggestion
            assertThat(buffer, not(equalTo(copyBuffer)));
```

test/freenet/crypt/CryptByteBufferTest.java
```diff
@@ -259,14 +264,14 @@ public void testSuccessfulRoundTripByteArrayReset() throws GeneralSecurityExcept
...
-                assertTrue("CryptByteBufferType: "+type.name(), plain.equals(decipheredtext1));
-                assertTrue("CryptByteBufferType: "+type.name(), plain.equals(decipheredtext2));
-                assertTrue("CryptByteBufferType: "+type.name(), plain.equals(decipheredtext3));
+                assertThat("CryptByteBufferType: "+type.name(), plain, equalTo(decipheredtext1));
+                assertThat("CryptByteBufferType: "+type.name(), plain, equalTo(decipheredtext2));
+                assertThat("CryptByteBufferType: "+type.name(), plain, equalTo(decipheredtext3));
             }
     }
     
     private void assertNotEquals(Object o1, Object o2) {
```
bertm, line 273: Please inline this method, it's only used in a handful of places and from its signature it's unclear how it handles arrays.

Bombe: Inline which method? 😕

ArneBab: @bertm do you mean assertNotEquals?

ArneBab:
@bertm Inlined the method.

@Bombe imported Matchers.not statically.

Bombe:
> @Bombe imported Matchers.not statically.

Much nicer, isn’t it? 😁

test/freenet/crypt/CryptByteBufferTest.java
```diff
@@ -160,10 +164,10 @@ public void testSuccessfulRoundTripInPlace() throws GeneralSecurityException {
             byte[] buffer = Hex.decode(plainText[i]);
             byte[] plaintextCopy = buffer.clone();
             crypt.encrypt(buffer, 0, buffer.length);
-            assertTrue(!Arrays.equals(buffer, plaintextCopy));
+            assertThat((Object) buffer, not(equalTo(plaintextCopy)));
```
Bombe, line 167: That cast is not necessary.

test/freenet/crypt/CryptByteBufferTest.java
```diff
@@ -251,23 +255,19 @@ public void testSuccessfulRoundTripByteArrayReset() throws GeneralSecurityExcept
                     // Once we have initialised the cipher, it is treated as a stream.
                     // Repeated encryption of the same data will return different ciphertext, 
                     // as it is treated as a later point in the stream.
-                    assertNotEquals(ciphertext1, ciphertext2);
-                    assertNotEquals(ciphertext1, ciphertext3);
-                    assertNotEquals(ciphertext2, ciphertext3);
+                    assertThat((Object) ciphertext1, not(equalTo(ciphertext2)));
```
Bombe, line 258: And neither are these casts. 🙂

test/freenet/crypt/CryptByteBufferTest.java
```diff
@@ -327,14 +327,14 @@ public void testEncryptByteBufferToByteBuffer() throws GeneralSecurityException
             crypt.encrypt(plaintext, ciphertext);
             assertEquals(plaintext.position(), header+origPlaintext.length);
             assertEquals(ciphertext.position(), header+origPlaintext.length);
-            assertTrue(Arrays.equals(buf, cloneBuf)); // Plaintext not modified.
+            assertThat(buf, equalTo(cloneBuf)); // Plaintext not modified.
             plaintext.position(header);
             ciphertext.position(header);
-            assertTrue(!Arrays.equals(ciphertextBuf, copyCiphertextBuf));
+            assertThat((Object) ciphertextBuf, not(equalTo(copyCiphertextBuf)));
```
Bombe, line 333: More unnecessary casting.

Bombe (changes requested):

ArneBab:
@bertm @Bombe Removed the unnecessary casts and merged — thank you for your reviews!

PR #866, opened 2023-10-08
Replace GregorianCalendar with java.time.OffsetDateTime in CurrentTimeUTC

venfernand:
Now `CurrentTimeUTC` uses newer `java.time` API.
This change also removes unnecessary synchronization on the `GregorianCalendar` object.

### DIFF
--- a/src/freenet/support/CurrentTimeUTC.java
+++ b/src/freenet/support/CurrentTimeUTC.java
@@ -21,9 +21,10 @@ public class CurrentTimeUTC {
     /**
      * Get the current time in milliseconds.
      *
-     * In the current implementation, this just returns System.currentTimeMilis(). You should however use CurrenTimeUTC.getInMillis() instead because
-     * the JavaDoc of System.currentTimeMilis() does not explicitly state what time zone it returns. Therefore, by using this wrapper function, your code
-     * clearly states that it uses UTC time.
+     * In the current implementation, this just returns {@link System#currentTimeMillis()}.
+     * You should however use CurrentTimeUTC#getInMillis() instead because
+     * the JavaDoc of {@link System#currentTimeMillis()} does not explicitly state what time zone it returns.
+     * Therefore, by using this wrapper function, your code clearly states that it uses UTC time.
      */
     public static long getInMillis() {
         return System.currentTimeMillis();
--- a/src/freenet/support/CurrentTimeUTC.java
+++ b/src/freenet/support/CurrentTimeUTC.java
@@ -15,7 +15,7 @@ import java.util.Date;
 public class CurrentTimeUTC {
 
     public static Date get() {
-        return Date.from(nowUtc().toInstant());
+        return new Date();
     }
 
     /**
--- a/src/freenet/support/CurrentTimeUTC.java
+++ b/src/freenet/support/CurrentTimeUTC.java
@@ -3,9 +3,8 @@
  * http://www.gnu.org/ for further details of the GPL. */
 package freenet.support;
 
-import java.time.OffsetDateTime;
+import java.time.LocalDate;
 import java.time.ZoneOffset;
-import java.time.temporal.ChronoField;
 import java.util.Date;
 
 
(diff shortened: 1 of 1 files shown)


src/freenet/support/CurrentTimeUTC.java
```diff
@@ -3,61 +3,49 @@
...
+
+    /**
+     * Get the current time in milliseconds.
+     *
+     * In the current implementation, this just returns {@link System#currentTimeMillis()}.
+     * You should however use CurrentTimeUTC#getInMillis() instead because
+     * the JavaDoc of {@link System#currentTimeMillis()} does not explicitly state what time zone it returns.
+     * Therefore, by using this wrapper function, your code clearly states that it uses UTC time.
+     */
+    public static long getInMillis() {
```
Bombe, line 29: Oh my, are we using this function anywhere? The documentation is horribly incorrect.

Bombe: Oof, the whole class is a serious clusterfuck. We should deprecate it right now (and nuke it from orbit, it’s the only way to be sure).

ArneBab: @Bombe this sounds like “this is a good change and it would be good to do more changes” — am I right in that understanding of your comment?

Nuking the class from orbit sounds dangerous, because it is available to plugins (our usual problem). I like getting rid of unnecessary synchronization.

src/freenet/support/CurrentTimeUTC.java
```diff
@@ -3,61 +3,49 @@
...
+
+    public static int getMonth() {
+        // Previously, this method used java.util.GregorianCalendar and Calendar.MONTH, which is zero-based.
+        // Newer java.time API returns months values starting with one,
+        // and subtraction is required to preserve backward compatibility.
+        return nowUtc().get(ChronoField.MONTH_OF_YEAR) - 1;
+    }
+
+    public static int getDayOfMonth() {
+        return nowUtc().get(ChronoField.DAY_OF_MONTH);
```
bertm, line 45: Finally, `LocalDate.now(ZoneOffset.UTC).getDayOfMonth()`.

src/freenet/support/CurrentTimeUTC.java
ArneBab: At work we have the policy that every deprecation needs a comment people can copy paste to replace the usage of the deprecated method. I would suggest doing that here, too.

We cannot remove this until there isn’t a single use in plugins.

(besides: welcome back, @bertm !)

src/freenet/support/CurrentTimeUTC.java
```diff
@@ -3,61 +3,49 @@
...
-
-	public static int getDayOfMonth() {
-		synchronized(mCalendar) {
-			mCalendar.setTimeInMillis(System.currentTimeMillis());
-			return mCalendar.get(Calendar.DAY_OF_MONTH);
-		}
-	}
-
+    public static Date get() {
+        return Date.from(nowUtc().toInstant());
```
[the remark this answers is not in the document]
venfernand: Yeah, that is true. That previous solution with `Calendar` pointed me into wrong direction.
Now this is changed to use the default `java.util.Date` constructor.

src/freenet/support/CurrentTimeUTC.java
```diff
@@ -3,61 +3,49 @@
...
+     * You should however use CurrentTimeUTC#getInMillis() instead because
+     * the JavaDoc of {@link System#currentTimeMillis()} does not explicitly state what time zone it returns.
+     * Therefore, by using this wrapper function, your code clearly states that it uses UTC time.
+     */
+    public static long getInMillis() {
+        return System.currentTimeMillis();
+    }
+
+    public static int getYear() {
+        return nowUtc().get(ChronoField.YEAR);
```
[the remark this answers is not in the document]
venfernand: You are right, those methods return values related to dates only. There is no need to use `*DateTime` for them.

venfernand:
I made changes mentioned in review.
Now it looks better.

I may also create a deprecation PR and remove usages of this class.

ArneBab:
> I made changes mentioned in review. Now it looks better.
> 
> I may also create a deprecation PR and remove usages of this class.

At work we have the policy that every deprecation needs a comment people can copy paste to replace the usage of the deprecated method. I would suggest doing that here, too.

We cannot remove this until there isn’t a single use in **plugins**.

bertm (approved):
Looks good to me :rocket:

ArneBab (approved):

ArneBab:
merged - thank you!

src/freenet/support/CurrentTimeUTC.java
Bombe: Yeah, take a guess in which plugins this is used… we’re never going to get rid of this. 😭

PR #868, opened 2023-10-14
Support schema hypha[net] and add tests for schema

ArneBab:


Bombe:
There’s an additional check for whether a key starts with `freenet:` in `WelcomeToadlet.java`, line 454, which looks like it could be thrown out (because `FreenetURI` will handle everything around that), while you’re at it. 🙂

Bombe (approved):

ArneBab:
@Bombe merged now — thank you for your review!

PR #869, opened 2023-10-14
Update sharesite plugin to 0.5.1

ArneBab:


Bombe (approved):
Didn’t check that the key actually contains Sharesite 0.5.1 but I trust you. 😄

ArneBab:
@Bombe merged — thank you. Though I hope we can get out of the "trust you" and into explicit validation again. I am still too central (but we need to keep momentum).

PR #870, opened 2023-10-14
Deprecate CurrentTimeUTC

bertm:
Follow-up on discussion in #866

### DIFF
--- a/src/freenet/support/CurrentTimeUTC.java
+++ b/src/freenet/support/CurrentTimeUTC.java
@@ -10,37 +10,62 @@ import java.util.Date;
 
 /**
  * A helper class, which returns the current time and date in UTC timezone.
+ *
+ * @deprecated use the Java 8 Date-Time APIs instead
+ * @see java.time.Instant
+ * @see java.time.LocalDate
+ * @see java.time.ZoneOffset#UTC
  */
+@Deprecated
 public class CurrentTimeUTC {
 
+    /**
+     * Get the current date.
+     * @deprecated use {@link Date#Date() new Date()} or {@link java.time.Instant}
+     */
+    @Deprecated
     public static Date get() {
         return new Date();
     }
 
     /**
-     * Get the current time in milliseconds.
-     *
-     * In the current implementation, this just returns {@link System#currentTimeMillis()}.
-     * You should however use CurrentTimeUTC#getInMillis() instead because
-     * the JavaDoc of {@link System#currentTimeMillis()} does not explicitly state what time zone it returns.
-     * Therefore, by using this wrapper function, your code clearly states that it uses UTC time.
+     * Get the current time in milliseconds since epoch.
+     * @return the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC.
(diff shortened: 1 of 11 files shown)


src/freenet/support/CurrentTimeUTC.java
```diff
@@ -29,30 +40,45 @@ public static Date get() {
...
 	 */
+	@Deprecated
 	public static long getInMillis() {
 		return System.currentTimeMillis();
 	}
 
+	/**
+	 * Get the current calendar year.
+	 * @deprecated use {@link java.time.LocalDate#now(java.time.ZoneId) LocalDate.now(ZoneOffset.UTC)}
+	 */
```
ArneBab, line 55: this replacement does not return the year.

src/freenet/support/CurrentTimeUTC.java
```diff
@@ -29,30 +40,45 @@ public static Date get() {
...
 		synchronized(mCalendar) {
 			mCalendar.setTimeInMillis(System.currentTimeMillis());
 			return mCalendar.get(Calendar.YEAR);
 		}
 	}
 
+	/**
+	 * Get the zero-indexed current calendar month.
+	 * @return the zero-indexed month number, where 0 indicates January and 11 indicates December
+	 * @deprecated use {@link java.time.LocalDate#now(java.time.ZoneId) LocalDate.now(ZoneOffset.UTC)}
```
ArneBab, line 67: this replacement does not return the month. It also does not explain the sad peculiarities of the month indexing.

src/freenet/support/CurrentTimeUTC.java
```diff
@@ -29,30 +40,45 @@ public static Date get() {
...
 		synchronized(mCalendar) {
 			mCalendar.setTimeInMillis(System.currentTimeMillis());
 			return mCalendar.get(Calendar.MONTH);
 		}
 	}
 
+	/**
+	 * Get the current day of the month.
+	 * @deprecated use {@link java.time.LocalDate#now(java.time.ZoneId) LocalDate.now(ZoneOffset.UTC)}
+	 */
```
ArneBab, line 80: this replacement does not return the day.

ArneBab:
Thank you for stepping up. Currently the comments do not enable quick replacement.

Please try to put yourself in the shoes of someone who wrote a plugin three years ago and since then has only programmed websites with vue or similar. The comment needs to enable them to adjust the code without investigating, otherwise it will never be possible to turn this into deprecation for removal. There are already modified plugins in use. If it’s not trivial (in the strictest sense) to replace the deprecation, removing the deprecated version will still break plugins in the next decade.

bertm:
I would not expect plugin developers to need this level of hand-holding in finding alternatives for these (until now mostly undocumented) methods. I've now updated the deprecation notices with verbatim copies of the new implementations provided in #866 to allow for a copy-paste migration path.

bertm:
Updated the commits to have GPG signatures.

Bombe (approved):
🔥🔥🔥

ArneBab:
@bertm merged — thank you! And welcome back! It’s great to see you again :heart:

PR #873, opened 2023-10-22
fix default bookmarks activelinks

TheSeeker:
Bookmarks may not point to the root of a site, so use regex to construct the activelink instead of blindly tacking activelink.png at the end.

This assumes activelinks are always in the root.

### DIFF
--- a/src/freenet/clients/http/WelcomeToadlet.java
+++ b/src/freenet/clients/http/WelcomeToadlet.java
@@ -15,6 +15,8 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.HashSet;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import freenet.client.ClientMetadata;
 import freenet.client.HighLevelSimpleClient;
@@ -39,6 +41,12 @@ public class WelcomeToadlet extends Toadlet {
 
     /** Suffix {@link #path()} with "#" + BOOKMARKS_ANCHOR to deep link to the bookmark list */
     public static final String BOOKMARKS_ANCHOR = "bookmarks";
+
+    // Regex Patterns to use when generating keys for activelinks.
+    // For USK and SSK we want to match from the start of the key to the first instance of -#/ or /#/.
+    protected static final Pattern PATTERN_USK_SSK = Pattern.compile("^.*[\/\-][0-9]+\/");
+    // For CHK and KSK we match from the start of the key to the first slash and append the activelink there.
+    protected static final Pattern PATTERN_CHK_KSK = Pattern.compile("^.*\/");
 
     final Node node;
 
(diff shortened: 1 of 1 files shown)


Bombe:
Yeah, you’re not gonna sneak that one in without a test. 😄

ArneBab:
Our build system seems unhappy about Capitalized Case.

Bombe:
Uh, yeah, as should have been @TheSeeker’s IDE?! Because that’s not valid Java…

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -91,10 +91,11 @@ private void addCategoryToList(BookmarkCategory cat, HTMLNode list, boolean noAc
                     // extract the key type
                     String keyType = initialKey.substring(1,3);
                     String key = '/' + initialKey + (initialKey.endsWith("/") ? "" : "/");
+					Matcher match = null;
```
Bombe, line 94: Seeing that `match` is not used anywhere outside the two branches within the `switch` statement, I believe “fix scope issue” and this change are diametrally opposed. 😀

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -91,16 +91,16 @@ private void addCategoryToList(BookmarkCategory cat, HTMLNode list, boolean noAc
                     // extract the key type
                     String keyType = initialKey.substring(1,3);
                     String key = '/' + initialKey + (initialKey.endsWith("/") ? "" : "/");
-                    switch keyType {
+                    switch (keyType) {
                         case "USK":
                         case "SSK": 
                             Matcher match = PATTERN_USK_SSK.matcher(key);
-                            if match.matches() {
+                            if (match.matches()) {
```
Bombe, line 99: Ahahaha, now I see, you’ve been writing too much Go, haven’t you? 🤣

Bombe (changes requested):
And the test is still missing! 🙂

src/freenet/clients/http/WelcomeToadlet.java
ArneBab: Yes, but the switch branches share the scope, so that’s how it has to be to avoid inventing new names for the different branches.

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -80,7 +88,42 @@ private void addCategoryToList(BookmarkCategory cat, HTMLNode list, boolean noAc
                 HTMLNode cell = row.addChild("td", "style", "border: none;");
                 if (item.hasAnActivelink() && !noActiveLinks) {
                     String initialKey = item.getKey();
-                    String key = '/' + initialKey + (initialKey.endsWith("/") ? "" : "/") + "activelink.png";
+                    // extract the key type
+                    String keyType = initialKey.substring(1,3);
```
ArneBab, line 92: this has to go from 0 to 3 — Java is zero-indexed.

src/freenet/clients/http/WelcomeToadlet.java
```diff
@@ -80,7 +88,42 @@ private void addCategoryToList(BookmarkCategory cat, HTMLNode list, boolean noAc
...
+                    String key = '/' + initialKey + (initialKey.endsWith("/") ? "" : "/");
+					Matcher match = null;
+                    switch (keyType) {
+                        case "USK":
+                        case "SSK": 
+                            match = PATTERN_USK_SSK.matcher(key);
```
ArneBab, line 98: this only matches when the key is already well-formed ⇒ ideally extract to a static method (uses no `this`; just put `static` before the function name to compiler-check that), and write a test that takes the existing bookmarks as input (and maybe some bad but valid examples) and returns the correct key for the activelink.

ArneBab (changes requested):

ArneBab:
I agree very much that we need this fix (thank you!), but the current implementation is still buggy. A test just for the key adjustment should make this much easier to develop.

PR #874, opened 2023-10-29
Safer healing

ArneBab:


src/freenet/client/async/SimpleHealingQueue.java
```diff
@@ -73,6 +81,57 @@ public boolean innerQueue(Bucket data, byte[] cryptoKey, byte cryptoAlgorithm, C
...
+	 */
+	static boolean shouldHealBlock(
+			double nodeLocation,
+			double keyLocation,
+			double randomBetweenZeroAndOne) {
+		double distanceToNodeLocation = Location.distance(nodeLocation, keyLocation);
+		// accept half the healings in our 20% of the keyspace.
+		// If the key is inside "our" 20% of the keyspace, heal it with 50% probability.
+		if (distanceToNodeLocation < 0.1) {
+			// accept 50%, specialized to our own location (0.5 ** 4 ~ 0.0625). Accept 70% which are going
```
Bombe, line 123: I spent several minutes mulling this over, and either this comment has nothing to do with is happening here, or something is very off. 😁

src/freenet/client/async/SplitFileFetcher.java
```diff
@@ -297,7 +298,10 @@ public void onSplitfileCompatibilityMode(CompatibilityMode min, CompatibilityMod
     @Override
     public void queueHeal(byte[] data, byte[] cryptoKey, byte cryptoAlgorithm) {
         try {
-            context.healingQueue.queue(BucketTools.makeImmutableBucket(context.tempBucketFactory, data), cryptoKey, cryptoAlgorithm, context);
+            RandomAccessBucket dataBucket = BucketTools.makeImmutableBucket(
```
Bombe, line 301: Is `RandomAccessBucket` necessary here? `SimpleHealingQueue.queue` seems to accept a `Bucket`, so we shouldn’t be handling more than `Bucket`s.

src/freenet/client/async/SimpleHealingQueue.java
Bombe, line 1: All the changes in this class feel wrong. Apparently this class is supposed to decide something, but everything it needs to make that decision has to be handed in from another class? That doesn’t make a lot of sense.

The decision and the factors for the decision should be close to one another. One possibility would be hand in a `Supplier` for the outcome of the decision; the implementation of said `Supplier` would then be in `NodeClientCore`, where the instance of this class is created. (It doesn’t have to _live_ in `NodeClientCore`, though, and if I take a look at the number of lines `NodeClientCore` already has I would most definitely suggest to move that code to its own class, where — bonus! — it can be tested without having to make methods static and give them default visibility. 😄)

Also, according to the code both the node’s location and the node’s opennet status _can_ change at runtime but the singular instance of this class is only created once so _that_ is most definitely not correct. 🙂

Bombe (changes requested):

src/freenet/client/async/SimpleHealingQueue.java
ArneBab: Damn, just up to the last paragraph I thought “but it’s just unchanging information, nothing problematic” — and then you drop that bomb of changing at runtime ☺ (pun not intended but fun ☺). This needs more work … thank you for your review!

src/freenet/client/async/SimpleHealingQueue.java
```diff
@@ -73,6 +81,57 @@ public boolean innerQueue(Bucket data, byte[] cryptoKey, byte cryptoAlgorithm, C
...
+			ClientContext context,
+			SingleBlockInserter sbi) {
+		// pure darknet is safer against sybil attacks, so we can heal fully
+		if (!opennetEnabled) {
+			return true;
+		}
+		// ensure that we have a routing key
+		sbi.tryEncode(context);
+		double keyLocation = sbi.getKeyNoEncode().getNodeKey().toNormalizedDouble();
+		double randomBetweenZeroAndOne = NodeStarter.getGlobalSecureRandom().nextDouble() / Double.MAX_VALUE;
```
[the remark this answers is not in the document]
ArneBab: Yikes, seems I misread the implementation of nextDouble() — thank you!

src/freenet/client/async/SimpleHealingQueue.java
ArneBab: This should actually be somewhat correct, but there are several jumps in abstraction layers and logic between the comment and the implementation.  I think this needs more prose in the javadoc to actually explain what it does.

src/freenet/client/async/SimpleHealingQueue.java
ArneBab: This is implemented now: The SimpleHealingQueue receives a HealingDecisionSupplier which holds a reference to the node so it can get the current values of isOpennetEnabled and the nodeLocation.

Thank you for the review!

test/freenet/client/async/HealingDecisionSupplierTest.java
```diff
@@ -0,0 +1,44 @@
+package freenet.client.async;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.util.Arrays;
+
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+public class HealingDecisionSupplierTest {
```
Bombe, line 10: This test is missing some stuff for the node’s opennet status but that has already been covered in the other comment. 🙂

src/freenet/client/async/SplitFileFetcher.java
```diff
@@ -21,7 +21,9 @@
 import freenet.keys.FreenetURI;
 import freenet.node.BaseSendableGet;
 import freenet.support.Logger;
+import freenet.support.api.Bucket;
 import freenet.support.api.LockableRandomAccessBuffer;
+import freenet.support.api.RandomAccessBucket;
```
Bombe, line 26: That import is not needed anymore, is it?

Bombe (changes requested):

test/freenet/client/async/HealingDecisionSupplierTest.java
ArneBab: I added a darknet nest now: always heals in darknet.

src/freenet/client/async/HealingDecisionSupplier.java
```diff
@@ -0,0 +1,74 @@
...
+   * on the network.
+   *
+   * The continuous function is gauged to accept 50% of the keys close to our node: a peak at our
+   * own location for which the area below the curve between 0 and 0.1 sums up to 0.5.
+   *
+   * Close keys are those in our 20% of the keyspace: the ones that would reach us if we were one of
+   * 5 long distance peers of a peer node. These are the keys for which we are most likely to be
+   * the best next hop when seen from the originator.
+   */
+  static boolean shouldHealBlock(
```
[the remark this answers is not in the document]
ArneBab: That worked out pretty well — thank you!

I provide a package-private constructor now to pass in a random value provider so this can be tested deterministically, and it now looks pretty clean to me.

test/freenet/client/async/HealingDecisionSupplierTest.java
```diff
@@ -12,31 +12,62 @@ public class HealingDecisionSupplierTest {
...
+          () -> true,
+          () -> i);
+      assertThat(String.format("Healing triggers at random value %g", i),
+          healingDecisionSupplier.shouldHeal(0.1999),
+          Matchers.equalTo(true));
     }
   }
 
   @Test
   public void healingAccepts10PercentOfKeysAtLongDistance() {
```
Bombe, line 67: This is the only test that actually tests both scenarios (i.e. `shouldHeal == true` and `shouldHeal == false`). I’m missing those cases for the the other tests as well. 🙂 Also, `isOpennetEnabled` is never false, and the location is always `0.1`. The way the tests are currently written you could probably write an implementation of the HealingDecisionSupplier that is quite broken without the brokenness being detected by the tests.

Bombe (changes requested):

test/freenet/client/async/HealingDecisionSupplierTest.java
ArneBab: The tests now include inverse tests, too.

ArneBab: Opennet is false in the first test. But the location is always 0.1, yes: this is not intended to catch intentional errors, just to catch accidental ones.

But anyway: adding more tests of the same structure doesn’t cost much: https://github.com/hyphanet/fred/pull/874/commits/cb837421717060dc3fd490f0deea27c872797e7c → different node location, and diff in positive and in negative direction.

ArneBab: Also a test around zero: https://github.com/hyphanet/fred/pull/874/commits/4b26dbc2760993b96db69efc7549ba4df1a1b065

test/freenet/client/async/HealingDecisionSupplierTest.java
Bombe, line 1: Okay, the tests seem pretty much complete but now there is a _lot_ of duplication in this class. Extracting the meat of the tests (the object creation plus the assertion) into a method and then only having small test methods that are very explicit about what they are actually testing would improve this.

src/freenet/client/async/HealingDecisionSupplier.java
Bombe, line 1: This class has some weird formatting choices; the main javadoc is indented for some reason, there’s empty lines in the constructors that serve no purpose, and there’s multiple empty lines at the end of the class, again serving no more purpose than a single empty line. 🙂

Bombe (changes requested):

src/freenet/client/async/HealingDecisionSupplier.java
ArneBab: I now re-indented that and tabified. Weird how the default IntelliJ indentation goes so far off …

test/freenet/client/async/HealingDecisionSupplierTest.java
Bombe: Argh, your tendency to split parameters up in lines is really giving me a headache. 😁

Bombe (approved):

test/freenet/client/async/HealingDecisionSupplierTest.java
ArneBab: I think that’s from work: We have classes with 10 and more parameters (structural overhead at 5 million lines of code) and there you need the splitting into lines, so I got used to it …

ArneBab:
Merged — thanks a lot for your reviews!

PR #875, opened 2023-11-26
Disable activelink for Index of Indexes (workaround, because it fails)

ArneBab:
This is the simpler workaround for the started fix from TheSeeker → https://github.com/hyphanet/fred/pull/873 

We should fix activelink handling, but I don’t want to hotfix that in before release and the Index of Indexes does not really need an activelink right now.


Bombe (approved):

ArneBab:
@Bombe thank you for checking!

PR #878, opened 2023-12-31
Remove security lowering kludge.

Juiceman:
As stated in the description of this option, if a large portion of the network disables packet padding it makes the network easier to trace. 15 years later now, if network overhead is a real concern for some users they can tweak this themselves.


Juiceman:
will try again with a cleaner commit

PR #882, opened 2024-01-09
 Remove security lowering kludge.

Juiceman:
As stated in the description of this option, if a large portion of the network disables packet padding it makes the network easier to trace. 15 years later now, if network overhead is a real concern for some users they can tweak this themselves.

### DIFF
--- a/src/freenet/node/NodeCryptoConfig.java
+++ b/src/freenet/node/NodeCryptoConfig.java
@@ -240,18 +240,6 @@ public class NodeCryptoConfig {
 		});
 		
 		paddDataPackets = config.getBoolean("paddDataPackets");
-		securityLevels.addNetworkThreatLevelListener(new SecurityLevelListener<NETWORK_THREAT_LEVEL>() {
-
-			@Override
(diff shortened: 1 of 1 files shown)


src/freenet/node/NodeCryptoConfig.java
```diff
@@ -240,18 +240,6 @@ public void set(Boolean val) throws InvalidConfigValueException, NodeNeedRestart
...
-			@Override
-			public void onChange(NETWORK_THREAT_LEVEL oldLevel, NETWORK_THREAT_LEVEL newLevel) {
-				// Might be useful for nodes which are running with a tight bandwidth quota to minimize the overhead,
-				// so turn it off for LOW. Otherwise is sensible.
-				if(newLevel == NETWORK_THREAT_LEVEL.LOW)
-					paddDataPackets = false;
-				if(oldLevel == NETWORK_THREAT_LEVEL.LOW)
-					paddDataPackets = true;
-			}
-		});
```
ArneBab, line 254: does this also remove the config option in the UI?

ArneBab:
Oh — you’re right. Since with normal network level the overhead is rarely more than 12%, I agree that we should simply remove this.

ArneBab:
Merged into next — thank you again!

PR #883, opened 2024-01-15
Change string comparison into a .isEmpty() check

Juiceman:
We are actually checking if the string is empty. Using .isEmpty() is clearer.

### DIFF
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -1773,26 +1773,26 @@ class CSSTokenizerFilter {
 			}
 
 		}
-		if(isIDSelector && "".equals(id)) return null; // No ID
+		if(isIDSelector && id.isEmpty()) return null; // No ID
 
 		boolean elementValid =
 		    "*".equals(HTMLelement) ||
 		    (ElementInfo.isValidHTMLTag(HTMLelement.toLowerCase())) ||
-		    ("".equals(HTMLelement.trim()) &&
-                    ((!className.equals("")) || (!id.equals("")) || attSelections!=null ||
-                            !pseudoClass.equals("")));
+		    (HTMLelement.trim().isEmpty() &&
+                    ((!className.isEmpty()) || (!id.isEmpty()) || attSelections!=null ||
+                            !pseudoClass.isEmpty()));
 		if(!elementValid) return null;
 		
-		if(!className.equals("")) {
+		if(!className.isEmpty()) {
 		    // Note that the definition of isValidName() allows chained classes because it allows . in class names.
 		    if(!ElementInfo.isValidName(className))
 		        return null;
-		} else if(!id.equals("")) {
+		} else if(!id.isEmpty()) {
 		    if(!ElementInfo.isValidName(id))
 		        return null;
 		}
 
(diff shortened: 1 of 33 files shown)


src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1773,26 +1773,26 @@ else if(HTMLelement.indexOf('#')!=-1)
 			}
 
 		}
-		if(isIDSelector && "".equals(id)) return null; // No ID
+		if(isIDSelector && id.isEmpty()) return null; // No ID
```
ArneBab, line 1776: are we sure here that id cannot be null? the check before was null-safe.

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -4502,9 +4502,9 @@ public boolean recursiveDoubleBarVerifier(String expression,ParsedWord[] words,F
 							}
 							// Against the rest of the pattern: the part that we've tried and failed plus the part that we haven't tried yet.
 							// NOT against the verifier we were just considering, because the double-bar operator expects no more than one match from each component of the pattern.
-							String pattern = ignoredParts+((("".equals(ignoredParts))||("".equals(secondPart)))?"":"a")+secondPart;
+							String pattern = ignoredParts+(((ignoredParts.isEmpty())||(secondPart.isEmpty()))?"":"a")+secondPart;
```
ArneBab, line 4505: are we sure that ignoredParts and secondPart cannot be null?

ArneBab:
Except for few cases where I’m not sure about null-safety, this looks good to me.

src/freenet/client/filter/CSSTokenizerFilter.java
Bombe: Yes, this is either an empty string (assigned at declaration), or it’s the output of a `trim()` method call so this cannot be `null`.

src/freenet/client/filter/CSSTokenizerFilter.java
Bombe: Yeah, same as above; assigned an empty string at declaration time, or assigned other values that are not `null`. :)

src/freenet/clients/fcp/DirPutFile.java
```diff
@@ -50,7 +50,7 @@ public static DirPutFile create(SimpleFieldSet subset, String identifier, boolea
 		if(name == null)
 			throw new MessageInvalidException(ProtocolErrorMessage.MISSING_FIELD, "Missing field Name", identifier, global);
 		String contentTypeOverride = subset.get("Metadata.ContentType");
-		if(contentTypeOverride != null && (!contentTypeOverride.equals("")) && !DefaultMIMETypes.isPlausibleMIMEType(contentTypeOverride)) {
+		if(contentTypeOverride != null && (!contentTypeOverride.isEmpty()) && !DefaultMIMETypes.isPlausibleMIMEType(contentTypeOverride)) {
```
Bombe, line 53: The parentheses around the `isEmpty()` are unnecessary.

Juiceman: Fixed in [a29fa9d].
Thank you

ArneBab:
Merged — thank you!

PR #885, opened 2024-01-26
Optimize message filters

ArneBab:


src/freenet/node/AnnounceSender.java
```diff
@@ -184,10 +184,8 @@ private void realRun() {
 				MessageFilter mfRejectedOverload = MessageFilter.create().setSource(next).setField(DMT.UID, uid).setTimeout(ACCEPTED_TIMEOUT).setType(DMT.FNPRejectedOverload);
 				MessageFilter mfOpennetDisabled = MessageFilter.create().setSource(next).setField(DMT.UID, uid).setTimeout(ACCEPTED_TIMEOUT).setType(DMT.FNPOpennetDisabled);
 
-				// mfRejectedOverload must be the last thing in the or
-				// So its or pointer remains null
-				// Otherwise we need to recreate it below
```
Bombe, line 189: Oof, this comment gives me goosebumps, and definitely _not_ in a good way… 😄

Have you made sure that whatever this comment is talking about is being handled correctly?

src/freenet/node/PeerMessageQueue.java
```diff
@@ -20,6 +20,8 @@
  * @author Matthew Toseland <toad@amphibian.dyndns.org> (0xE43DA450)
  */
 public class PeerMessageQueue {
+	// TODO: check whether we can use a standard PriorityBlockingQueue (since Java 8)
```
Bombe, line 23: This should also much rather be a ticket than a comment.

src/freenet/node/RequestSender.java
```diff
@@ -939,10 +939,8 @@ protected MessageFilter makeAcceptedRejectedFilter(PeerNode next,
 		MessageFilter mfRejectedLoop = MessageFilter.create().setSource(next).setField(DMT.UID, uid).setTimeout(acceptedTimeout).setType(DMT.FNPRejectedLoop);
 		MessageFilter mfRejectedOverload = MessageFilter.create().setSource(next).setField(DMT.UID, uid).setTimeout(acceptedTimeout).setType(DMT.FNPRejectedOverload);
 		
-		// mfRejectedOverload must be the last thing in the or
-		// So its or pointer remains null
-		// Otherwise we need to recreate it below
-		return mfAccepted.or(mfRejectedLoop.or(mfRejectedOverload));
+		// the order of these filters is performance critical. The last or-filter is checked first.
```
Bombe, line 942: ~~This seems really weird. Looking at the filter creation above it’s absolutely unclear _why_ these filters should show any difference in performance, they are all created the same way, only with a different message type. Now, I haven’t timed any of this but if the reordering of the filters changes the overall performance of this method, it’s probably _not_ because the filters themselves have differing performances but because the most-matched filter was checked last and thus many useless matchers were made, in which casethe comment should reflect that.~~

Okay, re-read the comment, it’s not what it said. 🙂 Anyway, if measuring has shown that reordering the filters increased performance because many useless matches were not made, the comment should say _that_, if anything. 😄

Bombe (changes requested):

src/freenet/node/AnnounceSender.java
ArneBab: I think this is a leftover from a time when this was spaghetti-code where they re-used the instances. That’s the only way how the comment could make sense.

We no longer do that, from what I can tell.

ArneBab:
I removed the TODOs again (reverted the commit adding them). These were just notes to myself, but I am not likely to get to them in the next few weeks, so they actually serve no purpose. When I get into that part of the code again at another time, I’ll see how to improve it.

Thank you for reminding me!

src/freenet/node/RequestSender.java
ArneBab: I added a second line now: **So the last filter in the or-"chain" must be the filter which matches most frequently**

Bombe (approved):

ArneBab:
Merged — thank you very much for your review!

PR #886, opened 2024-01-26
Increase default bandwidth to 160KiB upload, when detection fails

ArneBab:
Bandwidth in countries in which Hyphanet is used the most has increased quite a bit and using the low default bandwidth gave a bad first impression.

With 160KiB/s, fred aims for 26 opennet peers.


Bombe:
Generally this seems fine, I’m just really confused about all the different numbers; in some places a download limit is set to 1024, in other places it’s set to 2048, and then there’s also 768?!

ArneBab:
These other places in the diff are only information text so people get an idea what the numbers mean.

Bombe (approved):

ArneBab:
Thank you for your review! Merged.

PR #887, opened 2024-01-26
Remove hash generation to native big integer to reduce dependencies.

ArneBab:
In the Guix build Hash.java was a blocker that had to be circumvented, because the i2p tools are not packaged.

Since the two methods providing NativeBigInteger are unused, this PR removes them completely.

### DIFF
--- a/src/freenet/crypt/Hash.java
+++ b/src/freenet/crypt/Hash.java
@@ -6,7 +6,6 @@ package freenet.crypt;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
 
-import net.i2p.util.NativeBigInteger;
 import freenet.support.HexUtil;
 
 /**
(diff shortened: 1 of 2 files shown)


Bombe:
Well, _technically_, this is public API but a grep over my `~/Workspace` doesn’t get any hit outside of this file and its test.

Wait. There’s a test that you didn’t change and that now doesn’t compile, shouldn’t we have checks for that already or is that still WIP?

ArneBab:
This was only merged in 1495; I’m pretty sure no one is using it yet.

But the test should not be broken; I missed that (need to add a gradle run to the workflows).

ArneBab:
@Bombe is this OK to merge?

Bombe (approved):

ArneBab:
Thank you for your review! Merged.

PR #890, opened 2024-01-29
Convert tabs to spaces

Juiceman:
Start of a larger project to bring entire code-base to consistent standard.
Countless hours are lost in order to avoid bit-rot with old PRs; while a noble concept, it just furthers the technical debt of this already old code-base.
IMHO, tabs to spaces should be easiest to reconcile in case of conflict.
Start with some small files as proof of concept.

### DIFF
--- a/src/freenet/tools/AddRef.java
+++ b/src/freenet/tools/AddRef.java
@@ -20,108 +20,108 @@ import freenet.support.io.LineReadingInputStream;
 
 public class AddRef {
 
-	/**
-	 * Connects to a FCP server and adds a reference
-	 * @param args
-	 */
-	public static void main(String[] args) {
-		if(args.length < 1){
-			System.err.println("Please provide a file name as the first argument.");
-			System.exit(-1);
-		}
+    /**
+     * Connects to a FCP server and adds a reference
+     * @param args
+     */
+    public static void main(String[] args) {
+        if(args.length < 1){
+            System.err.println("Please provide a file name as the first argument.");
+            System.exit(-1);
+        }
 
-		final File reference = new File(args[0]);
-		if((reference == null) || !(reference.isFile()) || !(reference.canRead())){
-			System.err.println("Please provide a file name as the first argument.");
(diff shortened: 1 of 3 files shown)


Bombe:
Your proposed standard is in conflict with both our `CONTRIBUTING.md` and my personal taste. How dare you! 🤣 

However, `CONTRIBUTING.md` is also already in conflict with my personal taste. I do like using the amount of space per indentation level (4) but I also do like using a tab size of 4 and then using tabs. I of course consider this vastly superior to anything else but the good thing about this is that if somebody thinks 2, 3, 5, or 8 are good values for indentation levels, they can configure their editor in this way and everybody is happy. Using spaces for indentation robs you of the possibility to have more people happy.

I’m not sure _this_ is the place to have this discussion but I feel that we _need_ this discussion to happen somewhere.

ArneBab:
While the discussion can have merit, there is an important catch: if you do this, you have to also do it in all pending pull requests to avoid getting more merge conflicts. That may mean replacing pull-requests with your own versions when you don’t have access to the repository they were filed from.

Do you have the persistence to see this to the end? If yes, then a discussion can have merit.

ArneBab:
> I noticed the links for the mailing list and IRC are 404. Do we have a place for this discussion? :)

Where are the links 404? ⇒ I’ll need to fix them :-)

These should work: https://www.hyphanet.org/pages/help.html#mailing-lists

Juiceman:
The 404 links inside the "code review" paragraph (about 1/3 down contributing.md)
image

ArneBab:
> The 404 links inside the "code review" paragraph (about 1/3 down contributing.md) image

Thank you! Are you up for a review? ☺

https://github.com/hyphanet/fred/pull/893

ArneBab:
There is an essential question to which I did not see an answer:

> If you do this, you have to also do it in all pending pull requests. … Do you have the persistence to see this to the end? If yes, then a discussion can have merit.

PR #891, opened 2024-01-29
Increase max transparent passthrough to 200MiB for links in m3u-lists.

ArneBab:
At 10 cached files that would be at most 2GiB memory usage in the browser in the extreme case (see m3u-player.js), staying below the 2GiB in Chrome per tab, so it should still work well on most systems.


Juiceman (approved):
I see no errors in the changes.

ArneBab:
@Juiceman thank you for your review!

PR #892, opened 2024-01-30
Only link indexes via the index of indexes (required by law now!)

ArneBab:
With the decision by OLG Köln from 2023-11
https://www.heise.de/news/Oberlandesgericht-Cloudflare-haftet-als-Taeter-fuer-Urheberrechtsverletzungen-9355760.html caching services must stop access quickly when they get a report.

Therefore linking directly to an index becomes too risky: we’ll have link them only via the Index of Indexes so we can remove the link to the index quickly if an index fails to react to a report in time.


ArneBab:
Thank you! Merged.

PR #893, opened 2024-01-31
Fix links to mailing lists and IRC in CONTRIBUTING.md — thanks to Juiceman

ArneBab:


Juiceman (approved):
lgtm

CONTRIBUTING.md
```diff
@@ -37,8 +36,8 @@ producing software that is readable, correct, and sufficiently efficient.
 Breaking API creates an immense amount of work for developers of other projects,
 so it is *not* something to be done lightly. If you feel you must make a change
 that breaks API and cannot maintain backwards compatibility, please first raise
-the issue with the community - the [mailing list](https://emu.freenetproject.org/cgi-bin/mailman/listinfo/devl)
-and [IRC](https://freenetproject.org/irc.html) are good places to contact us.
+the issue with the community - the [mailing list](https://www.hyphanet.org/pages/help.html#mailing-lists)
+and [IRC](https://www.hyphanet.org/pages/help.html#chat-with-us) are good places to contact us.
```
Bombe, line 40: Hmm, while we’re at it, maybe we should actually point people to `#hyphanet` instead of `#freenet`?

ArneBab: I’m not sure about that. Don’t want to splinter places where people discuss without having a need.

ArneBab:
Merged — I think when it’s not required to prevent immediate damage, we should ask people in the community whether they want to switch channels and make the links mirror that.

PR #894, opened 2024-01-31
Remove time-dependence of compressor selection

ArneBab:
Max time for single compressor makes the insert compression CPU dependent, so it should not have been used in the first place.


Bombe:
You’ve missed the occurrences in the translation files. 🙂

ArneBab:
> You’ve missed the occurrences in the translation files. 🙂

ah, damn, I somehow always think I should not remove these, because they may be needed again — even if they won’t ever be needed again.

Thank you!

Bombe:
> > You’ve missed the occurrences in the translation files. 🙂
> 
> ah, damn, I somehow always think I should not remove these, because they may be needed again — even if they won’t ever be needed again.

Even _if_ they will be needed again, they’re still retrievable from Git. 🙂

ArneBab:
> > > You’ve missed the occurrences in the translation files. 🙂
> > ah, damn, I somehow always think I should not remove these, because they may be needed again — even if they won’t ever be needed again.
> Even _if_ they will be needed again, they’re still retrievable from Git. 🙂

yes, but … yes, they are :-)

Juiceman:
Are we sure a time limit is not a good idea? What is to keep the compression thread running for hours in order to discover it compressed a file 1% ?

ArneBab:
Yes: making compression time-dependent effectively makes them CPU dependent, so someone with another computer than you may not be able to re-create the same key as you (without exactly defining the compression settings).

Compression already only checks a fraction of the size to test whether compression will be worth the effort. We could require a higher improvement for slow compressors (that may actually help with LZMA), but this may need a new compatibility mode.

Juiceman:
If they are trying to recreate the same key you would think they would specify it.
That being said, 8Mib is not much data to try. I doubt anyone actually hit the timeout.
I see there is no sanity check on the size of 'amountOfDataToCheckCompressionRatio'. I will submit a patch limiting it to less than 100MiB

Juiceman:
See PR #910

ArneBab:
@Bombe is this OK to merge now?

Bombe (approved):
Some options throw an `InvalidConfigValueException` but other simply ignore the new value… do we have a guideline or something for when to use what method of not accepting the new value?

Other than that, looks good.

ArneBab:
I don’t think we have a guideline — that would be a step towards more consistent code.

ArneBab:
Thank you for your reviews! Merged.

PR #895, opened 2024-02-01
Adds the git attributes file to enforce consistent line endings. 

alvonellos:
I added the git attributes as follows: 

* text=lf (treat all text files as lf)
*.java text (treat all java files as text)
*.m3u text (treat all m3u as text)
*.gitattributes as lf also. 

How does this look @ArneBab, @Bombe

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+#Set default line endings to be in accodance with project specs.
+* text=lf
+*.java text
+*.m3u text
+*.gitattributes text=lf
+


ArneBab:
Does this fix the tests for you on windows?

alvonellos:
Not all of them. I'm looking into why for the others. But shouldn't we have a good gitattributes file?

Bombe:
> Not all of them. I'm looking into why for the others. But shouldn't we have a good gitattributes file?

IMHO, no. Ideally we wouldn’t need a `.gitattributes` file at all. We should have a `.gitattributes` file for cases where we absolutely need it but we should be using the defaults of whatever we are using (unless there are reasons not to). I’m pretty sure that e.g. `.java` files are already recognized as text so having extra configuration seems superfluous to me.

alvonellos:
It prevents inconsistent behavior because of line ending issues. See this warning that appears in my IDE after adding the .gitattributes file: 

image

The other problem is related to a path or permissions issue on my system and has no bearing on this commit. I think this works much better now. @Bombe 

I think we should set a readme info also and specify that the project line endings must be \n because of reasons. What do you think?

Bombe:
> It prevents inconsistent behavior because of line ending issues. See this warning that appears in my IDE after adding the .gitattributes file:
> 
> image

That looks like your IDE is configured to report these things, when they have no actual effect on what you’re doing. The IDE can use both line endings, the Java compiler can do both line endings; having your IDE complain about that is a problem that should be fixed in your IDE.

Also, it looks like you’re on Windows (because Git seems to have automatically checked out text files using your platform’s default line separator), and yet you’re not using the Windows default line separators for the project. Again, that’s an issue with the project’s configuration.

PR #897, opened 2024-02-03
Convert tabs to spaces L10n

Juiceman:
Going folder by folder to make review easier

### DIFF
--- a/src/freenet/l10n/BaseL10n.java
+++ b/src/freenet/l10n/BaseL10n.java
@@ -37,120 +37,120 @@ import freenet.support.io.FileUtil;
  */
 public class BaseL10n {
 
-	/**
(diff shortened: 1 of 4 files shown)


ArneBab:
Please merge them together — we must have this in one go. But before you spend more time on this:

There is an essential question to which I did not see an answer:

> If you do this, you have to also do it in all pending pull requests. … Do you have the persistence to see this to the end? If yes, then a discussion can have merit.

If you’re not up to actually finishing the task (including all PRs), this would create more problems than it would solve.

Juiceman:
Yes, I will put in the time to process all the PR with the same treatment.
Can you describe the workflow on this?
How can I get the PR that are sent against the project repo into my repo,
or is there a different way to do it?

PR #910, opened 2024-02-04
Apply sanity check for size

Juiceman:
There were no restraints on negative values or extremely large values of amountOfDataToCheckCompressionRatio. Add a less than zero and a max of 100 MiB check.

### DIFF
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -1603,6 +1603,12 @@ public class Node implements TimeSkewDetectorCallback {
 			@Override
 			public void set(Long amountOfDataToCheckCompressionRatio) {
 				synchronized(Node.this) {
+					if (amountOfDataToCheckCompressionRatio < 0 || amountOfDataToCheckCompressionRatio > 100 * 1024 * 1024) {
+						Logger.normal(Node.class, "Amount of data to check for compression should be 100 MiB max"
+								+ amountOfDataToCheckCompressionRatio);
+						return;
+					}
+
 					Node.this.amountOfDataToCheckCompressionRatio = amountOfDataToCheckCompressionRatio;
 				}
 			}
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -1604,8 +1604,8 @@ public class Node implements TimeSkewDetectorCallback {
 			public void set(Long amountOfDataToCheckCompressionRatio) {
 				synchronized(Node.this) {
 					if (amountOfDataToCheckCompressionRatio < 0 || amountOfDataToCheckCompressionRatio > 100 * 1024 * 1024) {
-						Logger.normal(Node.class, "Amount of data to check for compression should be 100 MiB max"
-								+ amountOfDataToCheckCompressionRatio);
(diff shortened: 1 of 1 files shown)


src/freenet/node/Node.java
```diff
@@ -1603,6 +1603,12 @@ public Long get() {
 			@Override
 			public void set(Long amountOfDataToCheckCompressionRatio) {
 				synchronized(Node.this) {
+					if (amountOfDataToCheckCompressionRatio < 0 || amountOfDataToCheckCompressionRatio > 100 * 1024 * 1024) {
+						Logger.normal(Node.class, "Amount of data to check for compression should be 100 MiB max"
+								+ amountOfDataToCheckCompressionRatio);
```
Bombe, line 1608: This will make the log output look like "Amount of data to check for compression should be 100 MiB max12312412414" which looks like the node just threw up in the logfile. 😄

src/freenet/node/Node.java
```diff
@@ -1603,6 +1603,12 @@ public Long get() {
 			@Override
 			public void set(Long amountOfDataToCheckCompressionRatio) {
 				synchronized(Node.this) {
+					if (amountOfDataToCheckCompressionRatio < 0 || amountOfDataToCheckCompressionRatio > 100 * 1024 * 1024) {
```
Bombe, line 1606: How do we handle the `0` case? Does the compression test stuff do something sensible here?

Bombe (changes requested):

src/freenet/node/Node.java
Juiceman: Fixed in 8c222a28c13e17437fb78cb98356f58372140042

src/freenet/node/Node.java
Juiceman: If I am reading this correctly, it just will never run checkCompressionEffect be cause the ratio would be 0 and the ++i  jumps past that.

	int bufferSize = 32768;
			byte[] buffer = new byte[bufferSize];
			long iterationToCheckCompressionRatio = amountOfDataToCheckCompressionRatio / bufferSize;
			int i = 0;
			while(true) {
				int l = (int) Math.min(buffer.length, maxReadLength - read);
				int x = l == 0 ? -1 : is.read(buffer, 0, l);
				if(x <= -1) break;
				if(x == 0) throw new IOException("Returned zero from read()");
				gos.write(buffer, 0, x);
				read += x;
				if(cos.written() > maxWriteLength)
					throw new CompressionOutputSizeException();

				if (++i == iterationToCheckCompressionRatio && minimumCompressionPercentage != 0) {
					checkCompressionEffect(read, cos.written(), minimumCompressionPercentage);
				}

Bombe: Heh, only for small values of “never,” though… with a large enough file `++i` can overflow. Granted, that will only happen after 140 TiB but you never know. 😁

Bombe (approved):
I am (of course) very appalled by the lack of tests for this but `Node` is incredibly hard to test. I should really spend some time on that…

ArneBab:
Merged — thank you!

PR #914, opened 2024-02-08
Tabs to spaces pr 828

Juiceman:
Bring outstanding PR 828 up to no-tabs-styel


ArneBab:
Please also include the names of the pull-requests in the title, otherwise we have to process the review in a two-step process to find out whether all are processed (first look up the PR, then check it here).

For PRs with submodule issues, you may need to either do git submodule relocating (likely for the contrib subrepo) or merge next into them (we removed the contrib submodule because it repeatedly caused problems and it is no longer needed with gradle). Merging next may produce merge-conflicts. If you can resolve these, it might get us closer to being able to merge some of them (but be warned that for some it’s hard: you’ll have to first understand the logic).

Juiceman:
I will fix these up soon.

PR #917, opened 2024-02-08
Convert tabs to spaces Master PR

Juiceman:
Zap all tabs to spaces and eliminate extra spaces at end of lines for entire codebase. Other outstanding PR will be converted also.


ArneBab:
Which method do you use to convert tabs to spaces? For review I’ll have to be able to run this locally.

ArneBab:
Please also reference the discussion about the style along with its result. As I wrote in https://github.com/hyphanet/fred/pull/890#issuecomment-1926750052 this must be discussed. Did you speak with @Bombe about it and reach a conclusion to follow the style used here?

Juiceman:
I am on Windows. I used https://tools.stefankueng.com/tabspace.html
Tabspace /ext:java
The same should be doable on *nix with a script but that is not in my skillset. Perhaps I can find something online.
Other than the EOL extra spaces removal it would be a simple find and replace of TAB with 4 spaces. If the EOL spaces makes this complicated I can redo this leaving that option disabled.

As far as checking that the files in your local branch match when processed: if you can't run a script, you can compile and verify the bytecode is the same?

@Bombe and I have not discussed it after I pointed out the code style of the project was supposed to be google-java. If that is no longer the case or, I misunderstood we should definitely have a wider discussion. 

I specifically wanted to process the tabs to spaces without reindenting the code to make merging easy. If I have to use tabs I will need to reindent everything and I am not looking forward to that spaghetti. 

After the tabs to spaces is merged on /next other programmers will have a clean merging point.
I will also finish processing the open pull requests.

Once that is complete I will run this
https://github.com/google/google-java-format
I will also do the open PRs.

Thoughts?

ArneBab:
Please discuss it with Bombe. Such a change that affects all annotations in all files needs to be discussed beforehand. Just replacing tabs by 4 spaces does not necessarily make the code consistent with google-java.

This will only have added value if it causes common IDEs to match the indentation automatically.

Juiceman:
The 4 spaces instead of tabs is a tiny part of google-java style.
The jar provided by google will do the rest. That is referenced at the end of my last comment.

Juiceman:
I will plan on closing this pull request in favor of #930

PR #935, opened 2024-02-10
Replace deprecated html tag

Juiceman:


src/freenet/clients/http/staticfiles/js/m3u-player.js
Bombe, line 1: I believe this file is protected by a checksum somewhere, that checksum needs to be updated as well.

src/freenet/clients/http/staticfiles/freenetjs/hosted.html
Bombe, line 1: Wow, this whole file looks completely broken. Is this actually still used anywhere? @ArneBab, do you know more about this?

Bombe (changes requested):

src/freenet/clients/http/staticfiles/js/m3u-player.js
Juiceman: Nice catch!
Curiously, the original file did not match the checksum in the code.
private static String generateRestrictedScriptSrc() {
		// TODO: auto-generate these hashes from the path to the source file
		String[] allowedScriptHashes = new String[] {
				"sha256-RY9OjosvFxocXEmcUqBJ2v1KByDRdUgnGHYSL3Qx/t8=" // freenet/clients/http/staticfiles/js/m3u-player.js
		};
image

ArneBab: The required checksum formatting to work in the browser is `sha256sum m3u-player.js | cut -d " " -f 1 | xxd -r -p | base64`.  On my computer this matches — if this does not match for you: did the linebreaks get changed on checkout?

src/freenet/clients/http/staticfiles/freenetjs/hosted.html
ArneBab: I think GWT was for web-pushing and has not been used for many years. I don’t think it still works, but someone needs to check that before removing it.

src/freenet/clients/http/staticfiles/js/m3u-player.js
Juiceman: If you say it matches I believe you. I do not have that command on Windows to replicate it.

I will back this commit out since it only fixed a spelling typo in a comment.

Bombe (approved):

ArneBab:

PR #936, opened 2024-02-10
Change more string comparisons into .isEmpty() checks

Juiceman:
We are actually checking if the string is empty. Using .isEmpty() is clearer.

These should have been included in #883


Bombe (approved):

ArneBab:

PR #937, opened 2024-02-10
Replace indexOf with .contains()

Juiceman:
This is much easier to read.


src/freenet/io/xfer/BlockReceiver.java
```diff
@@ -224,7 +224,7 @@ public void onMatched(Message m1) {
             	Logger.minor(this, "Received "+m1);
             if ((m1 != null) && m1.getSpec().equals(DMT.sendAborted)) {
 				String desc=m1.getString(DMT.DESCRIPTION);
-				if (desc.indexOf("Upstream")<0)
```
Juiceman, line 227: Can someone review the underlying logic here please? We are logging an "Upstream transmit error" if the message does NOT contain "Upstream"?

Bombe (approved):

src/freenet/io/xfer/BlockReceiver.java
Bombe: Well, we could probably add it to the never-ending list of things we’d need to check up on… 😄

ArneBab:

PR #941, opened 2024-02-17
break early when condition is met

Juiceman:
Break out of a loop once the condition is met


Juiceman:
Closing this mess. Will recreate cleanly.

PR #942, opened 2024-02-18
Java 21 fix: int used as return type but integer required

ArneBab:
This change is needed to compile on Java 21.


bertm (changes requested):
An `Integer` return type implies optionality of the result, as it introduces `null` as a possible value. However, `SortedIntSet` uses `-1` to indicate no result.

Can we solve this by casting the `Integer` alternative of the ternary expression to `int` instead?

bertm (approved):
Never mind my previous comment, I failed to realize that this has nothing to do with the ternary expressions (which would have been surprising indeed) but rather revolves around implementing `SortedSet<Integer>`. Your changes make perfect sense.

Looks good to me.

ArneBab:
Thank you for your reviews!

PR #943, opened 2024-02-18
Counting loop was not looping

Juiceman:
Disregard.
The section of code does not do anything anymore.

According to the comments:
We always decrement on starting a sender.
However, after that, if our HTL is above the no-cache threshold, we do not want to decrement the HTL for trivial rejections (e.g. RejectedLoop), because we would end up caching data too close to the originator. So allow 5 failures and then RNF.

### DIFF
--- a/src/freenet/node/CHKInsertSender.java
+++ b/src/freenet/node/CHKInsertSender.java
@@ -525,7 +525,7 @@ public final class CHKInsertSender extends BaseSender implements PrioRunnable, A
 			}
 			
 			innerRouteRequests(next, thisTag);
-			return;
+
 		}
 	}
 
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -393,7 +393,7 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
             
             innerRouteRequests(next, origTag);
             // Will either chain back to routeRequests(), or call onAccepted().
-           	return;
+
         }
 	}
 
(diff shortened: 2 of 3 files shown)


src/freenet/node/CHKInsertSender.java
```diff
@@ -525,7 +525,7 @@ protected void routeRequests() {
 			}
 			
 			innerRouteRequests(next, thisTag);
-			return;
+
```
ArneBab, line 528: If I ever saw treacherous indentation, this is it … 

How did you test this change? This part is in the critical area of the code. If there’s any condition that could cause an infinite loop here, it could cause endlessly spinning threads.

Juiceman: I have it running on my node for a little while now. Insert threads are not climbing abnormally.
Obviously, this needs to be tested by more than just me.

Juiceman: Two more files. The one has this comment "// Will either chain back to routeRequests(), or call onAccepted()." right above the return.
With the return the code never loops over the counter - why is there a counter there if it is not being used?

Juiceman:
Closing, not a bug.

src/freenet/node/CHKInsertSender.java
ArneBab: Unused code actually is a concern: it should be removed because it makes the code harder to understand.

What we need to preserve is existing API, though, because plugins might be using it.

PR #948, opened 2024-02-26
break early when condition is met

Juiceman:
Break out of the loop once the condition is met


Bombe (approved):

bertm:
These changes do not solve the underlying problem: Fred implements basic List semantics (such as: `contains`, `add` and `remove`) on top of arrays all over the place.

Nevertheless, these changes are an improvement. Looks good to me.

ArneBab:
Thank you for your PR and reviews! Merged.

PR #949, opened 2024-02-28
🐛 Fix JarClassLoaderTest to work with newer Java versions

Bombe:
Java 11’s new, non-native `ZipFile` implementation will return the same entry if you ask for an entry with the same name twice, whereas Java 8’s native implementation gave you the _next_ entry with the given name, leading to a test that will work with Java 8 but not with Java 11.

And also, all the hubbub with multiple entries in a JAR file with the same name is totally unnecessary, the test works just fine, the fixed version doesn’t even need reflection, and it verifies `JarClassLoader`’s functionality just as well as before.

Tested with Java 8, Java 11, Java 17, and Java 21.

### DIFF
--- a/test/freenet/support/JarClassLoaderTest.java
+++ b/test/freenet/support/JarClassLoaderTest.java
@@ -3,11 +3,9 @@ package freenet.support;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ServiceLoader;
-import java.util.Set;
 import java.util.jar.JarOutputStream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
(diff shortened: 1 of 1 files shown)


ArneBab:
Very nice! Just one nitpick: never → newer?

Bombe:
> Very nice! Just one nitpick: never → newer?

Goddammit, I’m really starting to worry about myself, it happens more and more often that I simply use the wrong word. 😕

ArneBab:
> > Very nice! Just one nitpick: never → newer?
> 
> Goddammit, I’m really starting to worry about myself, it happens more and more often that I simply use the wrong word. 😕

It’s the same for me. And my typo rate skyrocketed in the past two years. Something is off there …

ArneBab (approved):

bertm (approved):
Having two ZIP entries with the same name does not even make sense. Java's URLClassLoader does not support that either. Yay for reducing test weirdness :+1: 

Speaking of URLClassLoader, I do wonder why we don't just use that one instead of rolling our own. That's not something to be solved here, though.

Bombe:
> Having two ZIP entries with the same name does not even make sense. Java's URLClassLoader does not support that either. Yay for reducing test weirdness 👍

I can’t even tell you why I thought that I needed that for the test… and I also can‘t tell you (knowing _exactly_ how the URLs enumerated by `findResources()` are created) how I thought this could _ever_ work… 🤣 (I can tell you, however, how I found out: https://blog.pterodactylus.net/2024/02/29/breaking-jarurlconnection/ 😀)


> Speaking of URLClassLoader, I do wonder why we don't just use that one instead of rolling our own. That's not something to be solved here, though.

Hmm… the history of that file does not say a lot about _why_ we did that… it does contain some stuff for fixing Windows-related file-handling issues, and some fix for plugins where I’m not sure what it’s about, and does in general seem to be quite weird. But yeah, that’s an issue for another day… 🙂

Bombe:
> Having two ZIP entries with the same name does not even make sense.

I mean, depends, I guess… sure, it makes it hard to get a specific duplicate but you can still get everything by simply enumerating all entries. ZIP tried very hard to be a streaming format, after all… 🤣

ArneBab:
Thank you for the fix! Merged.

PR #950, opened 2024-02-29
Add Getters for Public/Protected Fields Accessed within Fred

Bombe:
This pull requests adds a whole bunch of public getters (and one setter!) to a whole bunch of Fred’s core classes, such as `Node`, `NodeClientCore`, or `NodeCrypto`. It also replaces all direct accesses to the exposed fields with getters from everywhere else in the node.

### Why, though?

Because exposing and subsequently accessing fields directly is icky, for a number of reasons:

* It makes the creation of mock objects for tests (e.g. using Mockito) really hard, because simply calling `Mockito.mock()` for a class is not enough, you also have to use reflection to set fields that are used during your test (but not by the code you’re testing!). This leads to less tests being written, as well as tests that are way more complex than they need to be because they need to take care of 15 exposed fields that your change doesn’t even care about but which will break unrelated code when they are `null`.
* It ties other code to the object exposing the field. There is no way for us to update the field in a way that will not break clients of that field (like plugins).
* It also increases the API surface of the node which is a rather bad thing for software with a promise of _some_ degree of privacy, anonymity, and security. 🙂 

### But isn’t accessing a method way slower than directly accessing a field?

Short answer: no.

Long answer: also no, but longer. Yes, back when most of the core code was originally written, the JVM was as young and careless, just like us. Nowadays, though, 25 years later, the JVM has gotten so incredibly good at aggressively optimizing simple stuff like accessors that there is virtually no difference between accessing a simple getter (which is all I’m adding) and the field itself.

### Where do we go from here?

Good question! One thing that I would like to do is to replace all the objects exposed as fields with proxy objects which forward everything to real object but logging the first request (at level ERROR or at least WARNING) arriving at it, so it’s clear in the logfile which plugins are still using the fields directly so they can be fixed (or at least some public pressure can be put on the authors 😁).

After that we could work on getting some separation between plugins and the node going. We absolutely need better separation between the node and the plugins; when all access is going through methods, it’s easier to implement e.g. a permission model (i.e. plugin A is allowed to get an object, plugin B isn’t). (Hmm, technically this is possible with proxies as well but… 🥶)

Then, further into the future, we really need to re-evaluate why some code needs to do stuff like `getA().getB().getC().doSomething()`, violating the Law of Demeter _so hard_ it’s not even funny anymore. If a class needs a `C`, why doesn’t it get handed a `C` directly, forcing it instead to go through `A` and `B`? However, the efforts to get rid of all of that are probably going to be massive, and I can’t promise not to introduce factories, and maybe even factories for factories… 🤣

However, for bigger refactorings such as this we need to reduce Fred’s API surface so we have less chance of breaking plugins when we change things, and we most definitely need to change a lot of things.

### DIFF
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -634,7 +634,13 @@ public class Node implements TimeSkewDetectorCallback {
 	@Deprecated
 	/* It’s not the field that is deprecated but accessing it directly is. */
 	public final LocationManager lm;
-	/** My peers */
+
+	/**
+	 * My peers
+	 * @deprecated Use {@link #getPeers()} instead of accessing this directly.
+	 */
+	@Deprecated
+	/* It’s not the field that is deprecated but accessing it directly is. */
 	public final PeerManager peers;
 	/** Node-reference directory (node identity, peers, etc) */
 	final ProgramDirectory nodeDir;
@@ -4983,6 +4989,10 @@ public class Node implements TimeSkewDetectorCallback {
 
     public RequestTracker getTracker() {
         return tracker;
+    }
+
+    public PeerManager getPeers() {
+        return peers;
     }
 
 }
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -901,7 +901,12 @@ public class Node implements TimeSkewDetectorCallback {
 	/* It’s not the field that is deprecated but accessing it directly is. */
 	final FailureTable failureTable;
 
-	// The version we were before we restarted.
+	/**
+	 * The version we were before we restarted.
+	 * @deprecated Use {@link #getLastVersion()} instead of accessing this directly.
+	 */
+	@Deprecated
+	/* It’s not the field that is deprecated but accessing it directly is. */
 	public int lastVersion;
 
 	/** NodeUpdater **/
@@ -5213,6 +5218,10 @@ public class Node implements TimeSkewDetectorCallback {
 
     public FailureTable getFailureTable() {
         return failureTable;
+    }
+
+    public int getLastVersion() {
+        return lastVersion;
     }
 
 }
--- a/src/freenet/node/Node.java
+++ b/src/freenet/node/Node.java
@@ -4589,7 +4589,7 @@ public class Node implements TimeSkewDetectorCallback {
 		peers.addPeer(node,false,false);
 	}
 	public void connect(Node node, FRIEND_TRUST trust, FRIEND_VISIBILITY visibility) throws FSParseException, PeerParseException, ReferenceSignatureVerificationException, PeerTooOldException {
-		peers.connect(node.darknetCrypto.exportPublicFieldSet(), darknetCrypto.packetMangler, trust, visibility);
+		peers.connect(node.darknetCrypto.exportPublicFieldSet(), darknetCrypto.getPacketMangler(), trust, visibility);
 	}
 
 	public short maxHTL() {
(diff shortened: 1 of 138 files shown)


Juiceman:
I like the concept! Thank you for putting this together. No pun intended :)

I should have some time over the next few days to review the commits.

ArneBab:
I didn’t yet get through all the diff (and that will take a bit), but I already love both the change and the vision.

Though you missed mentioning that you may need a FactoryFactoryProviderProvider ☺ (not completely shitting: I know of a *ComponentFactoryConfigurationFactoryProvider — though not in fred ☺)..

Bombe:
> Though you missed mentioning that you may need a FactoryFactoryProviderProvider ☺ (not completely shitting: I know of a *ComponentFactoryConfigurationFactoryProvider — though not in fred ☺)..

Yeah, those things happen if the overall structure of a project is too convoluted, and I don’t see Fred _not_ being in the same category… there are definitely “fun” times ahead. 🤣

src/freenet/node/Node.java
```diff
@@ -4923,4 +5182,161 @@ public NodeDiagnostics getNodeDiagnostics() {
...
+
+    public NodeStats getNodeStats() {
+        return nodeStats;
+    }
+
+    public PersistentConfig getConfig() {
+        return config;
+    }
+
+    public NodeGetPubkey getGetPubKey() {
```
Juiceman, line 5194: Hehe. :)

Juiceman (approved):
You owe me some eyedrops :)

src/freenet/node/Node.java
```diff
@@ -4923,4 +5182,161 @@ public NodeDiagnostics getNodeDiagnostics() {
...
+    public NodeCrypto getDarknetCrypto() {
+        return darknetCrypto;
+    }
+
+    public Executor getExecutor() {
+        return executor;
+    }
+
+    public PacketSender getPacketSender() {
+        return ps;
```
ArneBab, line 5235: that makes code much easier to follow — thank you!

Bombe: In this particular instance it does, yes… most of it is boilerplate, though.

src/freenet/node/PeerNode.java
```diff
@@ -2255,19 +2255,19 @@ void maybeSendInitialMessages() {
 	protected void sendInitialMessages() {
 		loadSender(true).setSendASAP();
 		loadSender(false).setSendASAP();
-		Message locMsg = DMT.createFNPLocChangeNotificationNew(node.lm.getLocation(), node.peers.getPeerLocationDoubles(true));
+		Message locMsg = DMT.createFNPLocChangeNotificationNew(node.getLocationManager().getLocation(), node.getPeers().getPeerLocationDoubles(true));
 		Message ipMsg = DMT.createFNPDetectedIPAddress(detectedPeer);
 		Message timeMsg = DMT.createFNPTime(System.currentTimeMillis());
 		Message dRoutingMsg = DMT.createRoutingStatus(!disableRoutingHasBeenSetLocally);
-		Message uptimeMsg = DMT.createFNPUptime((byte)(int)(100*node.uptime.getUptime()));
+		Message uptimeMsg = DMT.createFNPUptime((byte)(int)(100*node.getUptimeEstimator().getUptime()));
```
ArneBab, line 2262: that finally makes it clear what happens here! :heart:

src/freenet/node/updater/UpdateOverMandatoryManager.java
```diff
@@ -1158,14 +1158,14 @@ public RequestClient getRequestClient() {
 	private void cancelSend(PeerNode source, long uid) {
 		Message msg = DMT.createFNPBulkReceiveAborted(uid);
 		try {
-			source.sendAsync(msg, null, updateManager.ctr);
+			source.sendAsync(msg, null, updateManager.getByteCounter());
```
ArneBab, line 1161: much nicer!

ArneBab (approved):

ArneBab:
Merged — thank you!

PR #952, opened 2024-03-16
Reduce recently failed and data not found wait times

ArneBab:
Since we have 5 minutes time to receive a USK update, RF should be at most that long. Otherwise RF will block keys if the uploader checked them a bit too early.


src/freenet/node/FailureTable.java
```diff
@@ -74,12 +74,12 @@ public void shouldUpdate(){
 	/** Terminate a request if there was a DNF on the same key less than 10 minutes ago.
```
Juiceman, line 74: Please update the time in the documentation as well.

ArneBab: I removed the value itself from the comment: that was needless duplication.

Thank you for catching it!

Juiceman (approved):

Juiceman (approved):

ArneBab:
Thank you! Merged!

PR #954, opened 2024-03-18
Boxing and unboxing

Juiceman:
Java > 1.4 handles autoboxing


src/freenet/client/DefaultMIMETypes.java
```diff
@@ -65,14 +65,14 @@ protected static synchronized void addMIMEType(short number, String type) {
 	 */
 	protected static synchronized void addMIMEType(short number, String type, String[] extensions, String outExtension) {
 		addMIMEType(number, type);
-		Short t = Short.valueOf(number);
+		Short t = number;
```
Bombe, line 68: Looks like we don’t need `t` at all, do we? We can use `number` in every place that `t` is used.

src/freenet/client/DefaultMIMETypes.java
```diff
@@ -119,7 +119,7 @@ public synchronized static String byNumber(short x) {
 	 */
 	public synchronized static short byName(String s) {
```
Bombe, line 120: This method looks like it can be replaced by

```java
return mimeTypesByName.getOrDefault(s, -1);
```

src/freenet/io/comm/MessageCore.java
```diff
@@ -616,9 +616,9 @@ public Map<String, Integer> getUnclaimedFIFOMessageCounts() {
 				String messageName = m.getSpec().getName();
 				Integer messageCount = messageCounts.get(messageName);
```
Bombe, line 617: This looks like a prime contender for `Map.merge()`…

src/freenet/node/NodeDispatcher.java
```diff
@@ -846,7 +846,7 @@ public void run() {
 	private boolean handleRoutedRejected(Message m) {
 		if(!node.enableRoutedPing()) return true;
 		long id = m.getLong(DMT.UID);
-		Long lid = Long.valueOf(id);
+		Long lid = id;
```
Bombe, line 849: We don’t need `lid` if we can use `id`, right?

src/freenet/node/NodeDispatcher.java
```diff
@@ -886,7 +886,7 @@ boolean handleRouted(Message m, PeerNode source) {
 		if(logMINOR) Logger.minor(this, "handleRouted("+m+ ')');
 
 		long id = m.getLong(DMT.UID);
-		Long lid = Long.valueOf(id);
+		Long lid = id;
```
Bombe, line 889: Probably also redundant.

src/freenet/node/NodeDispatcher.java
```diff
@@ -930,7 +930,7 @@ boolean handleRoutedReply(Message m) {
 		if(!node.enableRoutedPing()) return true;
 		long id = m.getLong(DMT.UID);
 		if(logMINOR) Logger.minor(this, "Got reply: "+m);
-		Long lid = Long.valueOf(id);
+		Long lid = id;
```
Bombe, line 933: Here as well.

src/freenet/node/PeerManager.java
```diff
@@ -1789,8 +1789,8 @@ public void maybeLogPeerNodeStatusSummary(long now) {
 	
 	public void changePeerNodeStatus(PeerNode peerNode, int oldPeerNodeStatus,
 			int peerNodeStatus, boolean noLog) {
-		Integer newStatus = Integer.valueOf(peerNodeStatus);
-		Integer oldStatus = Integer.valueOf(oldPeerNodeStatus);
+		Integer newStatus = peerNodeStatus;
```
Bombe, line 1792: These here smell redundant as well. The two further down in this file do, too.

src/freenet/node/PeerNode.java
```diff
@@ -3258,7 +3258,7 @@ public void addToLocalNodeSentMessagesToStatistic(Message m) {
 			if(count == null)
```
Bombe, line 3258: `Map.merge()` calls again, for the one below as well.

test/freenet/support/JVMVersionTest.java
```diff
@@ -83,7 +83,7 @@ public void testCompare() {
 	        String v1 = orderedVersions[i];
 	        for (int j = 0; j < orderedVersions.length; j++) {
 	            String v2 = orderedVersions[j];
-	            int expected = Integer.signum(Integer.valueOf(i).compareTo(Integer.valueOf(j)));
+	            int expected = Integer.signum(Integer.valueOf(i).compareTo(j));
```
Bombe, line 86: `Integer` has a static `compare` method that you can use instead of still manually boxing one of the `int`s.

Bombe (changes requested):

src/freenet/node/NodeDispatcher.java
Juiceman: Fixed

src/freenet/io/comm/MessageCore.java
Juiceman: I will look into it another time. Please don't hold up the PR waiting for me.

src/freenet/node/PeerNode.java
Juiceman: I will look into it another time. Please don't hold up the PR waiting for me.

ArneBab:
@Bombe is this ready to be merged now?

Bombe:
GitHub’s interface makes it exceedingly difficult to see whether everything was addressed but I do seem to have checked the “verified” checkbox on every single file so I guess this is good now.

ArneBab:
@Bombe thank you for re-checking! I’ll merge this after releasing 1498.

ArneBab:
Merged — thank you!

PR #956, opened 2024-04-01
Update NEWS for 1498

ArneBab:
It’s actually quite a huge change.


Bombe:
Yeah, this is why nobody uses commit messages to create changelos. 😄

Honestly, most of these are useless without context.

ArneBab:
Yes — I’ll have to re-organize that and turn it into an article for the changelog we insert and put on the website; as usual ☺

ArneBab:
I now grouped it by topic, the second step towards writing release news. It’s always crazy how long that takes …

Bombe (approved):

PR #958, opened 2024-04-26
Revert "Only link indexes via the index of indexes."

ArneBab:
This reverts commit fda9bdb81d4077747db365a922ee7afe30f10584.

There’s a contingency plan now to enable updating the index in case its uploader becomes inactive or cannot react when required by law. It needs multiple core devs working together, but should be fast enough.


Spider-Admin:
Sorry, I am not familiar with "Pull requests", so tested them in my own way:

I stopped my node, created a backup of my `bookmarks.dat` and copied the new `defaultbookmarks.dat` as `bookmarks.dat`. The links for "Clean-Spider" and "Index of Indexes" were automatically updated to their latest edition. Both links work fine.

ArneBab:
Rebased locally and pushed into next.

PR #959, opened 2024-04-28
Resilience against plugin list exception

ArneBab:
This is just a band-aid. We may want to copy the plugin list instead.

### DIFF
--- a/src/freenet/pluginmanager/PluginManager.java
+++ b/src/freenet/pluginmanager/PluginManager.java
@@ -637,7 +637,7 @@ public class PluginManager {
 
 	public void cancelRunningLoads(String filename, PluginProgress exceptFor) {
 		Logger.normal(this, "Cancelling loads for plugin "+filename);
-		for (PluginProgress progress : new ArrayList<PluginProgress>(loadedPlugins.getStartingPlugins())) {
+		for (PluginProgress progress : loadedPlugins.getStartingPlugins()) {
 			if ((progress != exceptFor) && filename.equals(progress.name)) {
 				progress.kill();
 				loadedPlugins.removeStartingPlugin(progress);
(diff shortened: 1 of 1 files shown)


Bombe:
Oof, yeah, that’s horrible. `getLoadedPlugins()` needs to make a copy. Actually, all the getters need to make copies. Who wrote that? (Oh yeah, it was me.)

ArneBab:
@Bombe I changed the getters to only return copies of the loaded and loading lists. They are only iterated over at all callers. And I found one instance where Toad was being rightfully paranoid about this :-) — now no longer needed.

Is this good to go?

ArneBab:
merged — thank you for the review!

src/freenet/pluginmanager/PluginManager.java
```diff
@@ -1601,10 +1608,15 @@ public void addStartingPlugin(PluginProgress pluginProgress) {
...
 		}
 
+		/**
+		 * @return a copy of the starting plugins. Do not modify this: modifications will get thrown away.
+		 */
 		public Collection<PluginProgress> getStartingPlugins() {
+			Set<PluginProgress> startingPluginsCopy;
 			synchronized (this) {
-				return startingPlugins;
+				startingPluginsCopy = Set.copyOf(startingPlugins);
```
torusrxxx, line 1617: Set.copyOf requires Java 10.
This is an error when using Java 8.

ArneBab: Thank you! I fixed the setting of the IDE and pushed a change into next right-away (because this broke the build for Java 8 users whom we still support): https://github.com/hyphanet/fred/commit/4c3ffb183452314cc777d69a90418a73d095adf8

PR #963, opened 2024-05-10
Support abridged IPv6 address

torusrxxx:
Compacted IPV6 address like `::1` and `2001:db8::1234` can now be used. Connections page show IPV6 address like `[address]:port`.

### DIFF
--- a/src/freenet/io/AddressIdentifier.java
+++ b/src/freenet/io/AddressIdentifier.java
@@ -53,7 +53,6 @@ public class AddressIdentifier {
 		 */
 		String ipv6AddressRegex = "::(?>(?>X:){0,6}X)?|X::(?>(?>X:){0,5}X)?|X:X::(?>(?>X:){0,4}X)?|X:X:X::(?>(?>X:){0,3}X)?|(?>X:){4}:(?>(?>X:){0,2}X)?|(?>X:){5}:(?>X:)?X?|(?>X:){6}:X?|(?>X:){7}(?>X|:)";
 		ipv6AddressRegex = ipv6AddressRegex.replaceAll("X", wordRegex);
-		//String ipv6ISATAPAddressRegex = wordRegex + "?:" + wordRegex + ':' + wordRegex + ':' + wordRegex + ":(0){1,4}:5[eE][fF][eE]:" + wordRegex + ':' + wordRegex + percentScopeIDRegex;
 		// case 0: :(?>(?>:X){1,3}:0{1,4}|:0{1,4}|):5EFE:X:X
 		// case 1: X:(?>:0{1,4}|:(?>X:){1,2}0{1,4}|):5EFE:X:X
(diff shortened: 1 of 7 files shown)


ArneBab:
This looks good to me.

src/freenet/io/AddressIdentifier.java
```diff
@@ -38,13 +39,31 @@ public class AddressIdentifier {
...
+		 * X:X::(?>(?>X:){0,4}X)?
+		 * X:X:X::(?>(?>X:){0,3}X)?
+		 * (?>X:){4}:(?>(?>X:){0,2}X)?
+		 * (?>X:){5}:(?>X:)?X?
+		 * (?>X:){6}:X?
+		 * (?>X:){7}(?>X|:)
+		 */
+		String ipv6AddressRegex = "::(?>(?>X:){0,6}X)?|X::(?>(?>X:){0,5}X)?|X:X::(?>(?>X:){0,4}X)?|X:X:X::(?>(?>X:){0,3}X)?|(?>X:){4}:(?>(?>X:){0,2}X)?|(?>X:){5}:(?>X:)?X?|(?>X:){6}:X?|(?>X:){7}(?>X|:)";
+		ipv6AddressRegex = ipv6AddressRegex.replaceAll("X", wordRegex);
+		//String ipv6ISATAPAddressRegex = wordRegex + "?:" + wordRegex + ':' + wordRegex + ':' + wordRegex + ":(0){1,4}:5[eE][fF][eE]:" + wordRegex + ':' + wordRegex + percentScopeIDRegex;
```
Bombe, line 56: Please remove the old code altogether.

src/freenet/io/Inet6AddressMatcher.java
```diff
@@ -66,18 +64,67 @@ public Inet6AddressMatcher(String pattern) {
...
+					addressBytesEnd[endCount * 2 + 1] = (byte) (addressWord & 0xff);
+					endCount++;
+				}
+			} else if(endCount == -1) {
+				endCount = 0;
+			} else {
+				throw new IllegalArgumentException(address + " is not an IPv6 address.");
+			}
+		}
+		if(endCount != -1 && count + endCount >= 8) { // Maybe not needed
```
Bombe, line 116: What do you mean, “maybe?” 😄 

You have the tests (that look rather exhaustive), when running them with coverage you should see immediately if this code is necessary.

Bombe (changes requested):

src/freenet/io/Inet6AddressMatcher.java
torusrxxx: I'm sure it is not needed, but there's no harm staying there, just in case. I have removed this part of code as you suggests.
The tests cannot enumerate all possible invalid inputs, and indeed some were missed, for example, IPV6 address with percentage scope and 1:2:3:4:5:6:7: is broken with AddressIdentifier and Inet6AddressMatcher respectively. I have fixed those bugs and added them into the test suite as well.

PR #964, opened 2024-05-10
When the update key is changed, subscribe to the new key.

ArneBab:


src/freenet/node/updater/NodeUpdater.java
```diff
@@ -511,8 +511,14 @@ public synchronized boolean canUpdateNow() {
 	/** Called when the fetch URI has changed. No major locks are held by caller. 
 	 * @param uri The new URI. */
 	public void onChangeURI(FreenetURI uri) {
-		kill();
+		kill(); // unsubscribes from the old uri
 		this.URI = uri;
+		try {
```
Bombe, line 516: This looks suspiciously like the `start()` method. Maybe we should simply call that method here?

Bombe (changes requested):

src/freenet/node/updater/NodeUpdater.java
ArneBab: That’s a good point. It’s not exactly the same (we won’t want to stop updating when the key is invalid), but very close. I factored out a subscribe() method.

src/freenet/node/updater/NodeUpdater.java
```diff
@@ -91,15 +91,23 @@ public abstract class NodeUpdater implements ClientGetCallback, USKCallback, Req
 	}
 
 	void start() {
+        try {
+            this.subscribe();
```
Bombe, line 95: The `this` qualifier is over-doing it a bit, methinks. 😄

src/freenet/node/updater/NodeUpdater.java
```diff
@@ -91,15 +91,23 @@ public abstract class NodeUpdater implements ClientGetCallback, USKCallback, Req
...
+
+    void subscribe() throws MalformedURLException {
 		try {
 			// because of UoM, this version is actually worth having as well
 			USK myUsk = USK.create(URI.setSuggestedEdition(currentVersion));
 			core.getUskManager().subscribe(myUsk, this, true, getRequestClient());
 		} catch(MalformedURLException e) {
 			Logger.error(this, "The auto-update URI isn't valid and can't be used");
-			manager.blow("The auto-update URI isn't valid and can't be used", true);
+            throw e;
```
Bombe, line 108: This irks me… catching an exception, just to log it, then re-throw it and ignore it in one of the methods. That feels like a lot of work.

Hmm, oh boy, here I go, making it overly complicated again… but what about:

```
private void subscribe(Runnable onError) {
    try {
    } catch (MalformedURLException e) {
        Logger.error(this, "The auto-update URI isn't valid and can't be used");
        onError.run();
    }
}
```
so we can have
```
void start() {
    subscribe(() -> manager.blow("The auto-update URI isn't valid and can't be used", true));
}
```
and
```
public void onChangeURI(FreenetURI uri) {
    kill();
    this.URI = uri;
    subscribe(() -> {});
    maybeUpdate();
}
```
?

Bombe (changes requested):

src/freenet/node/updater/NodeUpdater.java
```diff
@@ -91,15 +91,23 @@ public abstract class NodeUpdater implements ClientGetCallback, USKCallback, Req
...
 
 	void start() {
+        try {
+            this.subscribe();
+		} catch(MalformedURLException e) {
+            manager.blow("The auto-update URI isn't valid and can't be used", true);
+        }
+	}
+
+    void subscribe() throws MalformedURLException {
```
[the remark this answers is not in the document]
ArneBab: Not a good reason, no — fixed. Thanks!

src/freenet/node/updater/NodeUpdater.java
ArneBab: Oh yes — I blame Javascript ☺

Thank you!

src/freenet/node/updater/NodeUpdater.java
ArneBab: I wanted to say “no, this makes it too non-obvious”, but it’s actually much nicer with the Runnable.

So: yes ☺

Thank you!

Bombe (approved):

ArneBab:
merged — thank you for your review!

PR #966, opened 2024-05-19
Hide new load management statistics

torusrxxx:
This empty box looks like something's wrong. Hide it when new load management is disabled.


ArneBab (approved):
Looks good, though it could go one step further and just remove the NLM statistics code, since NLM is broken and statically disabled (and cannot be enabled anymore).

But NLM may be salvageable: it was given up after failing once and the bugs that broke it have been fixed since.

⇒ change looks good to me.

Bombe (approved):

ArneBab:
merged — thank you!

PR #968, opened 2024-06-12
remove recommendation to turn off javascript helpers in fproxy.

ArneBab:
It is actually safe, because the JS is reviewed.

JS on the web is unsafe, because it runs untrusted code. That’s not the case here.


Bombe (approved):

ArneBab:
merged — thank you!

PR #970, opened 2024-07-25
Provide text with UTF-8 MIME Type by default

ArneBab:
This avoids very common text encoding problems.


Bombe:
You have not avoided the very common test writing problem! 😄

ArneBab:
> You have not avoided the very common test writing problem! 😄

I had also not avoided the very common "my change does not have any effect and a test would have shown that" problem :sweat: 

Now it’s fixed: our plain text filter actually detects the charset from the BOM and uses UTF-8 by default.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -368,6 +361,16 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
 			}
 
 			if(handler.safeToRead) {
+				if(handler.takesACharset && ((charset == null) || (charset.isEmpty()))) {
+					byte[] charsetBuffer = new byte[CHARSET_DETECTION_FALLBACK_BUFFERSIZE];
+					int offset = readIntoBuffer(input, CHARSET_DETECTION_FALLBACK_BUFFERSIZE, charsetBuffer);
+					BOMDetection bom = CSSReadFilter.detectCharsetFromBOM(charsetBuffer, CHARSET_DETECTION_FALLBACK_BUFFERSIZE);
```
Bombe, line 367: I’m pretty sure this is 100% wrong. That method detects an encoding from the representation of the string `@charset`. It is also gloriously misnamed as it has nothing to do with a BOM. 😀

bertm:
Both new tests fail on my machine, not sure why they work on CI.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -343,16 +344,8 @@ public static FilterStatus filter(InputStream input, OutputStream output, String
...
-					int bytesRead = 0, offset = 0, toread=0;
-					while(true) {
-						toread = bufferSize - offset;
-						bytesRead = input.read(charsetBuffer, offset, toread);
-						if(bytesRead == -1 || toread == 0) break;
-						offset += bytesRead;
-					}
-					input.reset();
+					int offset = readIntoBuffer(input, bufferSize, charsetBuffer);
 					charset = detectCharset(charsetBuffer, offset, handler, maybeCharset);
```
bertm, line 349: I believe the correct solution to this problem is moving this block of code right before the `if(handler.readFilter != null)` check: text/plain does not have a `readFilter`, but does `takesACharset` so this would run the `detectCharset` appropriately.

Few things to consider:
* `handler.charsetExtractor.getCharsetBufferSize()` will NPE, so we need to choose the bufferSize to the max BOM length (5?) when `handler.charsetExtractor` is absent.
* Alternatively a dummy CharsetExtractor can be used that always fails to detect a charset, so the BOM one is used automagically.
* this will return `UTF-8` rather than `utf-8` so the related test would need some adjustment.

src/freenet/client/filter/ContentFilter.java
bertm: Yep, I agree with @Bombe  here. See my other comment for something that does appear to work.

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -380,6 +380,37 @@ public void testEvilCharset() throws IOException {
         }
     }
 
+
+    @Test
+    public void byteOrderMarkForUtf8IsDetectedCorrectly() throws IOException {
+        byte[] buf = { (byte) 0xef, (byte) 0xbb, (byte) 0xbf, 0x40 };
+        ArrayBucket out = new ArrayBucket();
+        FilterStatus fo = ContentFilter.filter(new ArrayBucket(buf).getInputStream(), out.getOutputStream(), "text/plain", null, null, null);
+        assertTrue("utf-8".equals(fo.charset));
```
bertm, line 389: Please use `assertThat(actual, equalTo(expected))` or `assertEquals(expected, actual)` for checking equality - this just yields a non-descriptive AssertionError when it fails instead of showing what the actual value was.

bertm (changes requested):

PR #971, opened 2024-07-25
Remove no longer needed garbage collector meddling.

ArneBab:
We now have far more memory than before and file sizes also increased.

I’m not sure whether we could just delete the whole gc code, though, and trust the JVM.


Bombe:
> I’m not sure whether we could just delete the whole gc code, though, and trust the JVM.

We absolutely should, especially because that code is not guaranteed to actually trigger a GC. The only way to force the JVM to do a garbage collection is by exhausting all memory; when the `OutOfMemoryError` is thrown, a garbage collection will have happened.

ArneBab:
@Bombe I now removed that code and also the aggressiveGC option that was only used there.

Also Hyphanet no longer dies during startup when an ignored option is used during startup (but I’m not happy about the error reporting: do we have a way to report this very loudly?).

src/freenet/node/MemoryChecker.java
```diff
@@ -14,8 +14,7 @@ public class MemoryChecker implements Runnable {
 	private volatile boolean goon = false;
 	private final Ticker ps;
 	private int aggressiveGCModificator;
-	private RunningAverage avgFreeMemory;
-	
+
```
Bombe, line 17: The field `aggressiveGCModificator` can be removed, it’s always `-1` now.

src/freenet/node/MemoryChecker.java
```diff
@@ -48,61 +47,14 @@ public void run() {
...
-				} catch (InterruptedException e) {
-				}
-
-				freeMemory = r.freeMemory();
-				avgFreeMemory.report(freeMemory);
-			}
-		}
-		
 		int sleeptime = aggressiveGCModificator;
 		if(sleeptime <= 0) { // We are done
```
Bombe, line 54: `aggressiveGCModificator` going away means this also doesn’t need an `if` anymore.

src/freenet/node/NodeStats.java
```diff
@@ -407,21 +407,7 @@ public void set(Integer val) throws InvalidConfigValueException {
 		threadLimit = statsConfig.getInt("threadLimit");
 
 		// Yes it could be in seconds insteed of multiples of 0.12, but we don't want people to play with it :)
```
Bombe, line 409: The comment can also be removed. 🙂

src/freenet/node/NodeStats.java
```diff
@@ -407,21 +407,7 @@ public void set(Integer val) throws InvalidConfigValueException {
...
-						if (get().equals(val))
-					        return;
-						Logger.normal(this, "Changing aggressiveGCModificator to "+val);
-						aggressiveGCModificator = val;
-					}
-		},false);
-		aggressiveGCModificator = statsConfig.getInt("aggressiveGC");
+		statsConfig.registerIgnoredOption("aggressiveGC");
 
 		myMemoryChecker = new MemoryChecker(node.getTicker(), aggressiveGCModificator);
```
Bombe, line 412: `aggressiveGCModificator` should also be removed from this class, including the comment above its declaration. 🙂

Bombe (changes requested):

src/freenet/config/SubConfig.java
```diff
@@ -172,47 +172,53 @@ public int getInt(String optionName) {
 		synchronized(this) {
 			o = (IntOption) map.get(optionName);
 		}
-		return o.getValue();
+		// return fallback value for ignored options (null). This avoids breaking plugins which try to get ignored options.
+		return o == null ? -1 : o.getValue();
```
Bombe, line 176: Do you know of any plugins that fail because of the addition of ignored options?

ArneBab: My fred broke during development because I had a place left that accessed the removed option.

I don’t know about plugins that get these options, but this change enables us to set options as ignored without having to worry about plugins.

bertm, line 176: Can we replace the ternaries with the more descriptive `Map.getOrDefault`?

src/freenet/node/MemoryChecker.java
bertm, line 1: Does this class still serve any purpose (aside from scheduling itself :smile:) after removing pretty much all of its code?

Bombe: Haha, that is an excellent observation! 😄 Kill it with 🔥! 🥳

src/freenet/config/SubConfig.java
ArneBab: I tried: not without increasing the complexity, because map returns an option which does not have a simple constructor from just a value.

src/freenet/node/NodeStats.java
ArneBab: Done — thanks!

src/freenet/node/MemoryChecker.java
ArneBab: It’s killed with :fire:  ☺

Good catch!

src/freenet/config/SubConfig.java
bertm: Ah, yeah. I kind of forgot about the mess the constructor of Option is :disappointed: 

Spoiler:
```java
Option(SubConfig config, String name, ConfigCallback<T> cb, int sortOrder, boolean expert, boolean forceWrite,
	String shortDesc, String longDesc, DataType dataType) {
```

Well then let's stick with the ternaries!

bertm (approved):

ArneBab:
Thank you for the reviews! Merged!

PR #973, opened 2024-08-14
Refactor methods for separation of concern

anonstudy:
Changes proposed in this pull request:

## Announcer.Java

This PR refactors the enoughPeers method by extracting the logic for shutting down the announcement into a separate method.

Changes:
- Created a new private method `shutdownAnnouncement` in the class 
- Updated the `enoughPeers` method to call `shutdownAnnouncement` where appropriate

Rationale:
- Single Responsibility Principle: The new method has a clear, single purpose of handling the announcement shutdown process.
- Easier maintenance: Changes to the announcement shutdown logic can now be made in one place.
- Potential for reuse: If announcement shutdown is needed elsewhere, it can now be easily called from other methods.

Note:
- This refactoring does not change any functionality; it only reorganizes the existing code.

## DarknetPeerNode.Java

This PR refactors the readExtraPeerDataFile method by extracting the file reading logic into a separate method.

Changes:
- Created a new private method `readFile` that takes a file path as input and returns the file contents as a string
- Updated readExtraPeerDataFile to use the new readFile method

Rationale:
- Improved separation of concerns: The readExtraPeerDataFile method now focuses on data processing, while file I/O is handled separately.
- Increased reusability: The readFile method can be used in other parts of the codebase where file reading is needed.
- Better error handling: File I/O errors can be managed consistently in one place.

Note:
- This refactoring does not change the overall functionality; it reorganizes the existing code for better structure.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
--- a/src/freenet/node/Announcer.java
+++ b/src/freenet/node/Announcer.java
@@ -359,6 +359,15 @@ public class Announcer {
 			}
 			return true;
 		}
+		if (shutdownAnnouncement()) return true;
+
+		synchronized(timeGotEnoughPeersLock) {
+			timeGotEnoughPeers = -1;
+		}
+		return false;
+	}
+
+	private boolean shutdownAnnouncement() {
 		boolean killAnnouncement = false;
 		if((!node.getNodeUpdater().isEnabled()) ||
 				(node.getNodeUpdater().canUpdateNow() && !node.getNodeUpdater().isArmed())) {
@@ -382,7 +391,7 @@ public class Announcer {
 			}
 
 		}
-		
+
 		if(killAnnouncement) {
 			node.getExecutor().execute(new Runnable() {
 
@@ -395,7 +404,7 @@ public class Announcer {
 						node.getPeers().disconnectAndRemove(pn, true, true, true);
 					}
 				}
-				
+
 			});
 			return true;
 		} else {
@@ -410,10 +419,6 @@ public class Announcer {
 				// No point announcing at the moment, but we might need to if a transfer falls through.
 				return true;
 			}
-		}
-		
-		synchronized(timeGotEnoughPeersLock) {
-			timeGotEnoughPeers = -1;
 		}
 		return false;
 	}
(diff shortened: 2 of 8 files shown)


src/freenet/node/DarknetPeerNode.java
```diff
@@ -510,12 +510,34 @@ public boolean readExtraPeerDataFile(File extraPeerDataFile, int fileNumber) {
 			return false;
 		}
 		Logger.normal(this, "extraPeerDataFile: "+extraPeerDataFile.getPath());
+		FileInputStream fis;
+		SimpleFieldSet fs = readFile(extraPeerDataFile);
+		if (fs == null) return false;
+		if(fs == null) {
```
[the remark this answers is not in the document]
Bombe: Also, wait, _what?_ `fs` has already been found to be non-`null` so this `if` can never be `true`.

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
```
ArneBab, line 3: The code changes look good, but please remove the intellij setup from the PR.

ArneBab (changes requested):
The code changes look good, but please remove the intellij setup from the PR.

src/freenet/node/Announcer.java
```diff
@@ -359,6 +359,15 @@ boolean enoughPeers() {
...
 		}
+		if (shutdownAnnouncement()) return true;
+
+		synchronized(timeGotEnoughPeersLock) {
+			timeGotEnoughPeers = -1;
+		}
+		return false;
+	}
+
+	private boolean shutdownAnnouncement() {
```
Bombe, line 370: “shutdownAnnouncement” sounds like a command but the method doesn’t do anything, and it returns a value. What is actually meant here? Is the shutdown announcement being shown? Should it be? Should announcements be shut down? Please rename the method in a way that makes it clearer what it does.

src/freenet/node/Announcer.java
```diff
@@ -359,6 +359,15 @@ boolean enoughPeers() {
 			}
 			return true;
 		}
+		if (shutdownAnnouncement()) return true;
```
Bombe, line 362: Always use braces for blocks, and only one statement per line, please.

Bombe (changes requested):

Bombe:
> The code changes look good, thank you!

Sorry for barging into this, apparently I had started a review a couple of weeks ago but never got around to finish it. :)

ArneBab:
> > The code changes look good, thank you!
> 
> Sorry for barging into this, apparently I had started a review a couple of weeks ago but never got around to finish it. :)

I reviewed just the change — the refactoring — not the original code. There are a lot more issues in the original code than just the ones fixed here, but I think they can be fixed in an independent PR.

ArneBab:
@anonstudy are you planning to still do changes for the review by @Bombe ?

PR #975, opened 2024-09-01
🐛 Run Git from the project directory

Bombe:
When using the gradle-daemon to build a project, the working directory the script is executed in is the local directory of the gradle-daemon running the script. By explicitely specifying the directory to execute Git in, this will now work with and without gradle-daemon.


ArneBab (approved):

ArneBab:
merged — thank you! :heart:

PR #976, opened 2024-09-02
Build JAR File Correctly

Bombe:
The default `jar` task a pre-defined input with `compileJava`’s output, which includes the original, non-replaced `Version.class`. I have not found a way to disable the pre-defined input, or modify it in a way that would still allow me to add a `Version.class` from a different task’s output.

What’s happening now is that there is a new task, `buildJar`. It is an almost verbatim copy of what the `jar` task previously contained with an added input (`compileVersion`’s output). This `Jar` task does not have a pre-defined input and as such only includes what is being specified.

The original `jar` task has been disabled and has been given the `buildJar` task as dependency. That way the `jar` task can still be run from the command line where it doesn’t do anything (because it’s disabled) but it will require the `buildJar` task to be run which then does the actual work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,35 +71,35 @@ tasks.withType(Javadoc) {
     options.encoding = "UTF-8"
 }
 
-task compileVersion (type: JavaCompile) {
-    copy {
-        from sourceSets.main.java.srcDirs
-        into version_buildir
-        include "${version_src}"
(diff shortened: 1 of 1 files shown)


ArneBab (approved):
the code looks good for me.

ArneBab:
merged — thank you! :heart:

PR #977, opened 2024-09-14
Gradle update to 8.9

ArneBab:
this builds on https://github.com/hyphanet/fred/pull/972


build.gradle
```diff
@@ -250,36 +253,36 @@ dependencies {
     implementation "org.unbescape:unbescape:1.1.6.RELEASE"
     implementation "org.slf4j:slf4j-api:1.7.25"
 
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.13'
     testImplementation "org.mockito:mockito-core:1.9.5"
-    testImplementation "org.hamcrest:hamcrest-library:1.3"
-    testImplementation "org.hamcrest:hamcrest-core:1.3"
+    testImplementation "org.hamcrest:hamcrest:2.2"
+    testImplementation "org.hamcrest:hamcrest-core:2.2"
```
Bombe, line 259: These are not the latest versions. `junit:junit` is at 4.13.2, and `org.hamcrest:hamcrest` is at 3.0 (and doesn’t need the split libraries anymore).

Bombe (changes requested):

build.gradle
Bombe: > I updated junit now, but hamcrest caused problems with validation.

```
testImplementation('junit:junit:4.13.2') {
    exclude group: 'org.hamcrest', module: 'hamcrest-core'
}
testImplementation("org.mockito:mockito-core:1.9.5") {
    exclude group: 'org.hamcrest', module: 'hamcrest-core'
}
```

And maybe we should decide at some point which kind of Strings we want to use in `build.gradle`… 😄

build.gradle
```diff
@@ -250,36 +253,35 @@ dependencies {
     implementation "org.unbescape:unbescape:1.1.6.RELEASE"
     implementation "org.slf4j:slf4j-api:1.7.25"
 
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.13.2'
```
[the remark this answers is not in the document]
ArneBab: I restored those for next. thank you!

ArneBab:
merged — thank you for your reviews!

PR #982, opened 2024-09-25
Add WAV content filter

torusrxxx:
WAV PCM audio is such a simple and safe format. RIFF format is supported so why not WAV files?


Bombe:
That test needs to be at least a dozen tests, and _none_ of them should be green if you remove all the code from the filter! 😄

src/freenet/client/DefaultMIMETypes.java
```diff
@@ -755,6 +755,9 @@ public synchronized static short byName(String s) {
 		addMIMEType((short)620, "audio/ogg", "oga");
 		addMIMEType((short)621, "audio/flac", "flac");
 		addMIMEType((short)622, "image/webp", "webp");
+		addMIMEType((short)623, "image/avif", "avif");
+		addMIMEType((short)624, "image/heic", "heic");
+		addMIMEType((short)625, "image/heif", "heif");
```
bertm, line 760: These change appear to be unrelated to this PR.

src/freenet/client/filter/ContentFilter.java
```diff
@@ -123,6 +124,11 @@ false, false, new M3UFilter(), false, false, false, false, false, false,
 		register(new FilterMIMEType("audio/mpeg", "mp3", new String[] {"audio/mp3", "audio/x-mp3", "audio/x-mpeg", "audio/mpeg3", "audio/x-mpeg3", "audio/mpg", "audio/x-mpg", "audio/mpegaudio"},
 				new String[0], true, false, new MP3Filter(), true, true, false, true, false, false,
 				l10n("audioMP3ReadAdvice"), false, null, null, false));
+		
+		// WAV - has a filter
+		register(new FilterMIMEType("audio/vnd.wave", "mp3", new String[] {"audio/vnd.wave", "audio/x-wav", "audio/wav", "audio/wave"},
```
bertm, line 129: This mentions `"mp3"` even though it is about WAV.

src/freenet/client/filter/WAVFilter.java
```diff
@@ -0,0 +1,124 @@
...
+	private final int WAVE_FORMAT_IEEE_FLOAT = 3;
+	private final int WAVE_FORMAT_ALAW = 6;
+	private final int WAVE_FORMAT_MULAW = 7;
+	
+	@Override
+	protected byte[] getChunkMagicNumber() {
+		return new byte[] {'W', 'A', 'V', 'E'};
+	}
+	
+	class WAVFilterContext {
```
bertm, line 23: This class is not used elsewhere, please make it private.

Additionally, the fields in this class should not be declared `public` as they are not effectively accessible outside this class.
Either make them `private`, or of you prefer a more compact look, make them package-private (by just omitting `public`).

test/freenet/client/filter/WAVFilterTest.java
```diff
@@ -0,0 +1,89 @@
...
+			OutputStream outStream = output.getOutputStream()
+		) {
+            if (expected != null) {
+                assertThrows(expected, () -> objWAVFilter.readFilter(inStream, outStream, "", null, null, null));
+            } else {
+            	objWAVFilter.readFilter(inStream, outStream, "", null, null, null);
+            }
+		}
+		return output;
+	}
```
bertm, line 88: Indentation is off in various places in this class. Please either use 4 spaces or 1 tab for indentation, and don't mix both styles.

src/freenet/client/filter/WAVFilter.java
```diff
@@ -0,0 +1,124 @@
...
+			throw new DataFilterException(l10n("invalidTitle"), l10n("invalidTitle"), "Unexpected header chunk was encountered, instead of fmt chunk");
+		}
+		if(ID[0] == 'd' && ID[1] == 'a' && ID[2] == 't' && ID[3] == 'a') {
+			if(ctx.format == WAVE_FORMAT_PCM || ctx.format == WAVE_FORMAT_IEEE_FLOAT || ctx.format == WAVE_FORMAT_ALAW || ctx.format == WAVE_FORMAT_MULAW) {
+				// Safe format, pass through
+				output.write(ID);
+				writeLittleEndianInt(output, size);
+				passthroughBytes(input, output, size);
+				if((size & 1) != 0) // Add padding if necessary
+					output.writeByte(input.readByte());
```
bertm, line 92: Please use braces around if/then/else blocks, even if they are otherwise one single line.

src/freenet/client/filter/WAVFilter.java
```diff
@@ -0,0 +1,124 @@
...
+	@Override
+	protected void readFilterChunk(byte[] ID, int size, Object context, DataInputStream input, DataOutputStream output,
+			String charset, Map<String, String> otherParams, String schemeHostAndPort, FilterCallback cb)
+			throws DataFilterException, IOException {
+		WAVFilterContext ctx = (WAVFilterContext)context;
+		if(ID[0] == 'f' && ID[1] == 'm' && ID[2] == 't' && ID[3] == ' ') {
+			if(ctx.hasfmt) {
+				throw new DataFilterException(l10n("invalidTitle"), l10n("invalidTitle"), "Unexpected fmt chunk was encountered");
+			}
+			if(size != 16 && size != 18 && size != 40) {
```
bertm, line 47: What do these magic numbers represent? (please introduce constants with descriptive names)

src/freenet/client/filter/WAVFilter.java
```diff
@@ -0,0 +1,124 @@
...
+			// Further checks
+			if((ctx.format == WAVE_FORMAT_ALAW || ctx.format == WAVE_FORMAT_MULAW) && ctx.wBitsPerSample != 8) {
+				throw new DataFilterException(l10n("invalidTitle"), l10n("invalidTitle"), "Unexpected bits per sample value");
+			}
+			return;
+		} else if(!ctx.hasfmt) {
+			throw new DataFilterException(l10n("invalidTitle"), l10n("invalidTitle"), "Unexpected header chunk was encountered, instead of fmt chunk");
+		}
+		if(ID[0] == 'd' && ID[1] == 'a' && ID[2] == 't' && ID[3] == 'a') {
+			if(ctx.format == WAVE_FORMAT_PCM || ctx.format == WAVE_FORMAT_IEEE_FLOAT || ctx.format == WAVE_FORMAT_ALAW || ctx.format == WAVE_FORMAT_MULAW) {
```
bertm, line 86: This condition will always evaluate to true (ctx.format is only written in line 51 and checked immediately after)

bertm (changes requested):

ArneBab:
@Bombe @bertm can you check whether your comments are addressed?

src/freenet/client/filter/WAVFilter.java
```diff
@@ -0,0 +1,131 @@
...
+			}
+			ctx.hasdata = true;
+		} else if(ID[0] == 'f' && ID[1] == 'a' && ID[2] == 'c' && ID[3] == 't') {
+			if(size < 4) {
+				throw new DataFilterException(l10n("invalidTitle"), l10n("invalidTitle"), "fact chunk must contain at least 4 bytes");
+			}
+			// Just dwSampleLength (Number of samples) here, pass through
+			output.write(ID);
+			writeLittleEndianInt(output, size);
+			passthroughBytes(input, output, size);
```
bertm, line 110: The passed-through bytes are format-specific. This would allow passing arbitrary unverified non-audio data.
However, the currently supported formats are uncompressed and should, if at all, only have the first `dwSampleLength` field.
Can we limit the number of passed-through bytes to 4 and ignore whichever bytes remain? Of course this should be adjusted accordingly in the size following the `fmt ` chunk ID.

test/freenet/client/filter/wav/test.wav
bertm, line 1: This file appears to be `TADA.WAV` taken from Microsoft Windows 3.1 or 3.11.

I am aware that that's 1992 calling, but I could not find any records that Microsoft would have put these sounds in the public domain, so I need to assume this file is still copyrighted and owned by Microsoft Corporation.

Unless this file can be shown to be in the public domain, please replace it with a file that does not infringe any license or copyright.

bertm (changes requested):
@torusrxxx with a few minor adjustments, this should be good to go.

src/freenet/client/filter/WAVFilter.java
torusrxxx: limiting the number of passed-through bytes doesn't guarantee the remaining 00 bytes won't have an effect, so I just discard this entire trunk when the size is not 4.

test/freenet/client/filter/wav/test.wav
torusrxxx: I replaced the content with silence

ArneBab:
@bertm is this good to go?

bertm (approved):
The testing could be a bit heavier, but the basics are covered.

ArneBab:
Merged -- thank you!

PR #983, opened 2024-10-01
Add image/webp as prefetch allowed types

torusrxxx:
Forgot to add image/webp here.


ArneBab (approved):
Good catch!

ArneBab:
merged — thank you!

PR #984, opened 2024-10-04
Fix SSL self-signed certificate, update cipher suite

torusrxxx:
Old self-signed certificate generation doesn't always work with Oracle JVM, so I replaced it with the one which use bouncycastle.
Also the node does not try to create a self-signed certificate on start up, because the entropy collection has not yet started when initializing SSL. The self-signed certificate is created when the user turns on SSL.
Finally, SSL cipher suites with weak CBC methods are removed, and SSL cipher suites with AES 256 are added.
In the future, we should allow to import valid SSL certificate and private key.


bertm:
> Old self-signed certificate generation doesn't always work with Oracle JVM

When does it not work? Can we have a unit test for that particular case?

src/freenet/node/NodeStarter.java
```diff
@@ -148,6 +144,11 @@ public Integer start(String[] args) {
 			return Integer.valueOf(-2);
 		}
 
+		String builtWithMessage = "freenet.jar built with freenet-ext.jar Build #" + ExtVersion.buildNumber + " r" + ExtVersion.cvsRevision+" running with ext build "+extBuildNumber+" r" + extRevisionNumber;
```
bertm, line 147: This seems unrelated to the PR.
What's the reason for moving this around?

torusrxxx:
> > Old self-signed certificate generation doesn't always work with Oracle JVM
> 
> When does it not work? Can we have a unit test for that particular case?

It's because Oracle doesn't want to let it work in all J2SE run-times. See https://stackoverflow.com/questions/29060064/sun-security-x509-certandkeygen-and-sun-security-pkcs-pkcs10-missing-in-jdk8

src/freenet/node/NodeStarter.java
torusrxxx: It appeared before logging is started, and I don't find this message in the log file, so I moved it to after logging started.

ArneBab (approved):

ArneBab:
Looks good to me — merged! Thank you!

PR #985, opened 2024-10-05
Remove pooling and other special cases for SHA-256 digests

bertm:
During review of #831 it was observed that the overhead of pooling outweights that of constructing new digest instances.
This PR removes the pooling mechanism from the SHA256 helper class. Because the return-to-pool method is now a no-op, its usages are removed. The method is now deprecated, but remains for backwards compatibility.

While cleaning up I encountered the HashType enum, which had a special case for pooling of SHA-256 too. This deprecates the related return-to-pool method, and removes the special handling of SHA-256 message digest creation in this class.


ArneBab (approved):
That’s great! I love seeing the double-try blocks become unnecessary and be removed.

Thank you!

ArneBab:
merged — thank you!

PR #989, opened 2024-10-13
Cleanup RunningAverage implementations

bertm:
the RunningAverage implementations have various issues:
* Too many responsibilities (like TimeDecayingRunningAverage, that tries to keep track of clock skew in the middle of its calculations)
* Too much mutable state (like SimpleRunningAverage, that keeps track of "how many items do I have" across 3 different fields)
* A generous amount of synchronization, but not everywhere
* Leftover methods from earlier times (custom serialization formats and unused code that does who-knows-what)
* A large amount of empty Javadoc comment
* A general lack of consistency in code formatting
* Zero tests

This PR attempts to improve the situation by:
* Separating out the clock skew tracking & reporting out of the main logic of TimeDecayingRunningAverage (and using a monotonic clock for the actual computations)
* Keeping the state in a single immutable object accessed through an AtomicReference where possible, so state is in one place and can easily be reasoned about
* Replacing synchronization with compare-and-set operations on the reference
* Removing the legacy format read and write method, and removing Serializable from the class hierarchy (this is technically a breaking, but in my view it is highly unlikely that anybody would be using these methods)
* Removing the empty Javadoc. Most of the operations are explained on the RunningAverage interface; others are sufficiently clear.
* Reformat everything (consistent use of tabs vs spaces, yay!), move constructors to the top, etc.
* Add some tests to cover most of the functionality

Also, deprecate MedianMeanRunningAverage as it's probably never a good idea to use it.

### DIFF
--- a/src/freenet/node/RequestSender.java
+++ b/src/freenet/node/RequestSender.java
@@ -44,7 +44,8 @@ import freenet.support.ShortBuffer;
 import freenet.support.SimpleFieldSet;
 import freenet.support.TimeUtil;
 import freenet.support.io.NativeThread;
-import freenet.support.math.MedianMeanRunningAverage;
+import freenet.support.math.RunningAverage;
+import freenet.support.math.TrivialRunningAverage;
 
 /**
  * @author amphibian
@@ -1541,9 +1542,9 @@ public final class RequestSender extends BaseSender implements PrioRunnable {
     	}
     }
     
-	private static MedianMeanRunningAverage avgTimeTaken = new MedianMeanRunningAverage();
+	private static final RunningAverage avgTimeTaken = new TrivialRunningAverage();
 	
-	private static MedianMeanRunningAverage avgTimeTakenTransfer = new MedianMeanRunningAverage();
+	private static final RunningAverage avgTimeTakenTransfer = new TrivialRunningAverage();
 	
 	private long transferTime;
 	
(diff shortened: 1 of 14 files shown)


test/freenet/support/math/BootstrappingDecayingRunningAverageTest.java
```diff
@@ -0,0 +1,93 @@
...
+        average.report(0);
+        assertThat(average.countReports(), equalTo(1L));
+
+        // Invalid reports
+        average.report(-1000);
+        average.report(1000);
+        average.report(Double.NEGATIVE_INFINITY);
+        average.report(Double.POSITIVE_INFINITY);
+        average.report(Double.NaN);
+        assertThat(average.countReports(), equalTo(1L));
```
ArneBab, line 46: also check that the value is 0?

src/freenet/support/math/TrivialRunningAverage.java
```diff
@@ -1,77 +1,69 @@
...
+        private final long reports;
+        private final double total;
+
+        Data(long reports, double total) {
+            this.reports = reports;
+            this.total = total;
+        }
+
+        Data updated(double d) {
+            return new Data(reports + 1, d + total);
```
ArneBab, line 61: This seems very expensive. Do we really need it?

src/freenet/support/math/MedianMeanRunningAverage.java
```diff
@@ -1,98 +1,80 @@
...
-		int middle = size / 2;
-		java.util.Collections.sort(reports);
-		return reports.get(middle);
-	}
+    @Override
+    public double currentValue() {
+        synchronized (reports) {
+            int size = reports.size();
+            int middle = size / 2;
+            Collections.sort(reports);
```
ArneBab, line 48: Is the default sort optimal for already mostly sorted data?

Yes, it is:

> If the input array is nearly sorted, the implementation requires approximately n comparisons https://docs.oracle.com/javase/8/docs/api/java/util/List.html#sort-java.util.Comparator-

src/freenet/node/RequestSender.java
```diff
@@ -1542,9 +1543,9 @@ public synchronized short waitUntilStatusChange(short mask) {
     	}
     }
     
-	private static MedianMeanRunningAverage avgTimeTaken = new MedianMeanRunningAverage();
+	private static final RunningAverage avgTimeTaken = new TrivialRunningAverage();
 	
-	private static MedianMeanRunningAverage avgTimeTakenTransfer = new MedianMeanRunningAverage();
+	private static final RunningAverage avgTimeTakenTransfer = new TrivialRunningAverage();
```
ArneBab, line 1548: Why does this use the TrivialRunningAverage, when there are much more efficient algorithms? This looks like a good case for the now more efficient `BootstrappingDecayingRunningAverage`.

src/freenet/support/math/TrivialRunningAverage.java
bertm: This adds two numbers and constructs a tiny object. What exactly seems very expensive here? :)

> Do we really need the trivial running average?

It is used all over the place (46 locations) so we cannot just remove it.

src/freenet/node/RequestSender.java
bertm: MedianMeanRunningAverage is expensive (and is therefore deprecated) as it has O(1) updates but O(n log n) queries and O(n) memory usage.

TrivialRunningAverage is cheap with O(1) updates, O(1) queries and O(1) memory usage.

BootstrappingDecayingRunningAverage might give slightly more useful metrics, but note that `avgTimeTakenTransfer` is only used when `logMINOR` is enabled for this class - this is mostly a debugging tool that probably nobody actually uses.

test/freenet/support/math/BootstrappingDecayingRunningAverageTest.java
bertm: Fair enough, added an assertion for that!

bertm:
@ArneBab thank you for your review!

Addressed your comment regarding the additional assertion and rebased this branch on the latest `next`.
Hope I cleared up the `TrivialRunningAverage` discussions too - just let me know if we're on the same page now :smile:

ArneBab:
merged -- thank you!

PR #955, opened 2024-03-18
Replace Java 1.4 era 'for' loops

Juiceman:
Use Enhanced 'for' loops

PR #990, opened 2024-10-17
Support more HTML elements and CSS attributes and fix CSS filter crashes

torusrxxx:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir


bertm (changes requested):
This looks reasonable but please do add a test for this. The bare minimum would be something like / an extension to what is introduced in
8d12ea0dfa8d50c462c9e6584617cbfd13792d83

torusrxxx:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ruby
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rp
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rt

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1620,6 +1621,11 @@ else if("word-wrap".equalsIgnoreCase(element))
 			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("normal", "break-word", "anywhere"),ElementInfo.VISUALMEDIA));
 			allelementVerifiers.remove(element);
 		}
+		else if("writing-mode".equalsIgnoreCase(element))
+		{
+			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("horizontal-tb", "vertical-rl", "vertical-lr", "lr", "lr-tb", "rl", "tb", "tb-lr", "tb-rl"),ElementInfo.VISUALMEDIA));
+			allelementVerifiers.remove(element);
```
ArneBab, line 1627: The shorthands are deprecated except for SVG (see https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode ), so I think we should avoid them.

Supporting SVG as content type would be a huge project, because that can embed far too many ways to access remote content.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -206,8 +215,8 @@ public class ElementInfo {
 			"nth-last-child",
 			"nth-of-type",
 			"nth-last-of-type",
-			"link",
-			"visited",
+			"link", // inverse of visited, probably should be blocked?
```
ArneBab, line 218: no, should not be blocked: this applies to any link. As long as visited is stripped out, there’s no privacy risk.

⇒ please adjust the comment.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
...
-			"target"
+			"target",
+			"any-link",
+			"default",
+			"defined",
+			"disabled",
+			"empty",
+			"enabled",
+			"focus-visible",
+			"dir",
```
ArneBab, line 238: is dir necessary here? We already have a dedicated parsing rule for it. If it is: please add a test that fails without your change.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
 			"first-letter",
 			"before",
 			"after",
-			"target"
+			"target",
+			"any-link",
```
ArneBab, line 231: this is the solution to the link/visited privacy problems.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
 			"first-letter",
 			"before",
 			"after",
-			"target"
+			"target",
```
ArneBab, line 230: target looks dangerous (it checks which link is used to access a page). It can be used to trigger background fetches when navigating (just verified: the network request is sent when the target is clicked), so it may be necessary to restrict its content to non-links to avoid timing attacks.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
...
+			"any-link",
+			"default",
+			"defined",
```
ArneBab, line 233: we do not have custom element support (it requires javascript), so defined does not make sense. Please remove it.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
...
+			"disabled",
+			"empty",
+			"enabled",
+			"focus-visible",
+			"dir",
+			"indeterminate",
```
ArneBab, line 239: also only used for forms.

src/freenet/client/filter/FilterUtils.java
```diff
@@ -370,70 +372,57 @@ public static boolean isValidCSSShape(String value)
 	public static boolean isMedia(String media) {
 		return cssMedia.contains(media);
 	}
+	
+	public static final Pattern hexColorPattern = Pattern.compile("#(?>[0-9a-f]{8}|[0-9a-f]{6}|[0-9a-f]{3,4})", Pattern.CASE_INSENSITIVE);
+	
 	public static boolean isColor(String value)
 	{
-		value=value.trim();
+		value=value.trim().toLowerCase();
```
ArneBab, line 380: good catch!

test/freenet/client/filter/FilterUtilsTest.java
```diff
@@ -44,4 +44,26 @@ public void testInvalidLengthUnits() {
...
+		assertTrue(FilterUtils.isColor("rgba(100 200 255 / 0.25)"));
+		assertTrue(FilterUtils.isColor("rgba(010 00200 255 / 25%)"));
+		assertTrue(FilterUtils.isColor("rgba(none 0 0% /)"));
+	}
+	
+	@Test
+	public void testInvalidColors() {
+		assertFalse(FilterUtils.isColor("rgb(0.1 0.2 0.3)"));
+		assertFalse(FilterUtils.isColor("rgb(/)"));
+		assertFalse(FilterUtils.isColor("#ABCDEFGH"));
```
ArneBab, line 67: can you add more color-tests? switching to regexes usually needs tests for edge-cases.

src/freenet/client/filter/HTMLFilter.java
```diff
@@ -953,49 +953,57 @@ private static Map<String, TagVerifier> getAllowedTagVerifiers()
...
+				"strong",
+				"sub",
 				"summary",
-				"details"};
+				"sup",
+				"tt",
+				"u",
+				"var",
+				"wbr",
+				"xmp"};
```
ArneBab, line 1006: which elements did you add? the reordering makes that hard to figure out.

ArneBab:
This is a neat addition, but the added elements still need tests. I know that these can be annoying to write, since they feel obvious, but the last two times I omitted them and Bombe reminded me with an ENOTEST, one of those would have actually shown a bug (that luckily got found before merging when I added the test Bombe had requested).

⇒ please do test. These filters are critical for safe operation in browsers that may query into the clearnet on errors here.

src/freenet/client/filter/ElementInfo.java
```diff
@@ -218,7 +227,27 @@ public class ElementInfo {
 			"first-letter",
 			"before",
 			"after",
-			"target"
+			"target",
+			"any-link",
+			"default",
+			"defined",
+			"disabled",
```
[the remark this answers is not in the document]
torusrxxx: Why do we need to discuss where these can be used? As long as it's valid it should be here, right? And if it has problems, we can add it into the BANNED_PSEUDOCLASS. Freesites do have forms, could be Library, commentmaker or some other future plugin.

src/freenet/client/filter/ElementInfo.java
torusrxxx: Remove it or add it into the BANNED_PSEUDOCLASS?

src/freenet/client/filter/CSSTokenizerFilter.java
torusrxxx: I think these are valid and not dangerous so should be here. Freenet also has lots of other deprecated things, but still allowed because it's safe.

src/freenet/client/filter/ElementInfo.java
torusrxxx: target is usually used to highlight a section heading. Clicking different links should already trigger different network effects because they can link to different pages, yeah.

src/freenet/client/filter/HTMLFilter.java
torusrxxx: "bdi","figcaption","figure","main","mark","rp","rt","ruby". I reordered them because it's hard to find which element is still missing. All these new elements are added to tests HTML5_TAGS and HTML5_BDI_RUBY.

src/freenet/client/filter/ElementInfo.java
ArneBab: BANNED_PSEUDOCLASS is better: CSSTokenizerFilter::HTMLElementVerifier defines that as `banned (but otherwise valid) selector`.

src/freenet/client/filter/ElementInfo.java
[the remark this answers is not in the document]
ArneBab: I’m being careful with adding selectors that might not be usable, because I have to check with each element, whether it could allow sneaking in an unfiltered URL. That’s why the review of these takes a while (need to look at mdn for each).

Also we have to need tests for them all.

That said: better CSS support is great and I love it that you’re pushing the state here!

src/freenet/client/filter/HTMLFilter.java
ArneBab: OK — thank you! Supporting ruby looks like a big improvement for Asian languages.

src/freenet/client/filter/ElementInfo.java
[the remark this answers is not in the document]
ArneBab: … so yes, disabled should be useful, even when we can’t use all of the capabilities of forms.

src/freenet/client/filter/ElementInfo.java
ArneBab: You’re right — if they get the link from outside, it could just be different links, and if they click it on the page, it could lead anywhere in the net.

⇒ good to go. Thank you!

src/freenet/client/filter/ElementInfo.java
ArneBab: A future improvement to keep pages working could be to transparently rewrite "link" to "any-link". But that may be beyond the scope of this PR (no need to fix everything in one step).

src/freenet/client/filter/CSSTokenizerFilter.java
ArneBab: OK.

torusrxxx:
Some more tests have been added. However, tests can only be written for expected inputs, but problems occur for unexpected inputs which cannot be tested before you find a problem! I already found two bugs that can crash the CSS validator within just a few days - very scary indeed!

ArneBab:
Please note when you finished adding elements so I can review it again.

torusrxxx:
Ok, I will freeze this one from now on. I do have further plans, but it seems getting code reviewed in this project is not as quickly as my first contribution being released in a new version in just a day. Hope you will have more free time in the future!

torusrxxx:
Checking MDN for updated elements and add them here usually takes just a small amount of time that could otherwise watch a video. But I am gonna watch videos as so many things are waiting to be merged.

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -39,7 +39,7 @@ public class CSSParserTest {
 	private final static HashMap<String,String> CSS1_SELECTOR= new HashMap<>();
 	static {
 		CSS1_SELECTOR.put("h1 {}","h1");
-		CSS1_SELECTOR.put("h1:link {}","h1:link");
+		CSS1_SELECTOR.put("h1:link {color:transparent}","");
```
ArneBab, line 42: Does this also filter out h1:link without content?

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -137,6 +137,10 @@ public class CSSParserTest {
 		CSS2_BAD_SELECTOR.add("h1[foo,=bar] {}");
 
 		CSS2_BAD_SELECTOR.add("h1:langblahblah(fr) {}");
+		// java.lang.StringIndexOutOfBoundsException
+		CSS2_BAD_SELECTOR.add("h1:golang {}");
```
ArneBab, line 141: good catch!

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -4871,18 +4870,15 @@ public static void main(String arg[]) throws Throwable {
...
 		final Bucket outputBucket = new FileBucket(fout, false, true, false, false);
-		InputStream inputStream = null;
-		OutputStream outputStream = null;
-		try {
-			inputStream = inputBucket.getInputStream();
-			outputStream = outputBucket.getOutputStream();
+		try (
+			InputStream inputStream = inputBucket.getInputStream();
+			OutputStream outputStream = outputBucket.getOutputStream()
+		) {
```
ArneBab, line 4946: nice!

src/freenet/client/filter/ElementInfo.java
```diff
@@ -427,11 +427,16 @@ public static boolean isBannedPseudoClass(String cname)
 				// Pseudo-classes can be chained, at least dynamic ones can, see CSS2.1 section 5.11.3
 				String[] split = cname.split(":");
 				for(String s : split)
-					if(isBannedPseudoClass(s)) return true;
+					if(isBannedPseudoClass2(s)) return true;
```
ArneBab, line 430: why the switch to isBannedPseudoClass2 ?

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1612,12 +1613,9 @@ else if("white-space".equalsIgnoreCase(element))
 			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("normal","pre","nowrap","pre-wrap","pre-line"),ElementInfo.VISUALMEDIA));
 			allelementVerifiers.remove(element);
 		}
-		else if("white-space-collapsing".equalsIgnoreCase(element))
+		else if("white-space-collapse".equalsIgnoreCase(element))
```
ArneBab, line 1657: good catch!

test/freenet/client/filter/CSSParserTest.java
```diff
@@ -985,6 +986,8 @@ public class CSSParserTest {
 		propertyTests.put("#x { text-shadow: white 2px 5px; }", "#x { text-shadow: white 2px 5px; }");
 		propertyTests.put("#x { text-shadow: 5px 10px; }", "#x { text-shadow: 5px 10px; }");
 		propertyTests.put("#x { text-shadow: 1px 1px 2px 1px black; }", "#x { }");
+		// not possible to parse a comma separated list?
+		//propertyTests.put("#x { text-shadow: 1px 1px 2px red, 0 0 1em blue, 0 0 0.2em blue; }", "#x { text-shadow: 1px 1px 2px red, 0 0 1em blue, 0 0 0.2em blue; }");
```
ArneBab, line 1005: might just not be implemented yet.

ArneBab:
I now reviewed all the commits since my last review ( https://github.com/hyphanet/fred/pull/990#issuecomment-2425690790 ). Once you address the comments, I’ll gladly merge to next.

(no need to address "good catch" ☺)

Thank you!

src/freenet/client/filter/ElementInfo.java
torusrxxx: isBannedPseudoClass2 is a newly added function, in order to avoid `String[] split = cname.split(":");` again for just a single class and improve performance.

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -375,7 +375,7 @@ public static <T> T[] concat(T[] a, T[] b) {
 		auxilaryVerifiers[78]=new CSSPropertyVerifier(null,null,Arrays.asList("70<1,2>"),null,true);
 
 		// <text-shadow>
-		auxilaryVerifiers[79]=new CSSPropertyVerifier(null, null, Arrays.asList("74a73"), null, true);
+		auxilaryVerifiers[79]=new CSSPropertyVerifier(null, null, Arrays.asList("73 72 72 72","73 72 72","72 72 72 73","72 72 73","72 72"), null, true);
```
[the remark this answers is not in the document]
torusrxxx: Yeah, it doesn't always work, either fails to recognize some valid forms, or throws an exception. In the end I switched to this, which passes all tests.

src/freenet/client/filter/CSSTokenizerFilter.java
torusrxxx: There was already #830 (and more broadly #954 #955 and so on). I think it would be nice if we can get these into next as well.

test/freenet/client/filter/CSSParserTest.java
torusrxxx: Yeah, it's still not supported yet, but I don't like texts glowing like rainbow anyway, will leave it to future.

src/freenet/client/filter/ElementInfo.java
torusrxxx: dir(ltr) and dir(rtl) as pseudoclasses, added here.

src/freenet/client/filter/ElementInfo.java
ArneBab: Nice! Thank you!

src/freenet/client/filter/CSSTokenizerFilter.java
ArneBab: #830 needs a review and #955 needs changes from review, but #954 is ready to be merged — will merge it.

ArneBab:
Merged — thank you!

PR #993, opened 2024-11-12
Reduce pings and dns lookups

ArneBab:
Only sort the list nominalPeer when necessary.

That sorting can cause Pings to be sent. Now sorting is only done when
receiving a new noderef and at random intervals.

This should make the node less visible in DNS requests and Pings (more
typical pattern instead of requesting all peers at the same time).

Also it fixes the startup time regression by not doing a ping in the
PeerNode constructor.

The sorting result in the constructor was overwritten anyway …

### DIFF
--- a/src/freenet/node/DNSRequester.java
+++ b/src/freenet/node/DNSRequester.java
@@ -102,9 +102,11 @@ public class DNSRequester implements Runnable {
         // Try new DNS lookup
         //Logger.minor(this, "Doing lookup on "+pn+" of "+nodesToCheck.length);
         pn.maybeUpdateHandshakeIPs(false);
+
+        int nextMaxWaitTime = 1000 + node.getFastWeakRandom().nextInt(60000);
         try {
             synchronized(this) {
-                wait(1000 + node.getFastWeakRandom().nextInt(60000));  // sleep 1-61s ...
+                wait(nextMaxWaitTime);  // sleep 1-61s ...
             }
         } catch (InterruptedException e) {
             // Ignore, just wake up. Just sleeping to not busy wait anyway
--- a/src/freenet/node/DNSRequester.java
+++ b/src/freenet/node/DNSRequester.java
@@ -3,6 +3,12 @@
  * http://www.gnu.org/ for further details of the GPL. */
 package freenet.node;
 
+import java.util.ArrayDeque;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Set;
+
 import freenet.support.LogThresholdCallback;
 import freenet.support.Logger;
 import freenet.support.Logger.LogLevel;
@@ -16,6 +22,8 @@ public class DNSRequester implements Runnable {
 
     final Node node;
     private long lastLogTime;
+    private final Set<Double> recentNodeIdentitySet = new HashSet<>();
+    private final Deque<Double> recentNodeIdentityQueue = new ArrayDeque<>();
     // Only set when doing simulations.
     static boolean DISABLE = false;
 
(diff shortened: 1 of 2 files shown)


torusrxxx:
Next thing I like to have is custom DNS server & DNS over HTTPS via dnsjava, to secure the DNS requests.

src/freenet/node/DNSRequester.java
```diff
@@ -56,23 +56,24 @@ public void run() {
...
-        for(PeerNode pn: nodes) {
-            //Logger.minor(this, "Node: "+pn);
-            if(!pn.isConnected()) {
-                // Not connected
-                // Try new DNS lookup
-            	//Logger.minor(this, "Doing lookup on "+pn+" of "+nodes.length);
-                pn.maybeUpdateHandshakeIPs(false);
-            }
+        // check a randomly chosen node to avoid sending bursts of DNS requests
+        PeerNode pn = nodes[node.getFastWeakRandom().nextInt(nodes.length)];
```
bertm, line 66: This approach turns the peer node DNS resolving into the [Coupon collector's problem](https://en.wikipedia.org/wiki/Coupon_collector%27s_problem). For 80 peer nodes, the expected number of tries before all are hit is 397 (taking roughly 8 minutes), and it takes 563 attempts (~10 minutes) before 95% probability is reached - even assuming the peer node list is static, which it is not. This may not be desirable.

This can be resolved by ignoring connected nodes prior to random selection.

bertm (changes requested):

src/freenet/node/DNSRequester.java
ArneBab: That’s a great catch — thank you!

Does only checking non-connected nodes really resolve it? I’d think it would only reduce the problem.

ArneBab:
> Next thing I like to have is custom DNS server & DNS over HTTPS via dnsjava, to secure the DNS requests.

Wouldn’t DNS over HTTPS create a stronger way to track nodes? DNS requests via UDP are heavily cached, so watchers can never trust to see DNS requests. Also our connections are secured via signatures, so a false IP address cannot cause a MitM attack.

Custom DNS servers would be neat, though. A great synergy could be DNS over GNS (gnunet name system).

src/freenet/node/DNSRequester.java
ArneBab: @bertm the new commit should fix this: it uses a set and a deque to ensure that we only check a node again if at least 80% of the other unconnected nodes have been checked (as cheapskate’s ordering trick ☺).

It uses the location to check for equality, because these must not be equal, and because a set and a deque of doubles is cheap.

ArneBab: See https://github.com/hyphanet/fred/pull/993/commits/ced3b161f8480ff570927be43450cd38e3443617

ArneBab:
This PR is a blocker for 1499 — the only hard blocker — so it would be great if we could get it done soon.

src/freenet/node/PeerNode.java
```diff
@@ -788,9 +779,17 @@ private boolean parseARK(SimpleFieldSet fs, boolean onStartup, boolean forDiffNo
 	*/
 	@Override
 	public synchronized Peer getPeer() {
+		if (detectedPeer == null && !nominalPeer.isEmpty()) {
+			sortNominalPeer();
+			detectedPeer = nominalPeer.get(0);
```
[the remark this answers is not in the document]
ArneBab: fixed in https://github.com/hyphanet/fred/pull/993/commits/1cf9ed50656ab819e324299f077ab8f6a44a49d3

src/freenet/node/DNSRequester.java
```diff
@@ -56,23 +56,24 @@ public void run() {
     private void realRun() {
         PeerNode[] nodes = node.getPeers().myPeers();
         long now = System.currentTimeMillis();
-        if((now - lastLogTime) > 1000) {
-        	if(logMINOR)
+        if((now - lastLogTime) > 100) {
```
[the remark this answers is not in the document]
ArneBab: done, thanks!

torusrxxx:
> > Next thing I like to have is custom DNS server & DNS over HTTPS via dnsjava, to secure the DNS requests.
> 
> Wouldn’t DNS over HTTPS create a stronger way to track nodes? DNS requests via UDP are heavily cached, so watchers can never trust to see DNS requests. Also our connections are secured via signatures, so a false IP address cannot cause a MitM attack.
> 
> Custom DNS servers would be neat, though. A great synergy could be DNS over GNS (gnunet name system).

Watchers are not just at the core router. They can be just at the WiFi AP you use (with unknown firmware controlled by ISP with possible preinstalled spyware). For the mass surveillance by ISP, it doesn't matter whether DNS is cached, or which DNS server is used, but it does matter whether DNS is encrypted. Anyone requesting the domain name of a Freenet node is likely using Freenet. DNS over HTTPS can also use a HTTP proxy server. The only thing I'm concerned about is that `DohResolver` on https://github.com/dnsjava/dnsjava is just Proof-of-concept.

bertm:
> Anyone requesting the domain name of a Freenet node is likely using Freenet.

Anyone sending UDP packets to a Freenet node is likely using Freenet too.

Opennet nodes can be enumerated with little effort, that cannot be changed due to the open nature of the network. If you're worried about DNS being monitored specifically, better solve it at the OS level. If you're worried about others knowing you participate in Freenet, use Darknet.

ArneBab:
> If you're worried about others knowing you participate in Freenet, use Darknet.

Darknet is why this PR reduces DNS requests: if a node is not known to be a Freenet node, sending large numbers of DNS requests in short, regular intervals could be used as pattern to find it.

I do not think that this is a problem at the moment, because I think few people use dyndns hostnames in their node config, but if many people used them — for example to get connectivity for the first connection (where they cannot get the new IP via an ark request), then that would become a problem.

This also points to the regular "1 second wait" as a problem. I’ll adjust that to "1 - 60s" now that we’re selecting the nodes to check efficiently.

src/freenet/node/DNSRequester.java
```diff
@@ -54,25 +62,51 @@ public void run() {
...
-            //Logger.minor(this, "Node: "+pn);
-            if(!pn.isConnected()) {
-                // Not connected
-                // Try new DNS lookup
-            	//Logger.minor(this, "Doing lookup on "+pn+" of "+nodes.length);
-                pn.maybeUpdateHandshakeIPs(false);
+        // check a randomly chosen node that has not been checked
+        // recently to avoid sending bursts of DNS requests
+        int unconnectedNodesLength = nodesToCheck.length;
+        PeerNode pn = nodesToCheck[node.getFastWeakRandom().nextInt(unconnectedNodesLength)];
```
bertm, line 86: This will throw `ArrayIndexOutOfBoundsException` if `unconnectedNodesLength` is zero. Can be solved by an early return.

src/freenet/node/DNSRequester.java
```diff
@@ -54,25 +62,51 @@ public void run() {
...
+            recentNodeIdentitySet.add(pn.getLocation());
+            recentNodeIdentityQueue.offerFirst(pn.getLocation());
+            while (recentNodeIdentityQueue.size() > (0.81 * unconnectedNodesLength)) {
+                recentNodeIdentitySet.remove(recentNodeIdentityQueue.removeLast());
             }
         }
+        //Logger.minor(this, "Node: "+pn);
+
+        // Try new DNS lookup
+        //Logger.minor(this, "Doing lookup on "+pn+" of "+nodesToCheck.length);
```
bertm, line 103: Can we get rid of the commented out debug code? :smile:

src/freenet/node/DNSRequester.java
```diff
@@ -54,25 +62,51 @@ public void run() {
...
+        if (unconnectedNodesLength < 5) {
+            // no need for optimizations: just clear all state
+            recentNodeIdentitySet.clear();
+            recentNodeIdentityQueue.clear();
+        } else {
+            // do not request this node again,
+            // until at least 80% of the other unconnected nodes have been checked
+            recentNodeIdentitySet.add(pn.getLocation());
+            recentNodeIdentityQueue.offerFirst(pn.getLocation());
+            while (recentNodeIdentityQueue.size() > (0.81 * unconnectedNodesLength)) {
```
bertm, line 96: The comment says 80% but the implementation says 81%. I don't have a strong opinion on the actual percentage used, but please do align the two.

bertm (changes requested):

src/freenet/node/DNSRequester.java
ArneBab: early return sounds like the best fix — implemented. Thanks!

src/freenet/node/DNSRequester.java
ArneBab: Deleted ☺

src/freenet/node/DNSRequester.java
ArneBab: I aligned the comment with the code — 0.81 is just to ensure that 5 nodes with 4 already checked is not an edge case (exactly matching the boundary).

bertm (approved):

ArneBab:
@bertm thank you for your review! I merged this and so 1499 is no longer blocked, but I did not expect the flurry of great PRs coming in today and my brain is already sloshed, so I’ll defer preparing 1499 to tomorrow so we all have some more time to review (and hopefully merge) PRs. :heart:

PR #994, opened 2024-11-12
refactor: split parsing peers into multiple steps

ArneBab:
The previous code nested too much error handling even though many parts are independent.

This is a cleanup as sideeffect of trying to multi-thread the peer initialization. That multi-threading turned out to be unnecessary, but the cleanup is nice without it.

### DIFF
--- a/src/freenet/node/PeerManager.java
+++ b/src/freenet/node/PeerManager.java
@@ -209,7 +209,7 @@ public class PeerManager {
 			if(peersFile.exists())
 				if(readPeers(peersFile, crypto, opennet, oldOpennetPeers)) {
 					String msg;
-					if(oldOpennetPeers) {
+					if (oldOpennetPeers) {
 						msg = "Read " + opennet.countOldOpennetPeers() + " old-opennet-peers from " + peersFile;
 					} else if(isOpennet) {
 						msg = "Read " + getOpennetPeers().length + " opennet peers from " + peersFile;
@@ -221,7 +221,7 @@ public class PeerManager {
 					return;
 				}
 		}
-		if(!isOpennet) {
+		if (!isOpennet) {
 			System.out.println("No darknet peers file found.");
 		}
 		// The other cases are less important.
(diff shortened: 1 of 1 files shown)


src/freenet/node/PeerManager.java
```diff
@@ -238,58 +238,60 @@ private boolean readPeers(File peersFile, NodeCrypto crypto, OpennetManager open
...
-				        droppedOldPeers.add(e, fs.get("myName"));
-				    }
-                    someBroken = true;
-                    continue;
-                }
+		List<SimpleFieldSet> peerEntries = new ArrayList<>();
+		// read the peers file
+		try {
+			while (true) {
+			  peerEntries.add(new SimpleFieldSet(br, false, true));
```
Bombe, line 245: You’re mixing different kinds of whitespace here.

src/freenet/node/PeerManager.java
```diff
@@ -238,58 +238,60 @@ private boolean readPeers(File peersFile, NodeCrypto crypto, OpennetManager open
...
+					// Ignore.
+					Logger.error(this, "Dropping too-old opennet peer");
+				} else {
+					// A lot more noisy
+					droppedOldPeers.add(e, fs.get("myName"));
+				}
+			}
+		}
+
+		for (PeerNode pn : createdNodes) {
```
Bombe, line 286: Even if this whole block is only copied, please add braces for all parts of `if` statements. 🙂

Bombe (changes requested):

src/freenet/node/PeerManager.java
ArneBab: Braces and spaces added (and some tabify) ☺

src/freenet/node/PeerManager.java
ArneBab: _resisting the urge to do a full file whitespace cleanup that would cause merging issues_

ArneBab:
ready to be merged, as discussed with Bombe in FLIP.

PR #995, opened 2024-11-12
Probabilistic queue priorization: do not starve bulk requests on very fast connections

ArneBab:


test/freenet/node/NewPacketFormatTest.java
```diff
@@ -33,7 +33,7 @@ public void setUp() {
 	@Test
 	public void testEmptyCreation() throws BlockedTooLongException {
 		NewPacketFormat npf = new NewPacketFormat(null, 0, 0);
-		PeerMessageQueue pmq = new PeerMessageQueue();
+		PeerMessageQueue pmq = new PeerMessageQueue(new MersenneTwister(new byte[16]));
```
bertm, line 36: Apparently we don't care about the randomness in the tests, otherwise the tests would have failed.
If we don't care about it anyways, how about using `new DummyRandomSource(0)` rather than the impressive looking (but otherwise irrelevant) MersenneTwister?

bertm (changes requested):
If the choice of 90% is backed by data or measurements, please add that justification the Javadoc.

Otherwise fine, just a few code style issues.

test/freenet/node/NewPacketFormatTest.java
torusrxxx: Yeah. And for extremely fast nodes, even this may slow down performance, could be replaced by an atomic counter 1-10.

src/freenet/node/PeerMessageQueue.java
```diff
@@ -13,6 +14,7 @@
 import freenet.support.Logger;
 import freenet.support.Logger.LogLevel;
 import freenet.support.MutableBoolean;
+import freenet.node.Node;
```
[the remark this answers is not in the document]
ArneBab: fixed — thanks!

src/freenet/node/PeerMessageQueue.java
```diff
@@ -853,6 +857,7 @@ public synchronized boolean mustSendSize(int minSize, int maxSize) {
 		return false;
 	}
 
+	// TODO only try realtime first 90% of the time to ensure that realtime cannot starve bulk.
```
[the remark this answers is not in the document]
ArneBab: yes — removed.

test/freenet/node/NewPacketFormatTest.java
ArneBab: @torusrxxx I’m careful about deterministic values, because if different nodes operate in lockstep, that could cause unexpected behavior. It should not be problematic: according to https://github.com/js42721/fastrandom Mersenne Twister creates 1 billion random numbers in less than 10s, so this should not be a bottleneck for our use-case. When routing only SSKs on a 10gigabit/s  connection (1.25 gigabyte per second), we’ll have roughly 1 million packages per second, so we have about factor 100 buffer until Mersenne Twister could become a bottleneck.

@bertm switched to DummyRandomSource(1234) — 1234, because that value is often used in other places.

src/freenet/node/PeerMessageQueue.java
```diff
@@ -876,22 +876,20 @@ public synchronized MessageItem grabQueuedMessageItem(int minPriority) {
...
 			tryRealtimeFirst = true;
-		} else if(queuesByPriority[DMT.PRIORITY_BULK_DATA].getNextUrgentTime(Long.MAX_VALUE, 0) >= queuesByPriority[DMT.PRIORITY_REALTIME_DATA].getNextUrgentTime(Long.MAX_VALUE, 0)) {
+		} else if(queuesByPriority[DMT.PRIORITY_BULK_DATA].getNextUrgentTime(Long.MAX_VALUE, 0)
+				>= queuesByPriority[DMT.PRIORITY_REALTIME_DATA].getNextUrgentTime(Long.MAX_VALUE, 0)) {
 			tryRealtimeFirst = true;
 		} else {
-			tryRealtimeFirst = false;
+			// 10% chance to use bulk in case of a draw to avoid starving the bulk queue.
+			tryRealtimeFirst = this.fastWeakRandom.nextInt(10) > 0;
 		}
```
ArneBab, line 893: This resolves a dumb mistake: the exhaustive if-else block always replaced the random value.

ArneBab:
@bertm there isn’t an empiric base for the 10% — it’s just some small number to break out of possible starving. It may be sufficient to go to 1%.

bertm (approved):

ArneBab:
merged — thank you for the reviews!

PR #997, opened 2024-11-12
Remove references to & checks for long-gone "evil JVM bugs"

bertm:
The JVMs referred to can no longer run fred as they never got to 1.8 compatibility.
It remains unclear what the `EvilJVMBug` is exactly, but we can at least be sure that whatever is checked here can no longer lead to any positive detection.

Remove the code and translations associated with JVM checking altogether.


ArneBab (approved):
Looks good! Thank you!

ArneBab:
Merged — thank you again!

PR #999, opened 2024-11-17
update README badge for build status

qupo1:
Show GitHub Actions status using shields.io


bertm (approved):

ArneBab:
merged — thank you!

Using shields.io uses a remote service, but it’s no regression compared to travis, it is a free software service (could selfhost if needed), and it makes us a bit less dependent on github.

PR #1001, opened 2024-11-17
update URLs in README

qupo1:


bertm (changes requested):
The `current` version of Gradle is a moving target, and the link may break when the documentation structure changes. Please link to the actual version that is used (currently 8.9, after #996 it will be 8.11).

ArneBab (approved):
looks good to me — thank you!

Merging before #996 because while that’s still in review, the URLs are valid and #996 just waits for validation that the fix is OK and the link changes in the readme are useful right away, not only at release.

PR #1002, opened 2024-11-17
Remove code related to NLM load sending

bertm:
This code was effectively inactive since 2013 (which is when the FIXME in PeerNode.makeLoadStats was added: "re-enable when try NLM again").

Removing the related code not only gets rid of a few hundred lines of code, but also slightly reduces the (huge) number of things PeerNode needs to keep track of and simplifies the NPF logic substantially.

Note: this does not remove receiving of load stats from peer nodes.

### DIFF
--- a/src/freenet/node/NewPacketFormat.java
+++ b/src/freenet/node/NewPacketFormat.java
@@ -3,6 +3,8 @@
  * http://www.gnu.org/ for further details of the GPL. */
 package freenet.node;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
+
 import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -26,8 +28,6 @@ import freenet.support.LogThresholdCallback;
 import freenet.support.Logger;
 import freenet.support.Logger.LogLevel;
 import freenet.support.SparseBitmap;
-
-import static java.util.concurrent.TimeUnit.MINUTES;
 
 public class NewPacketFormat implements PacketFormat {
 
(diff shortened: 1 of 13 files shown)


src/freenet/node/NewPacketFormat.java
```diff
@@ -590,71 +590,28 @@ NPFPacket createPacket(int maxPacketSize, PeerMessageQueue messageQueue, Session
...
-									if((haveAddedStatsRT == null) && wrapper.getItem().sendLoadRT) {
-										addStatsRT = true;
-										// Add the lossy message outside the lock.
-										break addOldLoop;
-									}
-								}
-							}
+			synchronized(sendBufferLock) {
+				// Always finish what we have started before considering sending more packets.
+				// Anything beyond this is beyond the scope of NPF and is PeerMessageQueue's job.
```
ArneBab, line 599: This may cause changes in packet sending, so it could be critical. Do you see any code path where this could change the effective ordering/prioritization of messages (in a bad way)?

ArneBab:
Thank you for the many explanatory comments in message handling parts (including the concurrency warnings). These are critical and comments there should easy maintenance a lot!

src/freenet/node/NewPacketFormat.java
bertm: It's a lot of code touched, I see that. Not sure if that can be split up in smaller chunks.

There should not be any effective change, as all of these code paths are effectively unused. In this file specifically, the code paths for adding lossy messages are removed (those were used exclusively for load stats). Unfortunately, that code was quite convoluted.

I effectively started by removing this method in PeerNode:
```java
	@Override
	public MessageItem makeLoadStats(boolean realtime, boolean boostPriority, boolean noRemember) {
	    // FIXME re-enable when try NLM again.
	    return null;
```
And then pruned my way down all code that became effectively unreachable or effectively no-op.
* assume all invocations of `pn.makeLoadStats(...)` return `null` (which is clearly true as per the above FIXME)
* consider all instances of `sendStatsBulk`, `sendStatsRT` and their variations to be effectively `false`, as there is nothing to send
* consider all variations that could register the need to send stats (like `setSendLoadStatsASAP(...)`) no-ops
* every check that solely guards the above can now be removed

Then the changes probably make more sense :smile:

bertm:
I'm not aware of adding any explanatory comments in the message handling parts; those are probably attributed to me due to indentation changes (a few outer `while(true)` loops were removed).

bertm:
Rebased on `next` to resolve merge conflicts in NewPacketFormatTest.

ArneBab:
I now reviewed directly in Emacs instead of on github and there (with ignore whitespace) this was clear to review. Merged — thank you!

PR #1003, opened 2024-11-17
Fix copy-paste errors in backoff

bertm:
From what I can understand from the logic, this would cause transfer backoff to not be taken into account for bulk.

* The realtime parameter of getRoutingBackedOffUntil was set to `true` for both realtime and bulk.
* The bulk backoff parameter in getPeerNodeStatus was not used due to it having the wrong name (the equally named field was used instead).

Fixes db3329d567679c562f280aad53b80713aa19d73a

### DIFF
--- a/src/freenet/node/PeerNode.java
+++ b/src/freenet/node/PeerNode.java
@@ -14,7 +14,6 @@ import java.security.interfaces.ECPublicKey;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.EnumMap;
 import java.util.GregorianCalendar;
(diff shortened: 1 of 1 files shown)


ArneBab:
That’s a frighteningly good catch! Seeing your fix I wondered how this could have worked at all. The answer seems to be that if routingBackedOffUntilBulk was not set, it fell back on transferBackedOffUntilBulk:
```
        public synchronized long getRoutingBackedOffUntilBulk() {
		return Math.max(routingBackedOffUntilBulk, transferBackedOffUntilBulk);
	}
```

ArneBab (approved):

ArneBab:
merged — thank you!

PR #1007, opened 2024-11-19
Support IPV6_ADDR_PREFERENCES until Java 24 and beyond

bertm:
While reflection on Java's internals is slowly being crippled, Unsafe access is still possible (at least until JEP 417 [0] reaches phase 3 in Java 26 or later).

This requires moving from DatagramSocket to DatagramChannel as the former is substantially rewritten in Java 15 (JEP 373, see [1]) as a wrapper around the latter. The channel-based socket cannot be used in Java versions prior to 14 due to Java bug JDK-8232673 [2] which prevents read/write concurrency, effectively stalling the node.

Tested on OpenJDK 8, 11, 17, 21, 23, and 24.
Fixes bug 07217.

[0] https://openjdk.org/jeps/471
[1] https://openjdk.org/jeps/373
[2] https://bugs.openjdk.org/browse/JDK-8232673

### DIFF
--- a/src/freenet/io/comm/UdpSocketHandler.java
+++ b/src/freenet/io/comm/UdpSocketHandler.java
@@ -1,16 +1,14 @@
 package freenet.io.comm;
 
-import java.io.FileDescriptor;
 import java.io.IOException;
 import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.net.DatagramPacket;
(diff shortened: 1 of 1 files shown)


ArneBab (approved):
Looks good to me

ArneBab:
Merged — thank you!

PR #1008, opened 2024-11-21
Fix unhandled UnsupportedOperationException in Fallocate

torusrxxx:


src/freenet/support/io/Fallocate.java
```diff
@@ -19,121 +19,138 @@
...
+				errno = result == 0 ? 0 : Native.getLastError();
+			} else if (IS_POSIX) {
+				errno = FallocateHolderPOSIX.posix_fallocate(fd, offset, final_filesize-offset);
+			} else {
+				isUnsupported = true;
+			}
+		} else {
+			if (Platform.isWindows()) {
+				// On Windows, just write a byte at the end.
+				channel.write(ByteBuffer.allocate(1), final_filesize - 1);
```
ArneBab, line 93: why is this safe?

src/freenet/support/io/Fallocate.java
```diff
@@ -19,121 +19,138 @@
...
+		this.fd = fd;
+		this.final_filesize = final_filesize;
+		this.channel = channel;
+	}
+
+	public static Fallocate forChannel(FileChannel channel, long final_filesize) {
+		try {
+			return new Fallocate(channel, getDescriptor(channel), final_filesize);
+		} catch (final UnsupportedOperationException exception) {
+			return new Fallocate(channel, 0, final_filesize);
```
ArneBab, line 44: It would be good to have a comment here that explains why 0 is the right file descriptor.

ArneBab:
Please stick to the original indentation style. Re-indenting the file breaks annotations to trace changes to commits.

That said: Thank you for working on this!

torusrxxx:
Recreated in #1010

PR #1011, opened 2024-11-30
Add margin-*, padding-*, scroll-snap-* CSS properties

torusrxxx:
Supported the following new CSS properties:

1. margin-block
2. margin-block-end
3. margin-block-start
4. margin-inline
5. margin-inline-end
6. margin-inline-start
7. padding-block
8. padding-block-end
9. padding-block-start
10. padding-inline
11. padding-inline-end
12. padding-inline-start
13. scroll-behavior
14. scroll-snap-align
15. scroll-snap-stop
16. scroll-snap-type

### DIFF
--- a/src/freenet/client/filter/CSSTokenizerFilter.java
+++ b/src/freenet/client/filter/CSSTokenizerFilter.java
@@ -238,10 +238,16 @@ class CSSTokenizerFilter {
 		allelementVerifiers.add("list-style-position");
 		allelementVerifiers.add("list-style-type");
 		allelementVerifiers.add("list-style");
-		allelementVerifiers.add("margin-right");
-		allelementVerifiers.add("margin-left");
-		allelementVerifiers.add("margin-top");
+		allelementVerifiers.add("margin-block");
+		allelementVerifiers.add("margin-block-end");
+		allelementVerifiers.add("margin-block-start");
 		allelementVerifiers.add("margin-bottom");
+		allelementVerifiers.add("margin-inline");
+		allelementVerifiers.add("margin-inline-end");
+		allelementVerifiers.add("margin-inline-start");
+		allelementVerifiers.add("margin-left");
+		allelementVerifiers.add("margin-right");
+		allelementVerifiers.add("margin-top");
 		allelementVerifiers.add("margin");
 		allelementVerifiers.add("max-block-size");
 		allelementVerifiers.add("max-height");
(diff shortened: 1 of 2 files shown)


src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1444,6 +1525,29 @@ else if("right".equalsIgnoreCase(element))
...
 			allelementVerifiers.remove(element);
 		}
+		else if("scroll-behavior".equalsIgnoreCase(element))
+		{
+			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("auto","smooth"),ElementInfo.VISUALMEDIA));
+			allelementVerifiers.remove(element);
+		}
+		else if("scroll-snap-align".equalsIgnoreCase(element))
+		{
+			auxilaryVerifiers[82]=new CSSPropertyVerifier(Arrays.asList("none","start","end","center"),null,null,null,true);
```
bertm, line 1535: Is there any reason for not describing this `auxilaryVerifiers` entry and the ones for `scroll-snap-type` in the static initializer together with the existing ones?

src/freenet/client/filter/CSSTokenizerFilter.java
```diff
@@ -1211,6 +1232,36 @@ else if("margin-bottom".equalsIgnoreCase(element))
...
+			allelementVerifiers.remove(element);
+		}
+		else if("margin-block-end".equalsIgnoreCase(element))
+		{
+			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("auto"),ElementInfo.VISUALMEDIA,Arrays.asList("le","pe")));
+			allelementVerifiers.remove(element);
+		}
+		else if("margin-block-start".equalsIgnoreCase(element))
+		{
+			elementVerifiers.put(element,new CSSPropertyVerifier(Arrays.asList("auto"),ElementInfo.VISUALMEDIA,Arrays.asList("le","pe")));
```
bertm, line 1247: Can this refer to `auxilaryVerifiers[36]` as `36` instead of hard-coding `length | percentage`?

(same for `margin-block-end`, similar for all the others padding/margin variants added)

bertm (changes requested):

src/freenet/client/filter/CSSTokenizerFilter.java
torusrxxx: There are lots of such cases, I think it is because if there's no use of this property in the CSS, then it does not need to initialize this entry.

bertm (approved):

ArneBab:
https://github.com/hyphanet/fred/pull/1011/commits/b4d5b8328cae532a52b4a83afc165f4b3975e16b looks good, the rest is reviewed by bertm ⇒ merged. Thank you!

PR #1012, opened 2024-11-30
Don't synchronize on the class while writing in FileUtil.fill(...)

bertm:
By substituting the CipherInputStream that is used as a random number source by a simple MersenneTwister, we no longer need to synchronize on the FileUtil class for each 32k bytes written. The MersenneTwister itself is sufficiently cheap to create on the fly.


ArneBab (approved):
This is neat!

PR #1013, opened 2024-11-30
Use NIO helpers in FileUtils for copy/move/rename

bertm:
Copy was implemented through InputStream/OutputStream. The OS will be perfectly capable of doing this, no need to read and write the individual bytes. Use NIO's `Files.copy` to perform the copy instead.

Rename and move had two separate implementations without a compelling difference (apart from `renameTo` being unable to move across filesystem boundaries). Unify the two implementations and use the NIO's `Files.move` to move the file atomically, or if that fails, non-atomically.


ArneBab (approved):
looks good to me

PR #1014, opened 2024-11-30
Don't use reflection for updating file attributes

bertm:
As we've dropped support for Java 5 for a while now, we can access these methods without reflection.


ArneBab (approved):
Good find!

PR #1015, opened 2024-11-30
Reduce stream handling duplication in FileUtil

bertm:
There are a lot of `while (not completed reading/writing) { read/write some more! }` implementations in FileUtil for similar purposes. Reuse existing `copy` and `readUTF` implementations where possible.


ArneBab (approved):
looks good to me

ArneBab:
Merged — thank you!

PR #1016, opened 2024-11-30
Remove Logger.OSThread-based PID logging that was always disabled

bertm:
The PID logging is always disabled and there is no API for enabling it. Remove the entire class and all the calls to its logPID method.

### DIFF
--- a/src/freenet/support/Logger.java
+++ b/src/freenet/support/Logger.java
@@ -1,204 +1,73 @@
 package freenet.support;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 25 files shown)


src/freenet/support/Logger.java
```diff
@@ -1,206 +1,17 @@
...
 import freenet.support.FileLoggerHook.IntervalParseException;
 import freenet.support.LoggerHook.InvalidThresholdException;
-import freenet.support.io.Closer;
 
 /**
  * @author Iakin
 
 */
 public abstract class Logger {
-	public final static class OSThread {
```
ArneBab, line 25: used in SCGIPublisher ⇒ can’t just remove.

ArneBab: ⇒ we need a shim (that doesn’t do anything).

bertm: No sane plugin^W^W^W Restored the Logger$OSThread methods, but removed their implementation.
Thanks for catching!

ArneBab: Thank you! SCGIPublisher is actually sane and useful (hosting hyphanet inproxies with strict allow-list), but it is also old and unmaintained.

ArneBab (approved):

ArneBab:
merged — thank you!

PR #1017, opened 2024-11-30
Refactor TimeUtil: use Java 8 Date-Time API and improve time intervals

bertm:
The time intervals were subtly broken:
* Parsing 0.1s would yield 1 ms instead of the correct 100 ms. Fix this by parsing the seconds.millis as fractional seconds.
* Formatting would format fractional seconds locale-sensitively where parsing would assume point as a decimal separator, so toMillis(fromMillis(...)) would fail depending on locale. Fix this by always formatting in the root locale (decimal point).

As a tiny bonus, time interval parsing is now approximately 3x faster.

### DIFF
--- a/src/freenet/support/TimeUtil.java
+++ b/src/freenet/support/TimeUtil.java
@@ -24,9 +24,15 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
-import java.text.DecimalFormat;
-import java.text.SimpleDateFormat;
-import java.util.*;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
+import java.util.Date;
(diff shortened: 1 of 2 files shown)


src/freenet/support/TimeUtil.java
```diff
@@ -135,77 +138,49 @@ public static String formatTime(long timeInterval, int maxTerms) {
...
-                        millis += MILLISECONDS.convert(Short.parseShort(term.substring(0, term.length() - 1)), SECONDS);
-                    }
-                    break;
-                default:
-                    throw new NumberFormatException("Unknown format: " + (sign > 0 ? "" : "-") + timeInterval);
-            }
+        if ((group = matcher.group(1)) != null) { // weeks
+            millis += DAYS.toMillis(7 * Long.parseLong(group));
+        }
+        if ((group = matcher.group(2)) != null) { // days
```
hernic, line 151: Better with else if ?

bertm: That wouldn't work, all `if` clauses can (and should) match independently as they operate on different parts of the input.

ArneBab (approved):
Looks good to me.

PR #1018, opened 2024-11-30
Simplify IO statistics collection for UdpSocketHandler

bertm:
This removes a bunch of complicated but unused code, and makes the stats collector interface more self-explanatory.

### DIFF
--- a/src/freenet/io/comm/IOStatisticCollector.java
+++ b/src/freenet/io/comm/IOStatisticCollector.java
@@ -4,221 +4,36 @@
 package freenet.io.comm;
 
 import java.net.InetAddress;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.LinkedHashMap;
-import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


src/freenet/io/comm/IOStatisticCollector.java
```diff
@@ -4,221 +4,36 @@
...
-	}
-	
-	public int[][] getTotalStatistics() {
-		synchronized (this) {
-			return _getTotalStatistics();
-		}
-	}
+    public void reportReceivedBytes(InetAddress addr, int bytes) {
+        if (isLocal(addr) || bytes <= 0) {
+            return;
```
ArneBab, line 16: I’m pretty sure that we have calculation errors with overhead when most traffic is local, but that does not have to be solved in this PR.

ArneBab (approved):
Looks good to me. I don’t think that the per-address tracking is still needed.

PR #1019, opened 2024-11-30
Rewrite USKDateHint using Java 8 Date-Time APIs (fixing week hints)

bertm:
This also fixes a bug with the week-based hints returning the same value for the start of the year and the end of the year.


ArneBab (approved):
that’s really neat — and getting rid of the effectively deprecated Calendar.

PR #1020, opened 2024-11-30
Remove obsolete methods from Logger that are deprecated since 10+ years

bertm:
LogLevel enum based log levels are preferred over int-based priorities since 2010. Remove the obsolete int-based methods, the int priority constants and their int-to-enum conversion method.

### DIFF
--- a/src/freenet/support/Logger.java
+++ b/src/freenet/support/Logger.java
@@ -216,35 +216,7 @@ public abstract class Logger {
 		public boolean matchesThreshold(LogLevel threshold) {
 			return this.ordinal() >= threshold.ordinal();
 		}
-		
-		@Deprecated
-		public static LogLevel fromOrdinal(int ordinal) {
-			for(LogLevel level : LogLevel.values()) {
-				if(level.ordinal() == ordinal)
-					return level;
-			}
-			
-			throw new RuntimeException("Invalid ordinal: " + ordinal);
-		}
 	}
-
-	@Deprecated
-	public static final int ERROR = LogLevel.ERROR.ordinal();
-
-	@Deprecated
-	public static final int WARNING = LogLevel.WARNING.ordinal();
-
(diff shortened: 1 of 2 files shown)


src/freenet/support/Logger.java
```diff
@@ -216,35 +216,7 @@ public enum LogLevel {
...
-	public static final int ERROR = LogLevel.ERROR.ordinal();
-
-	@Deprecated
-	public static final int WARNING = LogLevel.WARNING.ordinal();
-
-	@Deprecated
-	public static final int NORMAL = LogLevel.NORMAL.ordinal();
-
-	@Deprecated
-	public static final int MINOR = LogLevel.MINOR.ordinal();
```
ArneBab, line 241: These are used from plugins. I do not think we can get rid of them earlier than the next decade, except if we manage to take up maintenance for all our plugins again.

bertm:
Plugins need to be updated first. Unless we introduce a legacy shim mechanism, this will never get to a state where it can be merged.

src/freenet/support/Logger.java
bertm: Yeah, we're getting to a point where fred can evolve less and less because of unmaintained plugin compatibility.

Until we have a solution for that (like introducing a shim that provides compatibility for legacy plugins), we'll have to live with that. Closing this PR.

ArneBab: Yes, that’s a reality we have to live with.

The steps to move forward would be to first create an actual defined plugin API (which does not expose the full node) and when there are plugins using it and it’s proven itself to be sufficiently powerful, analyze all known plugins that do not use the new API which parts of fred they use. Then annotate all those places as `@Published(plugins=["...", "..."])` or such to track when plugin updates enable removing those limitations.

I’m pretty sure Bombe already has ideas for such an API.

PR #1021, opened 2024-11-30
Remove dangerous Logger.fatal(...) method

bertm:
No sane logger method has System.exit() as a side effect. Fortunately this log method is never used anywhere in Fred. Let's assume it is not used in plugins either and remove it.


ArneBab (approved):
I grepped through all the plugins I have locally and found no Logger.fatal calls, so I think we can assume it’s unused.

ArneBab:
Merged — thank you!

PR #1026, opened 2024-12-15
Fix typo in text-decoration filter

torusrxxx:
There was a typo, wave is not a valid text-decoration value.


ArneBab (approved):

ArneBab:
merged — thank you! (including the zoom change)

PR #1027, opened 2024-12-28
Fix NullPointerException when disabling and reenabling SSL

torusrxxx:
Setting `keyStore` to `null` will result in `NullPointerException` when SSL is reenabled and `loadKeyStoreAndCreateCertificate` is executed. Load a null key store instead.
Avoid adding `strict-transport-security` header twice. The second addition of `strict-transport-security` header does not check whether fproxy SSL is enabled because `container` is not a static member. As a result, `strict-transport-security` header is sent when SSL is enabled but fproxy SSL is disabled.

### DIFF
--- a/src/freenet/clients/http/ToadletContextImpl.java
+++ b/src/freenet/clients/http/ToadletContextImpl.java
@@ -440,11 +440,6 @@ public class ToadletContextImpl implements ToadletContext {
 		mvt.put("x-content-security-policy", contentSecurityPolicy);
 		mvt.put("x-webkit-csp", contentSecurityPolicy);
 		mvt.put("x-frame-options", allowFrames ? "SAMEORIGIN" : "DENY");
-		String HSTS = SSL.getHSTSHeader();
-		if(!HSTS.isEmpty() && !mvt.containsKey("strict-transport-security")) {
-			// SSL enabled, set strict-transport-security so that the user agent upgrade future requests to SSL.
-			mvt.put("strict-transport-security", HSTS);
-		}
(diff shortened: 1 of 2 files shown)


src/freenet/clients/http/ToadletContextImpl.java
```diff
@@ -440,11 +440,6 @@ static void sendReplyHeaders(
 		mvt.put("x-content-security-policy", contentSecurityPolicy);
 		mvt.put("x-webkit-csp", contentSecurityPolicy);
 		mvt.put("x-frame-options", allowFrames ? "SAMEORIGIN" : "DENY");
-		String HSTS = SSL.getHSTSHeader();
-		if(!HSTS.isEmpty() && !mvt.containsKey("strict-transport-security")) {
-			// SSL enabled, set strict-transport-security so that the user agent upgrade future requests to SSL.
-			mvt.put("strict-transport-security", HSTS);
-		}
```
ArneBab, line 447: Does this disable HSTS?

torusrxxx: HSTS headers are already added at line 237 and with proper check of whether fproxy SSL is enabled. But I forgot to delete here.

torusrxxx: Having both SSL and fproxy SSL configuration is confusing TBH. Especially you can't modify fproxy SSL on the web interface, it looks like SSL configuration is not working.

ArneBab (approved):

PR #1028, opened 2024-12-28
Make wizard save button say "save and run" and make it internationali…

ArneBab:
…zable

This matches betterw hat it actually does.


bertm:
The usual way of finishing a wizard would be by means of the `Finish` button.



Would it be an idea to use that familiar phrasing?

ArneBab:
Sounds good (better).

bertm (approved):

Bombe (approved):

ArneBab:

PR #1031, opened 2025-01-04
fix busy loop if all darknet peers are connected.

ArneBab:
This needs to be merged and released as emergency release, because it turns darknet nodes with 100% connected nodes into electrical heaters.


bertm (approved):
Yikes, good catch, the early return had unintentional side effects.

This could easily have been avoided if the code performing the work would not have been mixed with the code in charge of scheduling that work. That could be accomplished by moving the waiting code from `realRun()` to `run()`.

The current change by itself is fine though.

ArneBab:

PR #1039, opened 2025-02-15
delete file for Travis CI

qupo1:
We currently use GitHub Actions.


ArneBab:
Thank you for the Pull Request.

We shouldn’t remove that, at the moment, though, because there are currently still usages of Travis, so removal would cause work elsewhere.

PR #1041, opened 2025-02-16
Reduce sleep times: we now route far more packages to more peers

ArneBab:


src/freenet/node/CHKInsertSender.java
```diff
@@ -991,9 +991,9 @@ private boolean waitForBackgroundTransfers(BackgroundTransfer[] transfers) {
 					
 					if(logMINOR) Logger.minor(this, "Waiting: transfer completion=" + completedTransfers + " notification="+completedNotifications); 
 					try {
-						backgroundTransfers.wait(SECONDS.toMillis(100));
+						backgroundTransfers.wait(SECONDS.toMillis(10));
```
bertm, line 994: This is not a sleep, but we're waiting for some other thread to wake us up (through `backgroundTransfers.notifyAll()`) after completing (successful or unsuccessful) a transfer, so we will loop until that has happened regardless of the maximum wait time here.

If anything, we should take our `deadline` into consideration that is checked at the start of the `while (true)` loop (currently set in the constructor: 1 minute for realtime, 5 minutes for bulk).

src/freenet/node/CHKInsertHandler.java
```diff
@@ -197,7 +197,7 @@ private void realRun() {
             synchronized(sender) {
                 try {
                 	if(sender.getStatus() == CHKInsertSender.NOT_FINISHED)
-                		sender.wait(5000);
+                		sender.wait(500);
```
bertm, line 200: CHKInsertSender notifies all waiter when the status (or receivedRejectedOverload) changes. From what I can tell reducing this timeout value only increases the frequency of the `while (true)` loop this is in, but should have no other measurable effect.

bertm (commented):
I'm not sure if this will have any measurable effect - at least it should not, as the timeout is merely a maximum time, not a fixed one.

If there is an effect, it's probably due to a bug elsewhere (preventing this thread from getting notified when there are relevant changes) - when that's true we might be better off chasing that bug instead.

ArneBab:
Then let’s close this. No need to muddy up already unclear code with not-clear-to-have-an-effect changes.

PR #846, opened 2023-07-29
Add core operations interface WIP

ArneBab:
Adding an interface to make inviting friends and communicating with friends easier and more obvious.

PR #1043, opened 2025-03-03
Quote n2ntms when replying

ArneBab:


src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -223,11 +225,35 @@ public HTMLNode renderAlert(UserAlert userAlert) {
...
 		alertContentNode.addChild(userAlert.getHTMLText());
+		if (userAlert.canReply()) {
+			alertContentNode.addChild(renderReplyButton(userAlert, userAlert.getSourceNode()));
+		}
 		alertContentNode.addChild(renderDismissButton(userAlert, null));
 
 		return userAlertNode;
 	}
 
+	public HTMLNode renderReplyButton(UserAlert userAlert, WeakReference<PeerNode> peerNode) {
```
bertm, line 236: This method could be `private`.

src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -223,11 +225,35 @@ public HTMLNode renderAlert(UserAlert userAlert) {
...
+		HTMLNode form = new HTMLNode("form",
+					new String[]{"method", "action"},
+					new String[]{"post", "/send_n2ntm/"});
+			form.addChild("input",
+					new String[]{"hidden", "name", "value"},
+					new String[]{"true", "replyTo", userAlert.getMessageText()
+							.replaceAll("\n", "NEWLINE")});
```
bertm, line 243: Technically this could work, but it does feel a bit hacky.

That said, even without the replacement it should technically still be valid HTML5 (with the form field having the String value `"Foo\n\tBar"` in the example below):
```html
<input type="hidden" value="Foo
	Bar">
```

Unfortunately, the HTMLNode indentation might mess up the message.

More suitable approaches might be:
* Encoding the entire message in some reasonable encoding that does not produce newlines, like Base64.
* Referencing the original message itself (by its identifier) as an additional parameter, and retrieving the message on-the-fly when rendering the compose message form.

src/freenet/node/useralerts/UserAlert.java
```diff
@@ -50,6 +60,18 @@ public interface UserAlert {
...
+	public default String replyButtonText() {
+		return dismissButtonText();
+	};
+
+	public default boolean canReply() {
+		return false;
+	}
+
+	public default WeakReference<PeerNode> getSourceNode() {
+		return null;
```
bertm, line 72: With this interface, the caller needs to deal with 2 variants of "source node not known": either the WeakReference can be `null` (when this method is not implemented), or its referent can be `null` (when cleared by the garbage collector).

Can we just return `PeerNode` here? Of course this only concerns the return type of this method; implementations that do hold a (weak)reference would just return `reference.get()`.

src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -223,11 +225,35 @@ public HTMLNode renderAlert(UserAlert userAlert) {
...
+			form.addChild("input",
+					new String[]{"hidden", "name", "value"},
+					new String[]{"true", "peernode_hashcode", Integer.valueOf(peerNode.get().hashCode()).toString()});
```
bertm, line 246: `peerNode.get()` would be `null` if the peer node has been garbage collected, this would throw a NPE .

src/freenet/node/useralerts/UserAlert.java
```diff
@@ -50,6 +60,18 @@ public interface UserAlert {
 	
 	public String dismissButtonText();
 	
+	public default String replyButtonText() {
```
bertm, line 63: The newly added methods feel out of place on `UserAlert`, as they are not so much about the alert itself, but quite specific to N2N messaging.

If we would replace the `if (userAlert.canReply()) {` in `UserAlertManager` to check if the UserAlert is an `N2NTMUserAlert`, we don't need to pollute this interface with N2N messaging specifics.

src/freenet/node/useralerts/UserAlert.java
ArneBab: That’s nicer, yes. Fixed. I also added null-safety.

src/freenet/node/useralerts/UserAlertManager.java
ArneBab: fixed -- thank you!

src/freenet/node/useralerts/UserAlert.java
ArneBab: You’re right -- it’s actually only needed in a single place. Removed.

src/freenet/node/useralerts/UserAlertManager.java
ArneBab: I’m now encoding as base64. Since this is hidden anyway, that’s cleaner. Thank you!

src/freenet/node/useralerts/UserAlertManager.java
ArneBab: fixed - thank you!

src/freenet/node/useralerts/DownloadFeedUserAlert.java
```diff
@@ -10,7 +10,7 @@
 import freenet.node.PeerNode;
 import freenet.support.HTMLNode;
 
-public class DownloadFeedUserAlert extends AbstractUserAlert {
+public class DownloadFeedUserAlert extends AbstractUserAlert implements UserAlertFromPeer {
```
Bombe, line 13: In my opinion, this is going into the wrong direction. You are putting more and more content into the user alerts, but it doesn’t belong there. If you get a message from someone, _that message_ should contain everything you need. User alerts are for showing, well, alerts. You can totally _generate a user alert_ from e.g. a message that you have received, but none of the actual message data should be included in the alert.

ArneBab: From what I can see, there is no message. The Alerts are generated directly from SFS’s in the peer notes. We currently have this double-structure where we can answer some alerts and others we cannot, so this actually are two different concerns, munged into one structure. There are even some we can accept or reject (file offers).

We could disentangle that into different objects with actual data handling (instead of managing data in what’s effectively UI objects) that get dispatched by type, but that requires more refactoring and it’s not what I currently want to achieve: my current short-term goal for next release is polishing user interface warts in friend-to-friend workflows.

But I see extracting dedicated interfaces as a path towards cleaning up the structure: once the different types of alerts are represented by interfaces we can more easily swap in proper data handling (and for example have dedicated peer communication commands via FCP).

Bombe: > From what I can see, there is no message. The Alerts are generated directly from SFS’s in the peer notes. We currently have this double-structure where we can answer some alerts and others we cannot, so this actually are two different concerns, munged into one structure. There are even some we can accept or reject (file offers).

That is exactly what I see as biggest part of the problem. N2NTMs are not entities in their own right, they have been violently shoe-horned into a system that was not made for them, and the thing we absolutely should _not_ do is entrench the notion that it’s okay for the messages (or any other actual data, for that matter) to live in the user alerts.

> We could disentangle that into different objects with actual data handling (instead of managing data in what’s effectively UI objects) that get dispatched by type, but that requires more refactoring and it’s not what I currently want to achieve: my current short-term goal for next release is polishing user interface warts in friend-to-friend workflows.

So you are intentionally increasing technical debt for a (questionable!) short-term goal, making it harder for us to ever fix the underlying problem? That’s quite rude! 🙂

(Why do I think your goal is questionable? Because in my opinion the biggest problems N2NTMs face – besides the name 😬 – is that nobody knows they exist, and if somebody does know they exist, they don’t know what to do with them. Adding “quote on reply” to functionality that is already obscure _at best_ will do _nothing_ to further its adoption, but it definitely will do is make it harder for us to actually separate N2NTMs – we need to fix this name – and user alerts, as well as everything else that is currently existing only in user alerts, making all of it available with a well-defined interface in the node and via FCP, so that you can use different clients to view and manage your N2NTMs.)

> But I see extracting dedicated interfaces as a path towards cleaning up the structure: once the different types of alerts are represented by interfaces we can more easily swap in proper data handling (and for example have dedicated peer communication commands via FCP).

As long as all the data is living in the user alerts, you will never have this separation, and forcing it on clients to use `instanceof` everywhere in their code just to know if this random user alert does maybe contain a message is quite the nasty code smell.

The only forward here (again, in my opinion) is to elevate N2NTMs to be an entity in their own right, with its own in-node infrastructure and FCP interface, and then the functionality you are trying to add here will only touch the N2NTM management code and not the user alerts.

ArneBab: I think the notion that no one knows that n2ntms exists is wrong. I’ve been seeing problem reports about them for years and people with darknet peers are actively using them to share bookmarks and to write messages. My first fix to the n2ntm’s was in 2018 (to lift the 1k size restriction). I think the reason we’re not seeing those reports far more often is that few people use darknet. Which is partly because our one distinguishing feature -- truly confidential messages -- has lots of UI warts.

The naming is fixed in the UI already ("send confidential message").

What I’m doing therefore is to extract a minimal interface so it doesn’t matter for the touched UI code where the data lives or whether the object is a UserAlert at all. That way we can split off storage of node messages from user alerts and only combine them in the alerts UI (for continuity), so other places can more easily get only the node to node messages (I need those as separate list in the core operations interface I’ve already been planning for too long).

I think we have to move to a better structure incrementally, because other approaches regularly fail either in getting stuck or in overloading reviewers.

… maybe the Interface should then not be called `UserAlertFromNode` but something like `MessageFromNode` or `FriendNote` (though message is already taken for FCP message -- do you have a better suggestion?).

Bombe: > I think the reason we’re not seeing those reports far more often is that few people use darknet. Which is partly because our one distinguishing feature -- truly confidential messages -- has lots of UI warts.

It’s hard to know which one is the cause and which one is the symptom. 😀

> What I’m doing therefore is to extract a minimal interface so it doesn’t matter for the touched UI code where the data lives or whether the object is a UserAlert at all. That way we can split off storage of node messages from user alerts and only combine them in the alerts UI (for continuity), so other places can more easily get only the node to node messages (I need those as separate list in the core operations interface I’ve already been planning for too long).

I would absolutely prefer if you would do _that_ first… 🙂 

> I think we have to move to a better structure incrementally, because other approaches regularly fail either in getting stuck or in overloading reviewers.

We absolutely should do that, but that change is a move into the wrong direction, in my opinion, because you are making the structure worse.

> … maybe the Interface should then not be called `UserAlertFromNode` but something like `MessageFromNode` or `FriendNote` (though message is already taken for FCP message -- do you have a better suggestion?).

I don’t have strong feelings either way about the name of _this_ interface; I believe its existence to be a mistake. 🙂

ArneBab: Core Problem is: I can’t get such a refactoring done in time for the next release (possibly next weekend), and I would really like to get communication with n2n messages more usable for the next release. If you see something that can be done without risk in a very short amount of time, I can try, otherwise I would really prefer to first get the UI improvement in and work on the structure in the background when I take up https://github.com/hyphanet/fred/pull/846 again (which is the first larger change I want to do after the release).

That needs messages by peers separate from UserAlerts.

And yes, cause and symptom are hard to distinguish.

Bombe: > Core Problem is: I can’t get such a refactoring done in time for the next release (possibly next weekend), and I would really like to get communication with n2n messages more usable for the next release.

Using a timelimit you set yourself as a reason to get your change into Fred feels like strong-arming. 😄 

> If you see something that can be done without risk in a very short amount of time

That kind of implies that your change is without risk, which it decidedly is not.

> I would really prefer to first get the UI improvement in and work on the structure in the background when I take up #846 again (which is the first larger change I want to do after the release).

Adding a function in a place as hidden as the list of 800 minor warnings is not a UI improvement; it makes our UI harder to use! An actual UI improvement would be to a) fix the name of N2NTMs (every time I’m thinking the name my brain stumbles), b) add a section in the menu bar that deals _only_ with those messages, and c) expose them via other UIs (such as FCP).

I can totally understand that you want this change to go in, and I can’t stop you, but I don’t think it’s a good change; while the intention may be pretty cool, the implementation currently does not allow this change to be made without fucking things up more than before.

ArneBab: I’m sorry that it appears like strong-arming. The timelimit is not because I particularly like the next weekend, but because it is likely the only weekend in the next months that I can mostly use freely.

But since this disturbs you that much, I’ll leave this PR out of the next release. As long as there are no new vulnerabilities we have to fix before making the code public, release 1503 should be doable without needing a dedicated weekend for it, so we should be good then, hopefully with this PR cleaned up sufficiently for inclusion.

Bombe: > I’m sorry that it appears like strong-arming. The timelimit is not because I particularly like the next weekend, but because it is likely the only weekend in the next months that I can mostly use freely.
> 
> But since this disturbs you that much, I’ll leave this PR out of the next release. As long as there are no new vulnerabilities we have to fix before making the code public, release 1503 should be doable without needing a dedicated weekend for it, so we should be good then, hopefully with this PR cleaned up sufficiently for inclusion.

You completely misunderstood the points of my last message. It’s not about the timelimit, it’s about that you are using the limit you yourself set as a reason to speed up integration of this PR. And it’s not about the form of this PR, it’s about what you are trying to do in the PR.

ArneBab: I understood that. And with "form" of this PR I meant adding to UserAlerts instead of refactoring first.

src/freenet/clients/http/N2NTMToadlet.java
```diff
@@ -138,6 +146,25 @@ public void handleMethodPOST(URI uri, HTTPRequest request, ToadletContext ctx)
...
+			try {
+				message = Base64.decodeUTF8(message);
+			} catch (IllegalBase64Exception e) {
+				Logger.error(this, "could not decode replyTo text", e);
+				message = "";
+			}
+			if (!message.isEmpty()) {
+				message = "\n> " + String.join("\n> ", message.split("\n")) + "\n\n";
+				message = message.replaceAll("\n> >", "\n>>")
+						.substring(1); // strip the initial linebreak again
```
Bombe, line 161: Oof, this block is hard to understand. 🙂 

```suggestion
                message = Arrays.stream(message.split("\n"))
                    .map(line -> "> " + line)
                    .map(line -> line.replaceFirst("^> >", ">>"))
                    .collect(joining("\n")) + "\n\n";
```

This is straight from the hip, so it might need a reality check, but this captures what (I think!) you’re trying to do quite nicely.

src/freenet/node/useralerts/UserAlert.java
```diff
@@ -24,6 +25,13 @@ public interface UserAlert {
 	 */
 	public String getText();
 
+	/**
+	 * Content of the body of the alert (no metadata).
+	 */
+	default public String getMessageText() {
+		return getText();
+	}
```
Bombe, line 33: Uhm… that’s a _third_ method now that returns the text of the user alert. Why do we need a _third_ method for this? All (one) places this is used from have a `N2NTMUserAlert` instance which implements the `UserAlertFromPeer` interface, giving us access to this method. There is no reason for further polluting `UserAlert`.

src/freenet/node/useralerts/UserAlert.java
```diff
@@ -4,6 +4,7 @@
 package freenet.node.useralerts;
 
 import freenet.clients.fcp.FCPMessage;
+import freenet.node.PeerNode;
```
Bombe, line 7: I wonder what this import is about, as apparently no other change in this file even so much as thinks of mentioning `PeerNode`… 😁

src/freenet/node/useralerts/UserAlertFromPeer.java
```diff
@@ -0,0 +1,13 @@
...
+ * Public License, version 2 (or at your option any later version). See
+ * http://www.gnu.org/ for further details of the GPL. */
+package freenet.node.useralerts;
+
+import freenet.node.PeerNode;
+
+public interface UserAlertFromPeer {
+
+	public PeerNode getSourceNode();
+    public String getMessageText();
```
Bombe, line 11: You know, the more I think about it, the more I think that this method should mirror the one that already exists on `UserAlert`. As these two interfaces seem to be kind of linked, they can probably share methods, too. 🙂

src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -243,11 +244,37 @@ public HTMLNode renderAlert(UserAlert userAlert) {
 		userAlertNode.addChild("div", "class", "infobox-header", userAlert.getTitle());
 		HTMLNode alertContentNode = userAlertNode.addChild("div", "class", "infobox-content");
 		alertContentNode.addChild(userAlert.getHTMLText());
+		if (userAlert instanceof UserAlertFromPeer) {
+			UserAlertFromPeer n2nUserAlert = (UserAlertFromPeer) userAlert;
+			alertContentNode.addChild(renderReplyButton(n2nUserAlert, n2nUserAlert.getSourceNode()));
```
Bombe, line 249: If you are already handing in the `UserAlertFromPeer`, why are you extracting the source node and handing that in separately?

src/freenet/node/useralerts/UserAlertManager.java
```diff
@@ -243,11 +244,37 @@ public HTMLNode renderAlert(UserAlert userAlert) {
...
+					new String[]{"method", "action"},
+					new String[]{"post", "/send_n2ntm/"});
+			form.addChild("input",
+					new String[]{"hidden", "name", "value"},
+					new String[]{"true", "replyTo", Base64.encodeUTF8(userAlert.getMessageText())});
+			form.addChild("input",
+					new String[]{"hidden", "name", "value"},
+					new String[]{"true", "peernode_hashcode", peerNode == null
+							? ""
+							: Integer.valueOf(peerNode.hashCode()).toString()});
```
Bombe, line 267: Hmm, I’m wondering, if we don’t have a peer node, who are we sending the reply to?

src/freenet/clients/http/N2NTMToadlet.java
```diff
@@ -269,6 +296,11 @@ public static void addUnsentMessageTextInfo(HTMLNode node, String message) {
 	public static void createN2NTMSendForm(boolean advancedMode,
 			HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers)
 			throws ToadletContextClosedException, IOException {
+		createN2NTMSendForm(null, advancedMode, contentNode, ctx, peers, "");
+	}
+	public static void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
```
Bombe, line 301: The `pageNode` is never used.

src/freenet/clients/http/N2NTMToadlet.java
```diff
@@ -269,6 +296,11 @@ public static void addUnsentMessageTextInfo(HTMLNode node, String message) {
...
+			HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers, String initalContent)
+			throws ToadletContextClosedException, IOException {
```
Bombe, line 303: The exceptions aren’t thrown anywhere.

src/freenet/clients/http/N2NTMToadlet.java
Bombe, line 301: And can I convince you to make this method private, because there’s no reason it shouldn’t be? 😁

src/freenet/clients/http/N2NTMToadlet.java
```diff
@@ -269,6 +296,11 @@ public static void addUnsentMessageTextInfo(HTMLNode node, String message) {
 	public static void createN2NTMSendForm(boolean advancedMode,
 			HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers)
 			throws ToadletContextClosedException, IOException {
+		createN2NTMSendForm(null, advancedMode, contentNode, ctx, peers, "");
+	}
+	public static void createN2NTMSendForm(HTMLNode pageNode, boolean advancedMode,
+			HTMLNode contentNode, ToadletContext ctx, HashMap<String, String> peers, String initalContent)
```
Bombe, line 302: There’s also a typo in the `initalContent`.

PR #1045, opened 2025-03-04
Do not show download to disk for large file page on public gateway

ArneBab:
This leaked the username — thank you for the report!

Currently there are no known public gateway nodes running, so exposing the fix here is safe.

There may also be leaked IP values that still need to be fixed: patches welcome (to speed this up).

### DIFF
--- a/src/freenet/clients/http/FProxyToadlet.java
+++ b/src/freenet/clients/http/FProxyToadlet.java
@@ -320,11 +320,9 @@ public final class FProxyToadlet extends Toadlet implements RequestClient {
 				        netLevel == NETWORK_THREAT_LEVEL.MAXIMUM))
 				filterChecked = false;
 		}
-		boolean downloadDisabled = !core.isDownloadDisabled()
-			&& (!ctx.getContainer().publicGatewayMode() || ctx.isAllowedFullAccess());
 		//Display FProxy option to download to disk if the user isn't at maximum physical threat level
 		//and hasn't disabled downloading to disk.
-		if (threatLevel != PHYSICAL_THREAT_LEVEL.MAXIMUM && !downloadDisabled) {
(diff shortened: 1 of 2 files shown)


src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -320,9 +320,11 @@ private static void addDownloadOptions(ToadletContext ctx, HTMLNode optionList,
 				        netLevel == NETWORK_THREAT_LEVEL.MAXIMUM))
 				filterChecked = false;
 		}
+		boolean downloadDisabled = !core.isDownloadDisabled()
+			&& (!ctx.getContainer().publicGatewayMode() || ctx.isAllowedFullAccess());
```
bertm, line 324: Regarding: `boolean downloadDisabled = !core.isDownloadDisabled()`

Is it disabled or not disabled? This logic seems contradicting; or is the naming flawed?

Assuming the naming is OK, per the PR title, I would expect the logic to read:

```java
boolean downloadDisabled = 
        // download is either disabled fully on this node
        core.isDownloadDisabled()
        // or we're accessing in public gateway mode and do not have full access
        || (ctx.getContainer().publicGatewayMode() && !ctx.isAllowedFullAccess())
```
(this would match the previous authorization logic in this class, combined with the authorization logic in QueueToadlet)

bertm (changes requested):

src/freenet/clients/http/FProxyToadlet.java
ArneBab: You’re right -- thank you for catching this!

ArneBab: This is now fixed, and many more places check for public gateway mode. Thank you again!

Bombe (approved):

ArneBab:
Merged -- thank you for the reviews!

PR #1046, opened 2025-03-23
Add dismiss all alerts and delete all messages buttons.

ArneBab:


src/freenet/clients/http/UserAlertsToadlet.java
```diff
@@ -36,18 +43,64 @@ public void handleMethodGET(URI uri, HTTPRequest request, ToadletContext ctx) th
...
+			}
+		}
+
+		if (request.isPartSet(DELETE_ALL_MESSAGES_PART) && request.isPartSet(REALLY_DELETE_AFFIRMED)) {
+			for (UserAlert alert : ctx.getAlertManager().getAlerts()) {
+				if (!(alert instanceof NodeToNodeMessageUserAlert)) {
+					continue;
+				}
+				if (alert instanceof AbstractNodeToNodeFileOfferUserAlert) {
+
```
ArneBab, line 95: Empty branch ⇒ delete file offers or not?

src/freenet/node/DarknetPeerNode.java
```diff
@@ -1019,7 +1021,17 @@ public void onRejected() {
 		}
 
 		protected void onReceiveFailure() {
-			UserAlert alert = new AbstractUserAlert() {
+			UserAlert alert = new AbstractNodeToNodeFileOfferUserAlert() {
```
Bombe, line 1024: This block (and the next two) lead me to believe that maybe you want to add another constructor to `AbstractNodeToNodeFileOfferUserAlert`? 🙂

(This ignores my even bigger belief that user alerts should have little to zero functionality in them, and everything we do here will make it harder for us to change that in the future.)

src/freenet/node/useralerts/AbstractNodeToNodeFileOfferUserAlert.java
```diff
@@ -0,0 +1,17 @@
+package freenet.node.useralerts;
+
+import freenet.support.HTMLNode;
+
+/**
+ * Wrapper class to create anonymous classes with interface NodeToNodeMessageUserAlert
```
Bombe: This is not a wrapper class, it doesn’t wrap anything.

src/freenet/node/useralerts/NodeToNodeMessageUserAlert.java
```diff
@@ -0,0 +1,11 @@
+package freenet.node.useralerts;
+
+import java.lang.ref.WeakReference;
+
+import freenet.node.PeerNode;
+import freenet.support.HTMLNode;
+
+public interface NodeToNodeMessageUserAlert {
```
Bombe, line 11: So, a node-to-node-message user alert consists of a `PeerNode` (okay, that makes sense) and the source node’s name, which sounds like it should be a part of the `PeerNode`? Why is this duplicated here? If this is a user alert (as the type name heavily implies), why does it have nothing else that a user alert has? Also, it has nothing of the node-to-node message, which the type name also implies this should have. What is going on here?

src/freenet/node/useralerts/BookmarkFeedUserAlert.java
```diff
@@ -115,4 +115,14 @@ public boolean isValid() {
 			sourceNodeName = pn.getName();
 		return true;
 	}
+
```
Bombe, line 118: Also, these extensions to this class (and the next two classes) really drive home the need for an additional constructor in the abstract class you added. 🙂

src/freenet/clients/http/UserAlertsToadlet.java
```diff
@@ -36,18 +43,64 @@ public void handleMethodGET(URI uri, HTTPRequest request, ToadletContext ctx) th
...
+		if (request.isPartSet(DISMISS_ALL_ALERTS_PART)) {
+			for (UserAlert alert : ctx.getAlertManager().getAlerts()) {
+				if (!alert.userCanDismiss()) {
+					continue;
+				}
+				if (alert instanceof NodeToNodeMessageUserAlert) {
+					continue; // keep all node to node messages
+				}
+				ctx.getAlertManager().dismissAlert(alert.hashCode());
+			}
```
Bombe, line 86: (Sorry for shaving this yak, feel free to ignore, but if you don’t ignore it, make sure I got it right, negations are confusing. 😄)

```suggestion
            Arrays.stream(ctx.getAlertManager().getAlerts())
					.filter(UserAlert::userCanDismiss)
					.filter(alert -> !(alert instanceof NodeToNodeMessageUserAlert))
					.map(Object::hashCode)
					.forEach(ctx.getAlertManager()::dismissAlert);
```

src/freenet/clients/http/UserAlertsToadlet.java
```diff
@@ -36,18 +43,64 @@ public void handleMethodGET(URI uri, HTTPRequest request, ToadletContext ctx) th
...
+					continue;
+				}
+				if (alert instanceof AbstractNodeToNodeFileOfferUserAlert) {
+					continue; // not deleting file offers, because these need a decision
+				}
+				if (!alert.userCanDismiss()) {
+					continue;
+				}
+				ctx.getAlertManager().dismissAlert(alert.hashCode());
+			}
```
Bombe, line 101: (More yak shaving.)

```suggestion
            Arrays.stream(ctx.getAlertManager().getAlerts())
					.filter(NodeToNodeMessageUserAlert.class::isInstance)
					.filter(alert -> !(alert instanceof AbstractNodeToNodeFileOfferUserAlert))
					.filter(UserAlert::userCanDismiss)
					.map(Object::hashCode)
					.forEach(ctx.getAlertManager()::dismissAlert);
```

Bombe (approved):

PR #1047, opened 2025-04-05
Privacy  do not check reachability

ArneBab:


bertm:
This class has become a bit of a mess of tangled conditions, and through all the negations, ones and minus-ones it's somewhat difficult to review whether the logic here is sound.

If we define a clever helper method `prefer` that takes a `Predicate<InetAddress>` and returns a `Comparator<InetAddress>` and extract the cached `isReachable` check to a separate method, we could reduce this to the more descriptive below definition using Comparator's `thenComparing` chaining operations (assuming I reverse engineered this correctly :smiley:):

```java
private final Comparator<InetAddress> innerComparator = prefer(Objects::nonNull)
		.thenComparing(prefer(not(InetAddress::isAnyLocalAddress)))
		.thenComparing(prefer(not(InetAddress::isLoopbackAddress)))
		.thenComparing(prefer(not(InetAddress::isLinkLocalAddress)))
		.thenComparing(prefer(is(InetAddress::isSiteLocalAddress).and(this::isReachable)))
		// NOTE: the last 3 lines below can also be replaced by simply 
		//   .thenComparing(InetAddressComparator.COMPARATOR);
		.thenComparing(prefer(is(Inet6Address.class::isInstance)))
		.thenComparingInt(Object::hashCode)
		.thenComparing(InetAddress::getAddress, Fields::compareBytes);
```

With the above defined, the actual `compare` implementation can be simplified to just a few lines:
```java
@Override
public int compare(InetAddress arg0, InetAddress arg1) {
	if (Objects.equals(arg0, arg1)) {
		return 0;
	}
	return innerComparator.compare(arg0, arg1);
}
```

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -45,37 +45,37 @@ public int compare(InetAddress arg0, InetAddress arg1) {
...
 		int b = arg1.hashCode();
-		Boolean reachable0 = reachabilityCache.get(a);
-		Boolean reachable1 = reachabilityCache.get(b);
-		if (reachable0 == null) {
-			try {
-				reachable0 = arg0.isReachable((int) DEFAULT_MAX_PING_TIME);
-			} catch (IOException e) {
-				reachable0 = false;
+		// prefer LAN addresses *if they are routable* over global addresses, because they should stay within VPNs
+		if (!arg0.isSiteLocalAddress() && arg1.isSiteLocalAddress()) {
```
bertm, line 52: This does not not prefer reachable site-local addresses over other site-local addresses, as the comparison is only made when one of the addresses is global. Is that intentional?

ArneBab: This is intentional, because this reachability check caused requests within VPNs to addresses outside the VPN and as such made a node much easier to detect. The reason is that Java falls back on non-standard requests when it cannot ping.

ArneBab:
> This class has become a bit of a mess of tangled conditions, and through all the negations, ones and minus-ones it's somewhat difficult to review whether the logic here is sound.
> …
> ```java
> private final Comparator<InetAddress> innerComparator = prefer(Objects::nonNull)
> ...
> ```

That looks much nicer, yes. I’ll try to get there. Thank you!

ArneBab:
@bertm could you have another look whether this is OK to be merged?

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -17,80 +20,64 @@ public class InetAddressIpv6FirstComparator implements Comparator<InetAddress> {
...
-		} else if(bytes1.length > bytes0.length) {
-			return 1;
+			reachabilityCache.put(hashCode, reachable);
 		}
+		return reachable;
+	}
+
+	// inverted prefer, because not(...) is only documented since Java 11
+	private Comparator<InetAddress> preferNot(Predicate<InetAddress> pred) {
+		return (arg0, arg1) -> -1 * predicateToCompare(pred, arg0, arg1);
```
bertm: This can be simplified to `return prefer(pred).reversed();`.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -17,80 +20,64 @@ public class InetAddressIpv6FirstComparator implements Comparator<InetAddress> {
...
+	}
 
-		// Sort by hash code as fallback. This is fast.
-		if(a > b) return 1;
-		else if(b > a) return -1;
-		return Fields.compareBytes(bytes0, bytes1);
-		// Hostnames in InetAddress are merely cached, equals() only operates on the byte[].
+	private Comparator<InetAddress> prefer(Predicate<InetAddress> pred) {
+		return (arg0, arg1) -> predicateToCompare(pred, arg0, arg1);
```
bertm: This can be simplified to `return (arg0, arg1) -> Boolean.compare(!pred.test(arg0), !pred.test(arg1));`

Also avoids having to test the predicates twice.

src/freenet/support/io/InetAddressIpv6FirstComparator.java
```diff
@@ -17,80 +20,64 @@ public class InetAddressIpv6FirstComparator implements Comparator<InetAddress> {
...
+	}
+
+	private int predicateToCompare(Predicate<InetAddress> pred, InetAddress arg0, InetAddress arg1) {
+		if (pred.test(arg0) && pred.test(arg1)) {
+				return 0;
+			}
```
bertm: The indentation is off, but this code can probably be removed as per the other comments.

bertm (commented):
This looks way more comprehensible already. Here's a few additional but minor remarks.

ArneBab:
Merged - thank you! @Bombe @bertm

PR #1048, opened 2025-04-05
Always include the source code in the freenet.jar

ArneBab:
This would enable to switch to AGPL as license which requires providing all who can connect to the node the sources on request. With the sources in freenet.jar, they are automatically offered via UOM, so however users patch their nodes, they are always automatically complying with the license.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,6 +97,7 @@ task compileVersion (type: JavaCompile) {
 task buildJar(type: Jar) {
     dependsOn compileVersion, processResources
     from(processResources)
+    from(sourceSets.main.java)
     from(compileJava.destinationDirectory) {
         exclude 'freenet/node/Version.class'
         exclude 'freenet/node/Version$1.class'


bertm:
Given that the sources are already publicly available on Github, does this add any value for us? I'm not even sure if publishing `sourceSets.main.java` is sufficient to qualify as "the Corresponding Source of your version" as it omits various files that are necessary for fred to build.

That said, even if this is sufficient, it might be hypothetically relevant for hyphanet forks that do not have a place to publish their source code anywhere accessible (like Github or even on their own network). Those projects would then magically comply with terms of a license that does not even currently apply to hyphanet.

In my opinion, this mostly just increases the size of `freenet.jar` without offering much benefit.

ArneBab:
Then let’s close this.

In discussions in FMS people also brought up the argument, that this would mean that private testing of changes could publish those changes, which would be a problem for a responsibly disclosed vulnerability.

PR #1051, opened 2025-04-26
Fix regression: compress parameter was inverted on upload. Thanks to NewOne@umLZL for investigating!

ArneBab:
The regression came in in a merge between different refactoring PRs.


ArneBab:
@Bombe thank you for your review!

bertm (approved):

PR #1052, opened 2025-05-02
Ask in first time wizard whether to donate disk space

ArneBab:
I’ve heard the argument "I don’t want stuff stored on my disk" so often now that if "it’s only in memory" suffices it may just be worth to have a checkbox "donate disk space?" and if toggled off use a RAM store and cache.

That may reduce the lifetime of files but if in exchange many more people use the network, that’s a net win.

Since at least half the nodes are only online 2 hours per day, it’s likely that caches already provide most of the data, so the effect of more RAM only nodes shouldn’t be too bad.

Besides adding the checkbox I fixed a nasty locking issue that prevented switching from salt-hash to ram and then back to salt-hash. This didn’t hit earlier, because switching to salt-hash required a node restart which masked the locking issue.

Also toggles in the first time wizard can now actually be instant.

### DIFF
--- a/src/freenet/clients/http/FirstTimeWizardNewToadlet.java
+++ b/src/freenet/clients/http/FirstTimeWizardNewToadlet.java
@@ -6,6 +6,8 @@ import freenet.clients.http.wizardsteps.BandwidthManipulator;
 import freenet.clients.http.wizardsteps.DATASTORE_SIZE;
 import freenet.config.Config;
 import freenet.config.ConfigException;
+import freenet.config.InvalidConfigValueException;
+import freenet.config.NodeNeedRestartException;
 import freenet.config.Option;
 import freenet.l10n.NodeL10n;
 import freenet.node.*;
@@ -21,6 +23,7 @@ import java.net.URI;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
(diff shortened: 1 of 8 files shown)


Bombe:
How easy will it be to reverse whatever decision one made during the first install? I know that I oftentimes do not pay that much attention to stuff in checkboxes when I’m installing something, and I don’t want anybody to have to reinstall, because they don’t know how to change stores or whatever…

ArneBab:
They just need to go into core settings as usual and change to salt-hash store.

Bombe:
> They just need to go into core settings as usual and change to salt-hash store.

If you remove the “just” from this sentence, it sounds like something you would first need to know about in order to actually do it…

Juiceman:
If you give the user a RAM only option (which imho is a bad idea), you
should have a nagging banner or message that the user must dismiss on every
startup.  Have the message have 2 buttons: "dismiss msg until restart" and
" launch wizard to configure datastore."

ArneBab:
I wouldn’t want to nag people all the time, because a node without store is still useful to the network (for routing) and this nagging would just cause people to stop running their node.

But  you’re (all) right that we need to make it easy to switch to a persistent store. ⇒ not yet to merge.

RealJosephKnapp:
> I’ve heard the argument "I don’t want stuff stored on my disk" so often now that if "it’s only in memory" suffices it may just be worth to have a checkbox "donate disk space?" and if toggled off use a RAM store and cache.
> 
> That may reduce the lifetime of files but if in exchange many more people use the network, that’s a net win.
> 
> Since at least half the nodes are only online 2 hours per day, it’s likely that caches already provide most of the data, so the effect of more RAM only nodes shouldn’t be too bad.
> 
> Besides adding the checkbox I fixed a nasty locking issue that prevented switching from salt-hash to ram and then back to salt-hash. This didn’t hit earlier, because switching to salt-hash required a node restart which masked the locking issue.
> 
> Also toggles in the first time wizard can now actually be instant.

One possible answer would be to allow the data store to be stored separately from the installation of hyphanet. That way, if you don't want to donate disk space, you can instead put the datastore in the temp directory of your operating system with a one-time self-destructing encryption key. When you reboot your computer, the temp directory would be wiped and the one time self-destructing encryption key would be wiped too. This should be an option because not everyone has lots of ram to spare.

ArneBab:
> When you reboot your computer, the temp directory would be wiped and the one time self-destructing encryption key would be wiped too. This should be an option because not everyone has lots of ram to spare.

You can set the store directory in the Configuration → Installation settings ⇒ /config/node.install

PR #1053, opened 2025-05-04
Make fproxy cross origin isolated

torusrxxx:
Added recommended headers by https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated. This allows freenet tabs to be hosted in separate OS process to mitigate some risks.

### DIFF
--- a/src/freenet/clients/http/ToadletContextImpl.java
+++ b/src/freenet/clients/http/ToadletContextImpl.java
@@ -435,6 +435,8 @@ public class ToadletContextImpl implements ToadletContext {
 		} else {
 			mvt.put("connection", "keep-alive");
 		}
+		mvt.put("cross-origin-embedder-policy", "require-corp");
+		mvt.put("cross-origin-opener-policy", "same-origin");
 		String contentSecurityPolicy = generateCSP(allowScripts, allowFrames);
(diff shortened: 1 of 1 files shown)


src/freenet/clients/http/ToadletContextImpl.java
```diff
@@ -437,6 +437,8 @@ static void sendReplyHeaders(
 		}
 		String contentSecurityPolicy = generateCSP(allowScripts, allowFrames);
 		mvt.put("content-security-policy", contentSecurityPolicy);
+		mvt.put("cross-origin-embedder-policy", "require-corp");
+		mvt.put("cross-origin-opener-policy", "same-origin");
 		mvt.put("x-content-security-policy", contentSecurityPolicy);
```
bertm, line 442: This seems very reasonable (although the headers are new to me). But could you move the new stuff down a few lines so it is not smack in the middle of several lines related to CSP? :smile:

bertm (changes requested):

src/freenet/clients/http/ToadletContextImpl.java
torusrxxx: I don't like adding something headers after the x-headers, so I moved them before CSP.

bertm (approved):

PR #1054, opened 2025-05-06
Add Server-Timing header for performance profiling

torusrxxx:
I wonder why opening http://localhost:8888/strangers/ takes up to 10 seconds and also slows down other freenet network traffic like hitting the brake pedal. The Server-Timing header is integrated into browser's debugging tool, providing an excellent tool to understand performance of each freenet page load.

This header is only enabled for clients with full access to the node.


bertm:
> I wonder why opening http://localhost:8888/strangers/ takes up to 10 seconds and also slows down other freenet network traffic like hitting the brake pedal.

`Server-Timing` is not a suitable approach to answering this question. It will only tell you that some high-level parts of the rendering were slow (of at all) - but it won't provide any insight into why that is the case.

To dive deeper, consider attaching a profiler or a debugger to the JVM to observe your request being handled. The answer is probably somewhere around "doing lots of expensive computation while holding some important locks". With expert mode enabled and on a highly connected node, that page can show a lot of data, especially considering the amount of peers we allow nowadays.

torusrxxx:
Sorry if I made troubles, I will make future updates to this non-public immediately. I'm glad to see you still active!

PR #1055, opened 2025-05-10
Simplify MultiHashInputStream and MultiHashOutputStream

bertm:
Simplify MultiHashInputStream and MultiHashOutputStream by extracting their common multi-digest hashing code into a separate class.

Additionally, ensure that MultiHashInputStream does not advertise or implement mark/reset support (default delegated to the underlying InputStream by FilterInputStream), as this could lead to digesting the same bytes multiple times.

### DIFF
--- a/src/freenet/crypt/MultiHashDigester.java
+++ b/src/freenet/crypt/MultiHashDigester.java
@@ -2,25 +2,26 @@ package freenet.crypt;
 
 import java.security.MessageDigest;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
 
 final class MultiHashDigester {
 
-    private final Digester[] digesters;
+    private final Collection<Digester> digesters;
 
-    private MultiHashDigester(Digester[] digesters) {
+    private MultiHashDigester(Collection<Digester> digesters) {
         this.digesters = digesters;
     }
 
     void update(byte[] input, int offset, int len) {
-        for (Digester digester : digesters) {
-            digester.update(input, offset, len);
-        }
+        digesters.forEach(digester -> digester.update(input, offset, len));
     }
 
-    HashResult[] getResults() {
-        return Arrays.stream(digesters)
+    List<HashResult> getResults() {
+        return digesters.stream()
                 .map(Digester::getResult)
-                .toArray(HashResult[]::new);
+                .collect(Collectors.toList());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/freenet/crypt/MultiHashDigester.java
```diff
@@ -0,0 +1,57 @@
+package freenet.crypt;
+
+import java.security.MessageDigest;
+import java.util.Arrays;
+
+final class MultiHashDigester {
+
+    private final Digester[] digesters;
```
Bombe: I am slightly (okay, a lot 😄) appalled by the unnecessary use of arrays, especially as they’re only being used to iterate over or to turn them into streams. In my opinion, arrays should only be used when interacting with low-level APIs, or in case you really, _really_ need to optimize for speed; in all other cases use something from the Collection API, because it makes the code so much more readable.

src/freenet/crypt/MultiHashDigester.java
```diff
@@ -0,0 +1,57 @@
...
+        this.digesters = digesters;
+    }
+
+    void update(byte[] input, int offset, int len) {
+        for (Digester digester : digesters) {
+            digester.update(input, offset, len);
+        }
+    }
+
+    HashResult[] getResults() {
```
Bombe: Here as well. We lose nothing by returning a `Set` or a `List`, and we gain nothing by using arrays.

src/freenet/crypt/MultiHashInputStream.java
Bombe, line 1: I wonder if I could pester you for tests for the input and output streams, too? 👉👈 😄

Bombe (changes requested):

src/freenet/crypt/MultiHashDigester.java
bertm: I'm trying to remember why this was using array, but it's probably just because the old code did so.
Let's change that to `List`! :rocket:

src/freenet/crypt/MultiHashInputStream.java
bertm: Consider it done! (they could be prettier but this serves the purpose)

Bombe (approved):

ArneBab:

ArneBab (commented):
@bertm This causes a change in the createt CHKs from fcpupload (from pyfreenet3). It shows up, because the hashes don’t match anymore when compression is enabled for larger files.

To reproduce: Download https://mirror.cyberbits.eu/fosdem/2026/k4601/LJXYCF-vlc-ffmpeg-librairies-kyber-2026.av1.webm then upload it. With compression it should give: CHK@fpmrlTBPfcMOyUK2mwh40GAyvPXy8PwLln07PEW4j8g,9X0pllMCjRj3UMDpQVud-11rxrk0CA-ISHJTKkMq0yc,AAMC--8/LJXYCF-vlc-ffmpeg-librairies-kyber-2026.av1.webm
After this change it instead gives:
CHK@pHlBBZBYNTViOjhDvryovT9gTDDzzPb8njhFzaVg6BE,eMQ4IY5i-W3chyWefCds7i6GFqXX5ysfWd9OFVSI32s,AAMC--8/LJXYCF-vlc-ffmpeg-librairies-kyber-2026.av1.webm

PR #1056, opened 2025-05-11
Use JCE AES implementation for AEAD when available

bertm:
On modern Java, the JCE implementation is about twice as fast as BC.


ArneBab (approved):
Looks good to me. I’m surprised to see a crypto change that’s actually easy to review. :+1:

PR #1058, opened 2025-05-11
Fix single-byte `read()` in various InputStream implementations

bertm:
There are multiple places where the read byte is not correctly converted to unsigned int (leading to incorrect data being returned), or where the single-byte `read()` path differs from the multi-byte `read(byte[])` or `read(byte[], int, int)` path.

Additionally, do not fear creating short-lived single-byte buffers - their overhead is minimal and can likely be optimized away to some extent.


Bombe (approved):
You have now written the same method about six times (seven, actually, remembering the other PR)… maybe we should put that into some `InputStreamUtils` helper class?

PR #1063, opened 2025-05-12
Do not fix case of header key

torusrxxx:
There isn't broken HTTP clients. All headers are forced lower-case by HTTP/2. Don't need to do this a dozen times for every request.

### DIFF
--- a/src/freenet/clients/http/ToadletContextImpl.java
+++ b/src/freenet/clients/http/ToadletContextImpl.java
@@ -447,7 +447,7 @@ public class ToadletContextImpl implements ToadletContext {
 		buf.append(replyDescription);
 		buf.append("\r\n");
 		for (Map.Entry<String, List<String>> entry : mvt.entrySet()) {
-			String key = fixKey(entry.getKey());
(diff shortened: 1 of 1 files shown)


bertm (approved):
*Technically* we should use proper casing where we can. [RFC 2616 (HTTP/1.1) section 4.2 Message Headers](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2) states the following:

> Applications ought to follow "common form", where
> one is known or indicated, when generating HTTP constructs, since
> there might exist some implementations that fail to accept anything
> beyond the common forms.

Per the [Hypertext Transfer Protocol (HTTP) Field Name Registry
](https://www.iana.org/assignments/http-fields/http-fields.xhtml) all the header names are capitalized, so that would be the "common form" for generating those header names.

That said: meh, let's not do magic header name rewriting (reasonable client should read them case insensitive). If this does cause an issue (which it won't), fix it at the source (where the incorrectly capatilized header name is used).

HTTP/2 is not relevant though, as fred does not speak that.

PR #1064, opened 2025-05-12
Do not synchronize on global variable in crypt Util.makeKey

bertm:
This method is used by Yarrow (our random number generator) for seeding, which is used in many places. We'd better not have a global bottleneck for its uses.

Instead of sharing a digest globally, just obtain a fresh one when in need.

### DIFF
--- a/src/freenet/crypt/Util.java
+++ b/src/freenet/crypt/Util.java
@@ -190,9 +190,6 @@ public class Util {
 		return freenet.support.Fields.byteArrayEqual(a, b, offset, offset, length);
 	}
 
-	private static final MessageDigest ctx;
-	private static final int ctx_length;
-
 	public static final Map<String, Provider> mdProviders;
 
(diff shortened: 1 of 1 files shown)


src/freenet/crypt/Util.java
```diff
@@ -355,13 +348,12 @@ public static void main(String[] args) throws Exception {
...
-				ctx.update((byte) 'b');
-				ctx.update((byte) 'c');
-				System.err.println(HexUtil.bytesToHex(ctx.digest()));
-			}
+			MessageDigest ctx = HashType.SHA1.get();
+			ctx.digest();
+			ctx.update((byte) 'a');
+			ctx.update((byte) 'b');
+			ctx.update((byte) 'c');
+			System.err.println(HexUtil.bytesToHex(ctx.digest()));
```
Bombe, line 356: This code smells like it should be killed with 🔥.

PR #1069, opened 2025-05-13
Remove unused code and parameters from NodeStats

bertm:
* Remove unused arguments for private methods
* Remove all kinds of unused fields, methods & constructors

Reduces the code size of this rather large class by roughly 110 lines.


bertm:
Note that there more reduction possible in the public methods and fields, but since there's often discussion regarding plugins, this starts with the (effectively) non-public ones.

PR #1071, opened 2025-05-13
Fix request distribution stats

bertm:
Turns out I broke this 9 years ago, so let's fix it :smile: 
For those who don't have a memory of these statistics panels, it's these two:
image

---

The `count` was never incremented, and consequently the Incoming Request Distribution and Outgoing Request Distribution panels would no longer show up on the statistics page.

Instead of keeping track of the total count independent of the bins, just sum the bins when we need a total count for display purposes.

While we're at it, let's get rid of the synchronization too.


Bombe (approved):

Bombe:
You know, sometimes I was wondering where these things went, but I never bothered to actually look for them… 😄

PR #1072, opened 2025-05-15
Fix rel attribute and add rel values

torusrxxx:
Fix rel attribute and add rel values noopener,nofollow,noreferrer,external,tag
Also fixed code style related to adding items to HashSet.

### DIFF
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


bertm (changes requested):
This appears to introduce 3 changes, but does so in 1 commit:
1. Change `for` to `addAll` - this change I'm happy to accept
2. Fix something in the `rel` attribute handling - please explain in the commit message for this change what was broken and how it is fixed & make sure to have a test for the improvement
3. Add new `rel` types, but not all of them - please explain explain in the PR description why we need exactly those to be added (and not some other ones that could also be supported)

In it's current state with the commented out code and the FIXME's unresolved, this PR is not ready to be merged.

I would assume change number 2 in the above list brings most value - if possible, please get that in a mergeable state without the parts that would require further discussion :)

torusrxxx:
Originally I just wanted to fix that doubled `token.equalsIgnoreCase("stylesheet")` check. But then I found more and more things to fix and the scope became unclear.
I now reverted to just coding style fixes. Thanks.

Bombe (approved):

ArneBab:
Merged -- thank you!

PR #1075, opened 2025-05-19
Add accessors for two attributes used in PeerNodeStatus

Bombe:
Once again, I am blocked in writing Tests for Fred by the presence of fields that I cannot mock using Mockito. Have some accessors instead!


bertm (commented):
Could the accessors (and the fields that they access) be package-private to reduce the size of the public API?

Bombe:
> Could the accessors (and the fields that they access) be package-private to reduce the size of the public API?

Hmm, they probably could… depends on whether these methods need to be mocked by something outside of Fred. I don’t think there’s too much stuff using actual `PeerNode` objects outside of Fred, though… then again, plugins might? Hmm.

bertm:
> plugins might?

That's always possible, at least in theory. But most plugins would not have a business tracking peer nodes. Even then, these particular fields are pretty niche. I cannot imagine legitimate reasons for plugins to meddle with the peer backoff statistics.

Bombe:
> > plugins might?
> 
> That's always possible, at least in theory. But most plugins would not have a business tracking peer nodes. Even then, these particular fields are pretty niche. I cannot imagine legitimate reasons for plugins to meddle with the peer backoff statistics.

I mean, it doesn’t even matter, because the original fields are still public. 😄

Bombe:
> Could the accessors (and the fields that they access) be package-private to reduce the size of the public API?

Oof, I really need to learn to read better. 😄 

Yeah, I’ll make the fields `private`, they weren’t used outside of `PeerNode` and `PeerNodeStatus` anyway.

bertm (approved):

PR #1077, opened 2025-05-23
Allow freesite to set robots, googlebot and referrer=no-referrer meta tag

torusrxxx:
Freesites should be able to set the robots tag to tell spiders on freenet to stop publishing them. Invalid robots value is filtered.
Freesites can also set referrer, and no-referrer is the only value supported.

### DIFF
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -30,6 +30,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.Stack;
 import java.util.StringTokenizer;
+import java.util.Arrays;
 
 import freenet.clients.http.ToadletContextImpl;
 import freenet.l10n.NodeL10n;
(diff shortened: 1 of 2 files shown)


src/freenet/client/filter/HTMLFilter.java
```diff
@@ -3001,6 +3020,24 @@ Map<String, Object> sanitizeHash(Map<String, Object> h,
 					} else if (name.equalsIgnoreCase("Viewport")) {
 						hn.put("name", name);
 						hn.put("content", content);
+					} else if (name.equalsIgnoreCase("referrer")) {
+						hn.put("name", name);
+						hn.put("content", "no-referrer"); // Only no-referrer is allowed in freesites for privacy
```
bertm, line 3025: What privacy exactly? This is about what Referer header is sent upon navigation, but considering the destination would always be the node that rendered this page in the first place (external links are bounced through an intermediate page), there's nothing to hide.

Also the implementation does not match the comment - this does not "only allow" `no-referrer`, but forces `no-referrer` whenever any value is set (but does not enforce it when **no** value is set).

I would suggest dropping support for this metadata name.

src/freenet/client/filter/HTMLFilter.java
```diff
@@ -3001,6 +3020,24 @@ Map<String, Object> sanitizeHash(Map<String, Object> h,
...
+					} else if (name.equalsIgnoreCase("robots") || name.equalsIgnoreCase("googlebot")) {
+						StringTokenizer tokenizer = new StringTokenizer(content, ",");
```
bertm, line 3027: From the `StringTokenizer` Javadoc:

> StringTokenizer is a legacy class that is retained for compatibility reasons although its use is discouraged in new code. It is recommended that anyone seeking this functionality use the split method of String or the java.util.regex package instead.

bertm (changes requested):

ArneBab (approved):

PR #1078, opened 2025-05-24
Add HTML elements meter,progress,input type=email,number,search,tel,url

torusrxxx:
New elements [`<meter>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meter), [`<progress>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress), `<input type="email">`, `<input type="number">`, `<input type="search">`, `<input type="tel">`, `<input type="url">`

Some of the input elements on Freenet may better use `<input type="search">` or `<input type="number">` than `<input type="text">`, so we should first support them.


Bombe:
There is not a single test for any of this, even though there easily could be.

PR #1083, opened 2025-06-29
Fix NullPointerException in input tag filter

torusrxxx:
I'm so sorry this bug was introduced, and the tests are completely unable to catch it.

### DIFF
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -2987,6 +2987,9 @@ public class HTMLFilter implements ContentDataFilter, CharsetExtractor {
 			HTMLParseContext pc) throws DataFilterException {
 			Map<String, Object> hn = super.sanitizeHash(h, p, pc);
 
+			// We drop the whole <input> if type is null
+			if(hn.get("type") == null)
+				return null;
 			// We drop the whole <input> if type isn't allowed (case-insensitive)
 			if(!allowedTypes.contains(hn.get("type").toString().toLowerCase())){
 				return null;
--- a/src/freenet/client/filter/HTMLFilter.java
+++ b/src/freenet/client/filter/HTMLFilter.java
@@ -2987,11 +2987,8 @@ public class HTMLFilter implements ContentDataFilter, CharsetExtractor {
 			HTMLParseContext pc) throws DataFilterException {
 			Map<String, Object> hn = super.sanitizeHash(h, p, pc);
 
-			// We drop the whole <input> if type is null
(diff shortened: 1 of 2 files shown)


src/freenet/client/filter/HTMLFilter.java
```diff
@@ -2988,6 +2988,8 @@ Map<String, Object> sanitizeHash(Map<String, Object> h,
 			Map<String, Object> hn = super.sanitizeHash(h, p, pc);
 
 			// We drop the whole <input> if type isn't allowed (case-insensitive)
```
Bombe, line 2990: Now this comment sits here very awkwardly, two lines too high… 🙂

torusrxxx: This is correct, because the value of nothing (case-insensitive) isn't allowed here.

Bombe: That’s not what the next line does, though.

src/freenet/client/filter/HTMLFilter.java
```diff
@@ -2987,6 +2987,9 @@ Map<String, Object> sanitizeHash(Map<String, Object> h,
 			HTMLParseContext pc) throws DataFilterException {
 			Map<String, Object> hn = super.sanitizeHash(h, p, pc);
 
+			// We drop the whole <input> if type is null
```
bertm: The `type` attribute on `input` is optional, defaulting to `text` as per [the spec](https://html.spec.whatwg.org/multipage/input.html#attr-input-type):

> The attribute's [missing value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default) and [invalid value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default) are both the [Text](https://html.spec.whatwg.org/multipage/input.html#text-(type=text)-state-and-search-state-(type=search)) state.

Apart from fixing an NPE, why would we drop these?

torusrxxx: I also wonder why we are filtering out some types given that any value is valid and safe. Anyway, nice to see you agree to not kill an input tag without type.

bertm: Mostly to guard against future input types that we may consider unsafe for any particular reason.
Also note that type=file is explicity not in the allowed list.

bertm (approved):

ArneBab:
Is there an example Freesite that’s broken by this? ⇒ do we need a hotfix release?

ArneBab:
Merged -- thank you!

PR #1084, opened 2025-06-30
Fix UserAlertManager Test

Bombe:
The test was using the wrong translation file, i.e. the production translation. By forcing it to use the translation file used in tests of the translation system, any key that does not exist in that file (which is pretty much every key outside of tests) will return the key as translation, making it trivial to test that the correct key is requested.


ArneBab (approved):

ArneBab:
Merged -- thank you!

PR #1085, opened 2025-07-02
Add Task to Build Source JAR

Bombe:
I have a local branch that adds a task to add the source JAR file that I use for uploading to my Maven repository, and I wondered, “why do I keep this from you?” So here you have it. 😁

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,6 +126,15 @@ task buildJar(type: Jar) {
 jar.enabled = false
 jar.dependsOn buildJar
 
+task sourcesJar(type: Jar) {
+    dependsOn generateVersionSource
+    archiveClassifier = 'sources'
+    from(sourceSets.main.allSource) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -126,6 +126,15 @@ task buildJar(type: Jar) {
 jar.enabled = false
 jar.dependsOn buildJar
 
+task sourcesJar(type: Jar) {
+    dependsOn generateVersionSource
+    archiveClassifier = 'sources'
```
bertm, line 131: Can we add this to ensure `freenet-sources.jar` is reproducible?

```suggestion
    archiveClassifier = 'sources'
    preserveFileTimestamps = false
    reproducibleFileOrder = true
```

ArneBab:
@Bombe did you catch the comment by @bertm here?

bertm (approved):

PR #1088, opened 2025-07-04
BlockTransmitter: delay BlockSenderJob asynchronously on the Ticker (improved)

bertm:
The current delay logic blocks the thread for the block transmitter, of which there can be many (roughly 1 per 32 KiB sent), leading to an increased thread count depending on the level of traffic.

Replace wait() with a Ticker-based rescheduling approach to delay the transmission in a non-blocking way.

This is a second iteration of #1087 which was incorrect.
See also #1086 for a simpler but slightly less effective approach.
Please review and test carefully.

### DIFF
--- a/src/freenet/io/xfer/BlockTransmitter.java
+++ b/src/freenet/io/xfer/BlockTransmitter.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import freenet.io.comm.AsyncMessageCallback;
 import freenet.io.comm.AsyncMessageFilterCallback;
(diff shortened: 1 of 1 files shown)


src/freenet/io/xfer/BlockTransmitter.java
```diff
@@ -156,24 +160,21 @@ public void run() {
 						// the variable count is used to count which message is being processed now
 						// the HTL is taken into consideration when adding delay too
 						count++;
-						if (isHighHtl() && count >= (Node.PACKETS_IN_BLOCK - 1)) {
-							try {
-								wait((int) (Math.random() * MAX_ARTIFICIAL_FINAL_PACKETS_DELAY));
-							} catch (InterruptedException e) {
-								// intentionally left blank: can continue
-							}
+						if (isHighHtl() && count >= (Node.PACKETS_IN_BLOCK - 2)) {
```
bertm, line 163: Note that `Node.PACKETS_IN_BLOCK - 1` has changed to `Node.PACKETS_IN_BLOCK - 2` as we now wait after sending rather than before.

ArneBab:
This looks really nice, but too complex for a hotfix release (needs a week with a test-release, I think), so I’d delay this until 1504 and use #1086 for 1503. Thank you!

ArneBab:
Merged - thank you!

PR #1090, opened 2025-07-07
Add support for new HDR chunks in PNG

torusrxxx:
New HDR chunks just became W3C Recommendation. From the W3C documentation there isn't any URLs in them, only numbers.

### DIFF
--- a/src/freenet/client/filter/PNGFilter.java
+++ b/src/freenet/client/filter/PNGFilter.java
@@ -266,12 +266,14 @@ public class PNGFilter implements ContentDataFilter {
 				}
 
 				if (!skip && "cICP".equals(chunkTypeString)) {
-					if (length != 4)
+					if (length != 4) {
 						throwError("cICP chunks invalid!",
 								"cICP chunks must be 4 bytes long!");
-					if (chunkData[2] != 0)
+					}
+					if (chunkData[2] != 0) {
 						throwError("cICP chunks invalid!",
 								"Unsupported color model other than RGB is specified in PNG!");
+					}
 					validChunkType = !(hasSeenPLTE || hasSeenIDAT);
 					hasSeenCICP = true;
 				}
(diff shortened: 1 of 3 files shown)


Bombe:
> Hey, I'm not even changing any code, just added some strings.

If the added strings do not change any functionality, why are they in there?

Ah, so they _do_ change functionality. And that change can be tested. Before this change, a `gAMA` block would have been removed, and now it will not, and that’s exactly what the test would verify.

Bombe:
Hey @torusrxxx, I created a [pull request](https://github.com/torusrxxx/fred/pull/1) for your pull request, so we can merge while we merge – and so we can test that your changes to the PNG filter actually do what they are supposed to do. 🙂

src/freenet/client/filter/PNGFilter.java
```diff
@@ -261,6 +261,12 @@ public void readFilter(InputStream input, OutputStream output, String charset, M
 				}
 
 				if (!skip && "cICP".equals(chunkTypeString)) {
+					if (length != 4)
```
Bombe: All code blocks need to be in braces.

test/freenet/client/filter/PNGFilterTest.java
```diff
@@ -137,6 +137,16 @@ public void testSuiteTest() throws IOException {
 		}
 	}
 
+	@Test
+	public void BrokencICPChunkIsFiltered() throws IOException {
```
Bombe: Method names should start with lowercase letters.

src/freenet/client/filter/PNGFilter.java
```diff
@@ -237,20 +247,56 @@ public void readFilter(InputStream input, OutputStream output, String charset, M
...
+				}
+
+				if (!validChunkType) {
+					for (int i = 0; i < HARMLESS_CHUNK_TYPES_BEFORE_PLTE.length; i++) {
+						if (HARMLESS_CHUNK_TYPES_BEFORE_PLTE[i].equals(chunkTypeString)) {
+							if (!hasSeenPLTE && !hasSeenIDAT) {
+								validChunkType = true;
+								break;
+							} else {
+								throwError("The chunk appeared in an unexpected order!",
```
Bombe, line 293: One of those messages should mention _which_ chunk is out of order, so the user can potentially do something about it.

Bombe (changes requested):

Bombe (approved):

ArneBab:

PR #1092, opened 2025-07-16
MersenneTwister (un)synchronization

bertm:
Dropping synchronization improves the random number throughput by over 250% (measured with OpenJDK 24 on a reasonably modern machine). This is especially noticeable when filling (padded) buffers with random bytes.

In some cases the random source is shared, keep those instances synchronized. Make synchronized/unsynchronized behaviour explicit by means of factory methods.

Additionally each PeerNode already owns its own Random source, use that one throughout *PeerNode instead of the global one on Node to further reduce potential contention.


src/freenet/support/io/FileUtil.java
```diff
@@ -722,11 +722,9 @@ public static boolean copyFile(File copyFrom, File copyTo) {
...
 	 */
 	public static void fill(OutputStream os, long length) throws IOException {
-		long seed;
-		synchronized (SEED_GENERATOR) {
-			seed = SEED_GENERATOR.nextLong();
-		}
-		writeRandomBytes(os, new MersenneTwister(seed), length);
+		byte[] seed = new byte[16];
+		SEED_GENERATOR.nextBytes(seed);
+		writeRandomBytes(os, MersenneTwister.createUnsynchronized(seed), length);
```
Bombe, line 727: This uses a longer seed to generate the new twister, but I’m guessing that’s both irrelevant and an improvement (if you assume longer seeds are “better”), right? 😁

src/freenet/support/math/MersenneTwister.java
```diff
@@ -78,4 +85,71 @@ public synchronized void setSeed(byte[] seed) {
...
+			unsynchronizedSetSeed(seed);
+		}
+
+		@Override
+		public void setSeed(byte[] seed) {
+			unsynchronizedSetSeed(seed);
+		}
+	}
+
+	final int unsynchronizedNext(int bits) {
```
Bombe, line 136: I don’t like that we’re adding non-`private` methods here, but oh well, I can’t think of anything else that’s not a mess, either…

Bombe (approved):

src/freenet/support/io/FileUtil.java
bertm: Correct. While chasing all the uses of MersenneTwister I encountered `Node.createRandom()` using a 128 bit seed - I figured there would be no good reason to use a smaller seed here.

src/freenet/support/math/MersenneTwister.java
bertm: Yep, this was the best thing I could think of without breaking the class hierarchy or changing the default behaviour. At least this does not pollute or break the *public* API.

ArneBab:

PR #1093, opened 2025-07-30
radiobuttons: special case WoT so the buttons show up

ArneBab:


src/freenet/clients/http/staticfiles/themes/sky-dark-static/theme.css
```diff
@@ -229,6 +229,11 @@ input[type=radio] {
   display:none;
```
bertm: While I agree that WoT should be improved, this stylesheet should be fixed for sure at that is the culprit - we should never be hiding the radio button itself (`display: none`).

Instead of hiding the `input[type=radio]` and styling the `input[type=radio] + label:before` to show a button, we can style the input instead for (almost) the same visual effect by asking the browser not to apply its default styles (`appearance: none`).

```css
input[type="radio"] {
    appearance: none;
    display: inline-block;
    width: 16px;
    height: 16px;
    border-radius: 50%;
    box-sizing: content-box;
    margin: 0 1px -2px 0;
    padding: 0;
    background: #6c6c6c;
    color: #535353;
    border: 2px solid #313131;
}

input[type="radio"]:checked {
    color: white;
}

input[type="radio"]::before {
    content: "";
    display: block;
    position: relative;
    width: 8px;
    height: 8px;
    left: 4px;
    top: 4px;
    border-radius: 50%;
    background-color: currentColor;
}
```

Visual result on WoT page:
image


As a bonus, this does not actively break screen readers and other accessibility tools like the current stylesheet does.

bertm (changes requested):

src/freenet/clients/http/staticfiles/themes/sky-dark-static/theme.css
ArneBab: That looks good -- does the new commit match your suggestion?

bertm: Yep, thanks!

PR #1094, opened 2025-07-30
Use subshell in command of readme to avoid changing the dir

ArneBab:


Bombe (approved):

ArneBab:

PR #1098, opened 2025-08-17
Fix code style

torusrxxx:
Adopted some quick fixed suggested by the Java editor that improves code readability or removes usage of deprecated functions.


src/freenet/clients/http/ConnectionsToadlet.java
```diff
@@ -774,7 +769,7 @@ public void handleMethodPOST(URI uri, final HTTPRequest request, ToadletContext
 	 * @param nodeReference - The reference to the new node
 	 * @param privateComment - The private comment when adding a Darknet node
 	 * @param trust 
-	 * @param request To pull any extra fields from
+	 * @param visibility
```
ArneBab: As you’re in the code right now, could you add a description of the use of the parameter?

(that should have been done for the `trust` parameter much earlier, too, but seems to have  been overlooked)

Also for the other added param descriptions.

PR #1099, opened 2025-08-17
Add copy-to-clipboard element for upload/download lists.

ArneBab:
The title text currently cannot be internationalized, because I wanted to keep the initialization out of the Toadlets.

### DIFF
--- a/src/freenet/clients/http/QueueToadlet.java
+++ b/src/freenet/clients/http/QueueToadlet.java
@@ -1979,7 +1979,11 @@ public class QueueToadlet extends Toadlet implements RequestCompletionCallback, 
 	private HTMLNode createKeyCell(FreenetURI uri, boolean addSlash) {
 		HTMLNode keyCell = new HTMLNode("td", "class", "request-key");
 		if (uri != null) {
-			keyCell.addChild("span", "class", "key_is").addChild("a", "href", '/' + uri.toString() + (addSlash ? "/" : ""), uri.toShortString() + (addSlash ? "/" : ""));
+			keyCell.addChild("span", "class", "key_is")
+					.addChild("a",
+							new String[]{"class", "data-key", "href"},
+							new String[]{"finished_key_link", uri.toString(), '/' + uri.toString() + (addSlash ? "/" : "")}, 
+							uri.toShortString() + (addSlash ? "/" : ""));
 		} else {
 			keyCell.addChild("span", "class", "key_unknown", l10n("unknown"));
 		}
@@ -2115,6 +2119,9 @@ public class QueueToadlet extends Toadlet implements RequestCompletionCallback, 
 		
 		HTMLNode formDiv = new HTMLNode("div", "class", "request-table-form");
 		HTMLNode form = ctx.addFormChild(formDiv, path(), "request-table-form-"+id+(advancedModeEnabled?"-advanced":"-simple"));
+        if (core.getNode().isFProxyJavascriptEnabled()) {
+            form.addChild("script", new String[] {"type", "src"}, new String[] {"text/javascript", "/static/js/clipboard.js"});
+        }
 		
 		createRequestTableButtons(form, pageMaker, ctx, mimeType, hasFriends, advancedModeEnabled, priorityClasses, true, queueType);
 
--- a/src/freenet/clients/http/QueueToadlet.java
+++ b/src/freenet/clients/http/QueueToadlet.java
@@ -1981,8 +1981,8 @@ public class QueueToadlet extends Toadlet implements RequestCompletionCallback, 
 		if (uri != null) {
 			keyCell.addChild("span", "class", "key_is")
 					.addChild("a",
-							new String[]{"finished_key_link", uri.toString(), l10n("copyToClipboardTitle"), '/' + uri.toString() + (addSlash ? "/" : "")}, 
 							new String[]{"class", "data-copytext", "data-copytitle", "href"},
+							new String[]{"copy-to-clipboard", uri.toString(), l10n("copyToClipboardTitle"), '/' + uri.toString() + (addSlash ? "/" : "")}, 
 							uri.toShortString() + (addSlash ? "/" : ""));
 		} else {
 			keyCell.addChild("span", "class", "key_unknown", l10n("unknown"));
--- a/src/freenet/clients/http/QueueToadlet.java
+++ b/src/freenet/clients/http/QueueToadlet.java
@@ -1981,8 +1981,8 @@ public class QueueToadlet extends Toadlet implements RequestCompletionCallback, 
 		if (uri != null) {
 			keyCell.addChild("span", "class", "key_is")
 					.addChild("a",
-							new String[]{"class", "data-key", "data-title", "href"},
 							new String[]{"finished_key_link", uri.toString(), l10n("copyToClipboardTitle"), '/' + uri.toString() + (addSlash ? "/" : "")}, 
+							new String[]{"class", "data-copytext", "data-copytitle", "href"},
 							uri.toShortString() + (addSlash ? "/" : ""));
 		} else {
 			keyCell.addChild("span", "class", "key_unknown", l10n("unknown"));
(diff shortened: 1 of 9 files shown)


src/freenet/clients/http/QueueToadlet.java
```diff
@@ -2115,6 +2119,9 @@ private HTMLNode createRequestTable(PageMaker pageMaker, ToadletContext ctx, Lis
 		
 		HTMLNode formDiv = new HTMLNode("div", "class", "request-table-form");
 		HTMLNode form = ctx.addFormChild(formDiv, path(), "request-table-form-"+id+(advancedModeEnabled?"-advanced":"-simple"));
+        if (core.getNode().isFProxyJavascriptEnabled()) {
+            form.addChild("script", new String[] {"type", "src"}, new String[] {"text/javascript", "/static/js/clipboard.js"});
+        }
```
bertm, line 2124: Loading it here causes the script to be included more than once when there are multiple tables on the page (like when there are both completed and running downloads), rendering duplicate buttons for each key.

(if we cannot prevent the script from loading twice for some reason, the script should guard for this condition)

src/freenet/clients/http/staticfiles/js/clipboard.js
```diff
@@ -0,0 +1,19 @@
+// @license magnet:?xt=urn:btih:1f739d935676111cfff4b4693e3816e664797050&dn=gpl-3.0.txt GPL-v3-or-Later
+function addCopyKeyFieldToFinishedTransfers() {
+  if (navigator && navigator.clipboard) {
+    var matching = document.getElementsByClassName("finished_key_link");
```
bertm: Can we make this copy-to-clipboard functionality agnostic of the context in which it is used?

One could search for `class="copy-to-clipboard"` and use `data-copy-text="{link or whatever text to copy}"` and `data-copy-title="{tooltip from i18n}"`.

src/freenet/clients/http/staticfiles/js/clipboard.js
```diff
@@ -0,0 +1,19 @@
...
+    for (var matchingElement of matching) {
+      var key = matchingElement.dataset.key;
+      if (key) {
+        var toClipboard = document.createElement("span");
+        toClipboard.textContent = " ⎘";
+        toClipboard.classList.add("copy-to-clipboard-element");
+        toClipboard.setAttribute("title", "Copy to Clipboard");
+        toClipboard.onclick=() => navigator.clipboard.writeText(key);
+        matchingElement.parentElement.appendChild(toClipboard);
```
bertm: This could directly use [`after()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/after) on `matchingElement` instead of traversing to the parent element.

src/freenet/clients/http/staticfiles/js/clipboard.js
```diff
@@ -0,0 +1,19 @@
+// @license magnet:?xt=urn:btih:1f739d935676111cfff4b4693e3816e664797050&dn=gpl-3.0.txt GPL-v3-or-Later
+function addCopyKeyFieldToFinishedTransfers() {
+  if (navigator && navigator.clipboard) {
```
bertm: The [Window `navigator`](https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator) property has been supported by browsers for over 10 years (including IE6 and the initial releases of Chrome, Firefox, and Edge).

There really is no need to check for that property to exist :smiley:

src/freenet/clients/http/staticfiles/js/clipboard.js
```diff
@@ -0,0 +1,19 @@
...
+    var matching = document.getElementsByClassName("finished_key_link");
+    for (var matchingElement of matching) {
+      var key = matchingElement.dataset.key;
+      if (key) {
+        var toClipboard = document.createElement("span");
+        toClipboard.textContent = " ⎘";
```
bertm: Best to not use whitespace characters for positioning. A better approach would be to set a `margin-inline-start` in CSS. While styling, also consider `user-select: none` to prevent the text from being selected and `cursor: pointer` to indicate that this is an action.

bertm: As for the character (⎘), this is not something that I would immediately recognize.

The emoji glyph for clipboard (📋) might be more visually recognizable (depending on OS support). What's your take?

src/freenet/clients/http/staticfiles/themes/boxed/color.css
bertm, line 1: Is this duplication necessary? If all is well .. all themes include `src/freenet/clients/http/staticfiles/color.css` anyways.

bertm (changes requested):
Feature-wise this is nice, the implementation could use some additional care.

Bombe:
> The title text currently cannot be internationalized, because I wanted to keep the initialization out of the Toadlets.

That can totally be worked around by putting the translated texts into an invisbile `<div>` somewhere on the page, and letting the javascript read the `<div>`s. 🙂

src/freenet/clients/http/staticfiles/themes/boxed/color.css
ArneBab: Some themes don’t include color.css, because that actually causes latency of displaying toadlets to increase (despite being a local service, our static file serving is pretty slow).

src/freenet/clients/http/staticfiles/js/clipboard.js
ArneBab: The character ⎘ is the only one that’s universally supported. I don’t actually see 📋 on my system (copied it as empty block).

bertm: Mmh. In that case we might be better off rendering an SVG, as the character may be too font-dependent.

That could be something like this (quick sketch based on distant memories of similar functionality):
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" style="height: 1em; width: 1em; margin-inline-start: .5ex; vertical-align: middle;">
  <path stroke="currentColor" stroke-width="2" d="M 2 16 v -12 a 2 2 0 0 1 2 -2 h 12 M 9 7 h 10 a 2 2 0 0 1 2 2 v 10 a 2 2 0 0 1 -2 2 h -10 a 2 2 0 0 1 -2 -2 v -10 a 2 2 0 0 1 2 -2" fill="none"></path>
</svg>
```

afbeelding

ArneBab: @bertm Took me a while to accept that, but I now think that you’re right.

ArneBab:
@bertm @Bombe Thank you for your review (despite the not so great shape of the code)!

I addressed all comments now:

- The script only runs once, regardless of how often it is included
- The icon is now the inline SVG by bertm
- the title is from the toadlet and localizable (via dataset.copytitle).
- it can’t be selected and the cursor is set as pointer

I left out the spacing adjustments because they were not needed with the SVG.

The SVG is added via innerHTML, but as static data, so it does not pose a risk (no escaping necessary).

Can you give this another look?

src/freenet/clients/http/staticfiles/js/clipboard.js
ArneBab: using it now. Thank you!

ArneBab: added :+1:

src/freenet/clients/http/staticfiles/js/clipboard.js
ArneBab: removed that now ☺

src/freenet/clients/http/staticfiles/js/clipboard.js
ArneBab: looks much nicer -- thanks!

src/freenet/clients/http/staticfiles/js/clipboard.js
ArneBab: it’s data-copytext and data-copytitle now. Avoided the second hyphen, because that causes dromedaryCase to appear in JS without being greppable in the source.

src/freenet/clients/http/QueueToadlet.java
ArneBab: guarding against being executed twice seemed more robust to me. Otherwise including twice would only be a single "let’s use it here, too" away.

Bombe: “Guarding against being executed twice” always sounds like “I have no control over when this is included,” which would really annoy me. 😄 Why not include the script once, in the header (or at the end of the body, if needs be), and be done with it?

PR #1101, opened 2025-08-18
Preserve the order of peers when updating handshake IPs

ArneBab:


src/freenet/node/PeerNode.java
```diff
@@ -849,8 +850,8 @@ private Peer[] updateHandshakeIPs(Peer[] localHandshakeIPs, boolean ignoreHostna
 				localHandshakeIP.getHandshakeAddress();
 			}
 		}
-		// De-dupe
-		HashSet<Peer> ret = new HashSet<Peer>();
+		// De-dupe with LinkedHashSet to preserve order
+		Set<Peer> ret = new LinkedHashSet<Peer>();
 		Collections.addAll(ret, localHandshakeIPs);
 		return ret.toArray(new Peer[0]);
```
bertm: Nitpick, feel free to ignore: wouldn't this do the exact same thing, but simpler?
```suggestion
		return Arrays.stream(localHandshakeIPs).distinct().toArray(Peer[]::new);
```

bertm (approved):

PR #1102, opened 2025-08-27
refactor(node): Overhaul DNSRequester with stateful logic and exponential backoff

deadulus:
## DNSRequester Overhaul: A Comprehensive Refactor

This pull request introduces a architectural redesign of `DNSRequester.java`.  The original implementation was a simple, stateless mechanism that suffered from significant inefficiencies. This new version refactors it into a sophisticated, stateful connection manager that is more robust, efficient, and intelligent in its handling of unconnected peers.

---

###  Original Code Problems

The original `DNSRequester` had three fundamental issues that impacted long-term node stability and performance:

1. **Peer Agnostic:** It never distinguished between IP-only and hostname peers. This meant that IP-only peers, which could be connected to immediately, were stuck in the same slow cycle of being randomly selected every 1-61 seconds, dramatically delaying initial connections.
2. **Potential Memory Leak:** It failed to clean up state for peers that were permanently removed from the node's peer list. This could lead to slow, unbounded memory growth over days or weeks of continuous operation.
3. **Stateless and Inefficient (Thrashing):** Its stateless design meant it had no memory of past connection failures. This led to endlessly retrying the same unresponsive peers, generating useless DNS queries and wasting network resources on peers that were likely offline.

---

### Key Improvements in the New Version 🚀

This refactor introduces several major architectural improvements to solve the original problems and make the process more intelligent.

#### 1. Stateful Peer Management
The most significant change is the introduction of a **state machine** for each unconnected peer (`PeerConnectionState`). The requester now tracks connection attempt windows, failure counts, and cooldown periods for every peer, allowing it to "learn" from connection failures instead of blindly retrying.

#### 2. Exponential Backoff
A direct result of stateful management, the new code implements an **exponential backoff** algorithm.
-   Peers that fail to connect within a defined window (`ATTEMPT_WINDOW_MS`) are placed into a cooldown period.
-   Each subsequent failure for the same peer **exponentially increases** the cooldown time, up to a defined maximum.
-   This intelligently de-prioritizes unresponsive peers, saving network resources and focusing efforts on viable candidates.

#### 3. Peer Categorization & Bimodal Sleep Strategy
Peers are now partitioned into **IP-only** and **Hostname** groups and treated differently:
-   **IP-only peers** are processed with a fast, fixed wait time, allowing the node to establish initial connections much more quickly.
-   **Hostname peers**, which require anonymity-sensitive DNS lookups, are processed using the original slow, randomized wait time to maintain a low-profile traffic pattern.

#### 4. Dedicated Startup Phase
The node now operates in an aggressive **startup phase** for the first few minutes of its runtime. During this window, it attempts to connect to IP-only peers without applying backoff logic to establish connections as quickly as possible.

#### 5. Proactive Memory Leak Prevention
A new periodic `cleanupStalePeers` method runs to actively remove state for peers that are no longer known to the node, **completely fixing the memory leak** present in the original code.

---

### Performance & Efficiency Analysis 📊

* **Connection Speed:** Dramatically faster at establishing initial connections by rapidly cycling through IP-only peers.
* **Resource Efficiency:** Massively reduces wasted CPU and network resources by ceasing attempts to contact known offline peers via the backoff mechanism.
* **Memory:** While using slightly more memory per-peer to store state, it prevents the long-term memory leak, making its memory profile far more stable and predictable for long-running nodes.

---

### Architectural Notes 🧠

* **Intelligent Agent:** The philosophy has shifted from simple **random selection** to an **intelligent agent** that observes outcomes, builds a model of peer availability, and adapts its strategy accordingly.
* **Immutable State:** The `PeerConnectionState` class is **immutable**. This is a robust concurrency pattern that prevents race conditions by creating a new state object for each transition instead of modifying an existing one in place. This greatly enhances the stability of the code in a multi-threaded environment.

> In summary, this refactor fixes fundamental flaws in the original design and replaces it with a modern, efficient, and stable system for managing peer connections.

### DIFF
--- a/src/freenet/node/DNSRequester.java
+++ b/src/freenet/node/DNSRequester.java
@@ -178,10 +178,10 @@ public class DNSRequester implements Runnable {
                 peerConnectionStates.remove(peer);
             } else {
                 // Unconnected: categorize for processing.
-                if (peer.getPeer().getFreenetAddress().hasHostname()) {
-                    hostnamePeers.add(peer);
+                if (pn.hasAnyHostnameAddress()) {
+                    hostnamePeers.add(pn);
                 } else {
-                    ipPeers.add(peer);
+                    ipPeers.add(pn);
                 }
             }
         }


ArneBab:
> Intelligent Agent: The philosophy has shifted from simple random selection to an intelligent agent that observes outcomes, builds a model of peer availability, and adapts its strategy accordingly.

I’ve seen the code on FMS, but I do not know where this description comes from. What does this mean apart from "there’s a cooldown" and "connection depends on whether there’s a hostname"?

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +90,316 @@ public void run() {
...
+
+    /**
+     * The state-tracking maps are cleaned up when a peer connects. However, if a peer is permanently
+     * removed from the network (i.e., it no longer appears in node.getPeers().myPeers()), its entry
+     * might persist in these maps indefinitely, leading to a slow memory leak. This helps keep maps clean.
+     * @param now The current time in milliseconds.
+     * @param partitioned The list of all unconnected peers (both IP and hostname).
+     */
+    private void cleanupStalePeers(long now, PartitionedPeers partitioned) {
+        if (now - lastCleanupTime > CLEANUP_INTERVAL_MS) {
```
ArneBab, line 142: Is this limiting necessary? And could this get the PeerNodes directly from `node.getPeers().myPeers()` so it would not have to collect them into one list again?

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +90,316 @@ public void run() {
...
+        if (hostnameCandidates.length > 0) {
             // check a randomly chosen node that has not been checked
             // recently to avoid sending bursts of DNS requests
-            PeerNode pn = nodesToCheck[node.getFastWeakRandom().nextInt(unconnectedNodesLength)];
-            if (unconnectedNodesLength < 5) {
+            PeerNode pn = hostnameCandidates[node.getFastWeakRandom().nextInt(hostnameCandidates.length)];
+
+            // 1. Add the selected peer to the recency set immediately.
+            // This ensures we don't pick it again right away, even if it's in cooldown.
+            if (hostnameCandidates.length < HOSTNAME_CANDIDATE_OPTIMIZATION_THRESHOLD) {
```
ArneBab: this branch just adds another codepath ⇒ could be removed (along with the constant).

And yes, the if branch is code by me that’s pretty much superfluous. Just kill it with fire ☺

src/freenet/node/DNSRequester.java
deadulus: Unless i am misunderstanding the loop incorrectly, i see  `realRun()` runs indefinitely, as started via `run()`, which is in a `while(true)` loop and so I felt using this "if" condition which by default is set to run once every 6 hours kept the code logic from constantly trying to search through and free memory at all times when it probably should be spending those cpu cycles on other Freenet stuff.   And yes, i am aware that there is a sleep that runs as low as 1s (most times it will be 10s) as determined by `nextMaxWaitTime`, but my thinking is that because it can take some time to build up these orphans, i figured it was more efficient to just hold off cleanup for much later.

The build up of orphan PeerNode's in these state maps do happen when i observed, but it can be random (some times very slow buildup; some times fast)  from what i observed.  Over days/weeks it can potentially get a person to sit up and take notice. So i figured letting it run once in a while seemed like the better compromise. 

For your other question about `node.getPeers().myPeers()` regarding why i split them to use ip-only and hostnames.  Yup, I can see what you are saying and interestingly enough what you are saying about `allPeers` is what i had in revision 2.  So i will revert back to that part for the method instead.  But to give you insight why i had it this current way (btw, sorry if i always keep getting so verbose)...

My original thinking was because it was only going to amount to be slightly less efficient for the added benefit of clarity and clear intent in the code, and as such, i decided to have it in its current design.  But apparently i was overthinking it 🤪

So yup, i will modify the code back to use `allPeers` way instead of this `partitioned` way.  Thanks for noticing.

src/freenet/node/DNSRequester.java
deadulus: 🤣 Ahh okay, I'll do that.  Thanks for the heads-up!

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +88,315 @@ public void run() {
...
+    /**
+     * Manages connection attempts for IP-only peers. This method iterates through all
+     * IP peers and applies the shared backoff and attempt logic to each one.
+     * @param now The current time in milliseconds.
+     * @param ipPeers The list of unconnected IP-only peers.
+     * @param isStartupPhase True if the node is in its initial startup window.
+     */
+    private void processIpOnlyPeers(long now, List<PeerNode> ipPeers, boolean isStartupPhase) {
+        for (PeerNode pn : ipPeers) {
+            managePeerAttempt(pn, now, true, isStartupPhase);
```
bertm, line 200: `maybeUpdateHandshakeIPs(isIpOnly);` with `isIpOnly = true` is effectively a no-op:
* `maybeUpdateHandshakeIPs(true)` is already called frequently in `PeerNode.shouldSendHandshake()`.
* No DNS queries will be made.

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +88,315 @@ public void run() {
...
+                } else {
+                    ipPeers.add(peer);
+                }
+            }
+        }
+        return new PartitionedPeers(ipPeers, hostnamePeers);
+    }
+
+    /**
+     * Manages connection attempts for IP-only peers. This method iterates through all
```
bertm, line 192: The name of this class reflects the purpose of being a DNS requester, I don't see why it should take care of "IP-only peers" at all, as there would be nothing to request.

Also, it does not manage connection attempts (despite added comments and logging indicating otherwise) - it just schedules resolving of IP addresses for the purpose of handshaking (when a connection is eventually attempted).

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +88,315 @@ public void run() {
...
+        List<PeerNode> ipPeers = new ArrayList<>();
+        List<PeerNode> hostnamePeers = new ArrayList<>();
+
+        for (PeerNode peer : allPeers) {
+            if (peer.isConnected()) {
+                // This peer is now connected. Clean up any lingering state from when it was unconnected.
+                peerConnectionStates.remove(peer);
+            } else {
+                // Unconnected: categorize for processing.
+                if (peer.getPeer().getFreenetAddress().hasHostname()) {
```
bertm: This will throw a `NullPointerException` if the peer has no `physical.udp` in their noderef.

bertm: Also this does not reflect whether there are any hostnames in the PeerNode's address list; it merely checks if the currently selected address (= peer) is a hostname. But `maybeUpdateHandshakeIPs` traverses and resolves all resolvable addresses, so there is a mismatch.

src/freenet/node/DNSRequester.java
```diff
@@ -61,60 +88,315 @@ public void run() {
...
+
+            if (logMINOR) {
+                String peerType = isIpOnly ? "IP" : "Hostname";
+                long cooldownDuration = newState.cooldownUntil - now;
+                Logger.minor(this, peerType + " Peer " + pn + " failed to connect/resolve. Applying backoff cooldown for "
+                        + (cooldownDuration / 60000) + " minutes. Failure count: " + newState.failureCount);
             }
-        } catch (InterruptedException e) {
-            // Ignore, just wake up. Just sleeping to not busy wait anyway
+            return false; // An attempt was NOT made; the peer was put into backoff instead.
```
bertm, line 315: This does not affect anything related to backoff on the peer.

bertm (changes requested):
There appears to be a fundamental misunderstanding of what `maybeUpdateHandshakeIPs` does. Despite the contributed code (comments, logging) and your PR description suggesting the opposite, invoking this method does not cause any connection to be established.

These changes boil down to:
* Peers that do not have hostnames at all (albeit with a flawed implementation for detecting that) do not block other peers' hostnames from being resolved.
* Peers that do not connect for a long time, have their hostname(s) resolved somewhat less frequently, improving perceived privacy.

Both of the above can be accomplished with substantially less code.

bertm:
While we're at it, please provide useful (but succinct) commit messages instead of "Update <filename>". Please refer to the [contributing guidelines](https://github.com/hyphanet/fred/blob/next/CONTRIBUTING.md) for details.

> In summary, this refactor fixes fundamental flaws in the original design and replaces it with a modern, efficient, and stable system for managing peer connections.

A little less [hyperbole](https://en.wikipedia.org/wiki/Hyperbole) would be appreciated :smiley:

src/freenet/node/DNSRequester.java
deadulus: Yes, I know.  It was part of the original v1503 (including the comment) by Arne, and my goal was to keep as much intact as possible so he and others could suggest modifications or removals.  So, I wasn't indiscriminately removing things unnecessarily.  Better to err on the side of caution in case I was misunderstanding something.  Your pointing it out now confirms it can be removed. 😊

src/freenet/node/DNSRequester.java
ArneBab: the addresses should already be sorted to have the hostname first, so it should be equivalent.

src/freenet/node/DNSRequester.java
ArneBab: A reason why people report faster connections could be that after a day most IP addresses no longer work, so the nodes with hostname are the majority of the peers that still work. With this change the non-hostname peers no longer slow down the peers with hostname.

bertm: Just grepping the `physical.udp` in my openpeers-* files I find exactly zero peers with DNS names (out of ~170 total). Unless we connect in totally different parts of the network, DNS names seem to be uncommon (apart from seed nodes).

I'm not currently convinced DNS resolution has much impact on reconnection speed at all. If this code does improve connectivity, it likely indicates that something somewhere else is seriously broken (given the peer resolution / handshake spaghetti, I would not be surprised). Maybe I'll try and dig into it at some point.

ArneBab: @bertm my guess is that this might (unintentionally) slow down checking the seednodes on startup. If you watch the wrapper.log during startup, connecting to seednodes only very slowly leads to announcements.

PR #1103, opened 2025-08-29
Continue deleting other files when some fail to delete in FileUtils

torusrxxx:
* `secureDeleteAll` should delete all files in a directory using `secureDeleteAll`.
* Continue deleting other files when some of the files fail to delete in `FileUtils`.
* Use distinct temp file names in tests.
* tests should delete temp directories on exit.

### DIFF
--- a/src/freenet/support/io/FileUtil.java
+++ b/src/freenet/support/io/FileUtil.java
@@ -592,12 +592,16 @@ final public class FileUtil {
 				return false;
 			}
 		} else {
+			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!removeAll(subfile)) return false;
+				if(!secureDeleteAll(subfile)){
+					success = false;
+				}
 			}
 			if(!wd.delete()) {
 				Logger.error(FileUtil.class, "Could not delete directory: "+wd);
 			}
+			return success;
 		}
 		return true;
 	}
@@ -613,12 +617,16 @@ final public class FileUtil {
 				return false;
 			}
 		} else {
+			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!removeAll(subfile)) return false;
+				if(!removeAll(subfile)){
+					success = false;
+				}
 			}
 			if(!wd.delete()) {
 				Logger.error(FileUtil.class, "Could not delete directory: "+wd);
 			}
+			return success;
 		}
 		return true;
 	}
--- a/src/freenet/support/io/FileUtil.java
+++ b/src/freenet/support/io/FileUtil.java
@@ -594,9 +594,7 @@ final public class FileUtil {
 		} else {
 			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!secureDeleteAll(subfile)){
-					success = false;
-				}
+				success &= secureDeleteAll(subfile);
 			}
 			if(!wd.delete()) {
 				Logger.error(FileUtil.class, "Could not delete directory: "+wd);
@@ -619,9 +617,7 @@ final public class FileUtil {
 		} else {
 			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!removeAll(subfile)){
-					success = false;
-				}
+				success &= removeAll(subfile);
 			}
 			if(!wd.delete()) {
 				Logger.error(FileUtil.class, "Could not delete directory: "+wd);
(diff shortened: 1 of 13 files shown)


test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -10,10 +10,8 @@
 
 import freenet.l10n.BaseL10n;
 import freenet.l10n.BaseL10nTest;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+
+import java.io.*;
```
bertm: Please don't collapse imports into wildcard imports.

test/freenet/client/filter/ContentFilterTest.java
```diff
@@ -357,6 +355,7 @@ public void testEvilCharset() throws IOException {
         try (
             FileOutputStream fos = new FileOutputStream("output.utf16")
         ){
+            new File("output.utf16").deleteOnExit();
```
bertm: This and the below instances should really use a `finally` block instead of resorting to JVM shutdown hooks with `deleteOnExit`.

bertm (changes requested):
Nice catch!

src/freenet/support/io/FileUtil.java
```diff
@@ -592,12 +592,16 @@ public static boolean secureDeleteAll(File wd) throws IOException {
 				return false;
 			}
 		} else {
+			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!removeAll(subfile)) return false;
+				if(!secureDeleteAll(subfile)){
```
Bombe:
```suggestion
				success &= secureDeleteAll(subfile);
```

src/freenet/support/io/FileUtil.java
```diff
@@ -613,12 +617,16 @@ public static boolean removeAll(File wd) {
 				return false;
 			}
 		} else {
+			boolean success = true;
 			for(File subfile: wd.listFiles()) {
-				if(!removeAll(subfile)) return false;
+				if(!removeAll(subfile)){
```
Bombe: Same here.

test/freenet/client/async/ClientRequestSelectorTest.java
```diff
@@ -86,6 +87,11 @@ public ClientRequestSelectorTest() throws IOException {
         jobRunner = new DummyJobRunner(executor, null);
     }
 
+    @After
+    public void tearDown() {
+        FileUtil.removeAll(dir);
```
Bombe: While that is certainly a possibility, JUnit offers a `TemporaryDirectory` rule that takes care of this automatically; can that be used, maybe?

test/freenet/crypt/EncryptedRandomAccessBucketTest.java
```diff
@@ -145,10 +145,11 @@ public void testBucketToRAF() throws IOException {
         }
     }
     
-    private File base = new File("tmp.encrypted-random-access-thing-test");
+    private File base;
     
     @Before
     public void setUp() {
+        base = new File("tmp.encrypted-random-access-thing-test");
```
Bombe: If a member variable is only assigned once and doesn’t depend on anything else and its initialization does not throw an exception, please make the member variable `final` and assign it in the declaration.

Bombe (changes requested):

test/freenet/client/async/ClientRequestSelectorTest.java
torusrxxx: JUnit 5 does not support `@Rule`. I think we will need to migrate to JUnit 5 and therefore stop using `@Rule`.

Bombe: I don’t see a pressing reason to migrate to JUnit 5, JUnit 4 certainly seems to work fine, so `@Rule` it is — unless it’s more of a hassle than it’s worth (which can happen if lots of objects get constructed before the actual test method is running). 😁

torusrxxx: No, we still need to migrate to JUnit 5 simply because we will not be using JUnit 4 forever. The migration to vintage engine should start as soon as possible so that new tests don't need to migrate from JUnit 4 to JUnit 5.

Bombe: Uhm, we are still on Java _8_, I’m pretty confident that a lot of time will pass until we actually _need_ to migrate to JUnit 5. But even if we do, migrating a `TemporaryFolder` from a `@Rule` to a `@TempDir`  is almost trivial.

test/freenet/client/async/ClientRequestSelectorTest.java
```diff
@@ -68,14 +67,13 @@ public class ClientRequestSelectorTest {
     final PersistentJobRunner jobRunner;
 
     public ClientRequestSelectorTest() throws IOException {
-        dir = new File("split-file-inserter-storage-test");
-        dir.mkdir();
+        tempFolder = new TemporaryFolder();
+        tempFolder.create();
```
Bombe, line 71: I’m not sure we should use the method that is documented with “for testing purposes only. Do not use.”

This is exactly the situation I was referring to in my earlier comment about how `TemporaryFolder` can’t be used during constructor invocation.

ArneBab:
@torusrxxx since #1107 and #1112 are merged now, can this PR be closed?

torusrxxx:
Let's close it.
I'm just getting into the AV1 world. Currently I'm making contributions to upstream AV1 projects. Will come back here when I'm more familiar with AV1 technology.

ArneBab:
That sounds awesome! (re AV1)

Thank you!

ArneBab:
@torusrxxx you may have a chance to get funding for AV1 work from NLnet: https://nlnet.nl/news/2026/20260616-67-new-projects.html (projects that got funded: your work would be a good fit, maybe more impactful) 
“Privacy preserving decentralized video platform based on Freenet/Hyphanet” -- with AV1 we’d be able to support HD video on demand. And thanks to existing moderation in Sone, the recommendation and moderation would already be solved.

torusrxxx:
I am in China, so I won't be able to accept funding for Hyphanet development. In fact none of my friends know I am a Hyphanet developer. Thanks.

PR #1108, opened 2025-10-26
Cleanup build.gradle

bertm:
Our Gradle build is using various deprecated options and features (syntax, conventions and properties), and has various custom tasks for what is supposed to be standard Gradle functionality (such as dealing with sources, compilation, resources copying, and JARs). Additionally, the source version (in Version.java) was not always updated appropriately unless a `clean` build was done.

Overhaul the `build.gradle` configuration to get rid of deprecation warnings & make it more maintainable.
Additionally, improve the following:
* Always update the Version.java when the Git version changes.
* Allow overriding the build version using the version Gradle parameter (`-PbuildXXXXX`)
* Disable incremental builds when for non-dirty builds. Those could be release builds, and incremental builds are not reproducible.
* Generate a sources JAR and Javadoc JAR on `gradlew assemble`.

@ArneBab could you please check that this did not break your release manager workflow?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,7 @@ sourceSets {
 }
 
 def gitrev
-task buildInfo {
+tasks.register('buildInfo') {
     try {
         def cmd = "git describe --always --abbrev=4 --dirty"
         def proc = cmd.execute([], project.rootDir)
@@ -92,7 +92,7 @@ tasks.withType(Javadoc) {
     options.encoding = "UTF-8"
 }
 
-task generateVersionSource(type: Copy) {
+tasks.register('generateVersionSource', Copy) {
     dependsOn buildInfo
     from(sourceSets.main.java.srcDirs) {
         include version_src
@@ -103,7 +103,7 @@ task generateVersionSource(type: Copy) {
     into version_buildir
 }
 
-task compileVersion (type: JavaCompile) {
+tasks.register('compileVersion', JavaCompile) {
     dependsOn generateVersionSource, compileJava
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
@@ -113,7 +113,7 @@ task compileVersion (type: JavaCompile) {
     destinationDirectory.set(layout.buildDirectory.dir("java/version/").get())
 }
 
-task buildJar(type: Jar) {
+tasks.register('buildJar', Jar) {
     dependsOn compileVersion, processResources
     from(processResources)
     from(compileJava.destinationDirectory) {
@@ -145,7 +145,7 @@ task buildJar(type: Jar) {
 jar.enabled = false
 jar.dependsOn buildJar
 
-task sourcesJar(type: Jar) {
+tasks.register('sourcesJar', Jar) {
     dependsOn generateVersionSource
     archiveClassifier = 'sources'
     preserveFileTimestamps = false
@@ -180,7 +180,7 @@ gradle.addBuildListener(new BuildAdapter() {
     }
 })
 
-task copyResourcesToClasses2 {
+tasks.register('copyResourcesToClasses2') {
     inputs.files sourceSets.main.allSource
     outputs.dir layout.buildDirectory.dir("classes/java/main/")
     doLast {
@@ -203,7 +203,7 @@ task copyResourcesToClasses2 {
 processResources.dependsOn copyResourcesToClasses2
 compileVersion.dependsOn copyResourcesToClasses2
 
-task copyTestResourcesToClasses2 {
+tasks.register('copyTestResourcesToClasses2') {
     inputs.files sourceSets.test.allSource
     outputs.dir layout.buildDirectory.dir("classes/java/test/")
     doLast {
@@ -244,14 +244,14 @@ tasks.withType(Test) {
     enableAssertions = false
 }
 
-task copyRuntimeLibs(type: Copy) {
+tasks.register('copyRuntimeLibs', Copy) {
     into "${buildDir}/output/"
     from configurations.runtimeClasspath
     from jar
 }
 copyRuntimeLibs.dependsOn jar
 
-task tar(type: Tar) {
+tasks.register('tar', Tar) {
   description = "Build a source release, specifically excluding the build directories and gradle wrapper files"
   compression = Compression.BZIP2
 
--- a/build.gradle
+++ b/build.gradle
@@ -67,10 +67,7 @@ sourceSets {
         }
         resources {
             srcDirs 'src'
-            include 'freenet/l10n/*properties'
-            include 'freenet/l10n/iso-*.tab'
-            include 'freenet/clients/http/staticfiles/**'
-            include 'freenet/clients/http/templates/**'
+            exclude '**/*.java'
         }
     }
     test {
@@ -80,10 +77,7 @@ sourceSets {
         }
         resources {
             srcDirs 'test'
-            include 'freenet/client/filter/*/**'
-            include 'freenet/crypt/ciphers/rijndael-gladman-test-data/**'
-            include 'freenet/l10n/*properties'
-            include 'freenet/clients/http/templates/**'
+            exclude '**/*.java'
         }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -5,11 +5,6 @@ plugins {
     id 'maven-publish'
 }
 
-configurations {
-    provided
-    compile
-}
-
 repositories {
     flatDir { dirs uri("${projectDir}/lib") }
     maven {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1,302 +1,278 @@
...
     }
     test {
         java {
-            srcDir 'test/'
+            srcDirs 'test'
+            include '**/*.java'
+        }
+        resources {
+            srcDirs 'test'
+            include 'freenet/client/filter/*/**'
```
Bombe: Hmm, why not `include **/*` and `exclude *.java`? Is there anything in these directories that is neither sources nor resources? (Same for the main source set above.)

build.gradle
```diff
@@ -1,302 +1,278 @@
...
-        }
-        copy {
-            from "${projectDir}/"
-            into layout.buildDirectory.dir("classes/java/main/")
-            include 'dependencies.properties'
-        }
+compileJava {
+    options.encoding = 'UTF-8'
+    if (!version.endsWith('-dirty')) {
+        // Make sure incremental compilation is disabled on release builds due to reproducibility issues.
```
Bombe, line 119: Is this necessary? Release builds should not be made from random developer machines, should they? Don’t we have build infrastructure for that?

test/freenet/support/io/TempBucketTest.java
Bombe, line 1: Did this file get mixed up in this PR somehow?

build.gradle
bertm: This just mimicks whatever `copyResourcesToClasses2` and friends were copying previously.
That said, turns out there are a few discrepancies.

Your proposal adds to **main resources**:
* `src/freenet/support/plugins/helpers1/package.html`
* `src/freenet/node/simulator/readme.txt`
* `src/freenet/node/probe/package.html`
Which is not much of an issue to add (although they are pretty useless files in the first place).

Your proposal removes from the **test resources**:
* `test/freenet/client/filter/PNGFilterTest.java`
* `test/freenet/client/filter/ResourceFileUtil.java`
* `test/freenet/client/filter/WAVFilterTest.java`
* `test/freenet/client/filter/OggPageTest.java`
* `test/freenet/client/filter/TagVerifierTest.java`
* `test/freenet/client/filter/FilterUtilsTest.java`
* `test/freenet/client/filter/OggBitStreamFilterTest.java`
* `test/freenet/client/filter/CSSParserTest.java`
* `test/freenet/client/filter/WebPFilterTest.java`
* `test/freenet/client/filter/ContentFilterTest.java`
* `test/freenet/client/filter/GIFFilterTest.java`
* `test/freenet/client/filter/MP3FilterTest.java`
* `test/freenet/client/filter/M3UFilterTest.java`
* `test/freenet/client/filter/TheoraBitstreamFilterTest.java`
* `test/freenet/client/filter/JPEGFilterTest.java`
* `test/freenet/client/filter/OggFilterTest.java`
* `test/freenet/client/filter/BMPFilterTest.java`
Which is not an issue at all, as those files should not be in the resources in the first place.

I'll simplify this, thanks!

build.gradle
bertm: To be fair I do not have a clue about what our build infrastructure looks like. This makes it somewhat less fragile to reproduce a build, though.

Necessary? Probably not. I'll be happy to drop 440aeeeba9f2bce13e083bc10a324cd6e9e8a0d5 if you believe there's no merit to it - Arne will just have to `./gradlew clean assemble` then. Just let me know what you prefer.

test/freenet/support/io/TempBucketTest.java
bertm: Not at all - if left untouched the build fails after 691049a5bcda010742f056972c866cd7dd583610 due to the class being non-abstract. However, looking at it again, there's a better way around this (we're currently executing a few tests multiple times) - let me fix that, and revert this file change.

bertm: Fixed in 0eddfb8aae5db13328b42c2047dfb5c44e7407dc

build.gradle
bertm: Fixed in 52e475fdedf3ac3980f533643796920a674bd89e

build.gradle
Bombe: Yeah, @ArneBab needs to chip in here, I also have no idea how our releases are built, I’m just _hoping_ there’s an automated process that does not depend on a single developer’s machine. 😄

ArneBab:
> > Gradle is simply a bad choice for tools that should be available in distributions.
> 
> Ooh, ooh, I know, I know! We should switch to Maven! 😄

Maven by now can have the plugin [maven-lockfile](https://github.com/chains-project/maven-lockfile), so it would be an option today. But I don’t know whether switching is worth all the cost surrounding it.

> > And in distributions is where we have to be, so while I did not consider it as important before, I nowadays think we should treat ant build support as requirement.
> 
> That also means that in the future we need to pay attention to PRs that touch functionality that Ant needs. You probably already do that, but I sure as hell don’t… 🙂

We could run the build via ant, too, with the dependencies pointing to the ones downloaded via gradle.

build.gradle
ArneBab: If you want to see how our releases are built, check the [sign-windows-installer workflow](https://github.com/hyphanet/sign-windows-installer/blob/main/.github/workflows/ci.yml). It does the necessary preparations and then forwards to [verify-build](https://github.com/hyphanet/scripts/blob/master/verify-build) from the scripts repo.

So while *inserting* the releases is dependent on a single devs machine (though the keys are distributed via ssss, so if I join a circus, you can recreate them), building the release works in the github workflow, too.

We’d actually get an error during the windows release signing, if my setup produced a different jar (different as in the check in verify build) than the one in the workflow.

bertm: The [`fred-release` script does use `gradlew jar`](https://github.com/hyphanet/scripts/blob/master/release-fred#L110) without cleaning, so that'd be an argument to keep this code. Or to fix the release script :smile:

bertm:
Thank you both for chiming in!

> but we do use the .tar archives for the internal auto-updates

Are you sure it's the Gradle-generated one? All the tar archives I can find in the scripts are created by `git archive` (for example, in [release-fred](https://github.com/hyphanet/scripts/blob/master/release-fred#L118)) rather than `gradlew :archive`. And `git tar` does a way better job as it includes only the sources known to git, not whatever untracked file might be in your project directory.

> We could run the build via ant, too, with the dependencies pointing to the ones downloaded via gradle.

That's an interesting thought. We could run the Ant build after the Gradle one and check that they produce the exact same JAR contents (note: they currently don't - the new first-time wizard HTML page is missing in the Ant build - and the instructions in the README require adjustments to get it working at all). That'd allow us verify the Ant build script using Gradle as guardrails - and vice-versa. That's something for another PR though :smile: 

> So yes on maven upload removal

I'll file a PR for that after this one is done.

> and yes on running the tests

Thanks! That'll become another PR, too.

build.gradle
```diff
@@ -1,302 +1,256 @@
...
-    main.compileClasspath += configurations.provided
-    test.compileClasspath += configurations.provided
+dependencies {
+    implementation 'org.bouncycastle:bcprov-jdk15on:1.59'
+    implementation 'net.java.dev.jna:jna:4.5.2'
+    implementation 'net.java.dev.jna:jna-platform:4.5.2'
+    implementation 'org.freenetproject:freenet-ext:29'
+    implementation 'io.pebbletemplates:pebble:3.1.5'
+    // dependencies of pebble
+    implementation 'org.unbescape:unbescape:1.1.6.RELEASE'
```
Bombe, line 31: Hmm, why are we including this here? From what I can see, pebble already depends on both unbescape and slf4j-api, so we don’t need to do that here, do we?

build.gradle
```diff
@@ -1,302 +1,256 @@
...
+    // Add copyRuntimeLibs so `gradlew assemble` collects all build artifacts.
+    // Note: jar, sourcesJar, javadocJar are added as dependency by default.
+    dependsOn copyRuntimeLibs
 }
 
-task tar(type: Tar) {
-  description = "Build a source release, specifically excluding the build directories and gradle wrapper files"
-  compression = Compression.BZIP2
+tasks.register('archiveDigests') {
+    description = 'Display SHA-256 digests of all archives that were created.'
```
Bombe, line 198: This task also strikes me as super-useless and nothing external tools couldn’t do equally well.

build.gradle
ArneBab: Does pulling them in via pebble still make gradle dependency validation work?

Bombe: Judging from all the dependencies listed in e.g. `verification-metadata.xml`, Gradle verifies everything _and_ its dependencies, so we _should_ be good. A small experiment would probably be prudent, though. 🙂

ArneBab: A risk is that we actually have to lock down the exact versions of the dependencies we have in dependencies.properties, else the development environment and the runtime would diverge.

Bombe: Unlikely; I’m reasonably confident that Maven Central does not allow updating releases, so the dependencies should never change.

PR #1109, opened 2025-11-08
Fix: open as text link on downloads now uses utf-8 encoding

ArneBab:
Tiny change, replaces most of the use-cases of the much more complex and incorrect https://github.com/hyphanet/fred/pull/970


src/freenet/clients/http/FProxyToadlet.java
```diff
@@ -949,7 +949,7 @@ public void run() {
 					}
 					option = optionList.addChild("li");
 					// FIXME: is this safe? See bug #131
-					MediaType textMediaType = new MediaType("text/plain");
+					MediaType textMediaType = new MediaType("text/plain;charset=utf-8");
 					textMediaType.setParameter("charset", (e.getExpectedMimeType() != null) ? MediaType.getCharsetRobust(e.getExpectedMimeType()) : null);
```
Bombe: It kinda looks like this line would immediately clobber whatever you’ve set in the constructor…

ArneBab: damn, sometimes I’m dumb and blind … thank you for catching it. This would have been so much easier if I hadn’t done that late at night and left it uncommitted for weeks …

Bombe: No, this would have been easier if you had written a test first. 😄 

(I’m currently trying to write one, but this would be easier if somebody merged https://github.com/hyphanet/fred/pull/1043 first, because that one contains a lot of shit for easier tests. 🙂)

ArneBab: Do you really mean the "quote n2ntm PR"?

(it feels like the quoting PR still needs so much work that I should only finish it after the next release)

Bombe: Actually, both #1043 and #1106, I need the added dependencies for this test I’m currently building for this… but I’d actually settle for #1106, it’s more important, we really need the updated Mockito version. 🙂 

…but I can already say: nope, this also doesn’t work. 😁 (Actually, the next line [953] doesn’t.)

Bombe (approved):

ArneBab:
merged -- @Bombe thank you for the huge improvements and the review!

PR #1111, opened 2025-11-09
DNSRequester: reduce delay without hostname 100x

ArneBab:
If there’s no hostname, we don’t need to delay, because there’s then no privacy risk.


ArneBab:
Closed, because it’s merged into #1131 (which also speeds up DNS checks while not connected).

PR #1119, opened 2026-01-18
i10n Italian Improvement

tajmone:
* Fix typos.
* Fix straight quotes and apostrophes to curly quotes.
* Fix wrong accents (é -> è; É -> È; E' -> È).
* Fix em-dashes (" - " -> " — ").
* Remove redundant spaces.
* Translate various untranslated entries.
* Complete translation of the `FirstTimeWizardToadlet` group.
----

## Contributor's notes

This is an initial polishing of the current Italian translation, since I'd like to take on its maintainance and completion from here onward (I saw that there's no listed maintainer for Italian). I don't like use Transferix, I prefer to work on the sourcefile via Git.

I hope it's OK that I change straight quotes to curly quotes in the strings — they look nicer and pose no problem for HTML rendering.

I've noticed by diffing the Italian source with the English that there many missing entries. Is there a way for you to integrate the missing entries via automation, after the merge? If not, I'll add them via a diff tool and start to translate them as I go along.

There are also some strings present in the Italian which are not present (or no longer present) in the English. Are these obsolete translation entries? should they be purged?


Bombe:
> I hope it's OK that I change straight quotes to curly quotes in the strings — they look nicer and pose no problem for HTML rendering.

That’s 100% awesome, as long as they are the type of quote characters used in Italian. Is Italian your native language?


> I've noticed by diffing the Italian source with the English that there many missing entries. Is there a way for you to integrate the missing entries via automation, after the merge? If not, I'll add them via a diff tool and start to translate them as I go along.

Not sure what you mean by that. All translation entries should be in a single file, there is no reason to integrate anything _after_ the merge; the merge _is_ the integration. 🙂


> There are also some strings present in the Italian which are not present (or no longer present) in the English. Are these obsolete translation entries? should they be purged?

If they don’t exist in the English translation, it’s highly likely that they have been removed, yes. If you want to be properly diligent about verifying that, I fear that looking through the history of the English translation (to locate where an entry has been introduced and/or removed) is the only way.

tajmone:
> That’s 100% awesome, as long as they are the type of quote characters used in Italian. Is Italian your native language?

My native languages are both English and Italian (born in the UK, raised in Italy with one parent from each nation), so I've been speaking both since childhood.

> Not sure what you mean by that. All translation entries should be in a single file, there is no reason to integrate anything after the merge; the merge is the integration. 

I diff-compared the English and Italian translation, and there are many strings present in the English source which are not carried over into the Italian sourcefile (neither in English nor translated), so I thought the Italian file might be lagging behind the English (assuming that the English is the reference i10n file).

> First of all, signing commits is 100% free (well, assuming your time is free), so I don’t see a reason to not sign your commits. 

True, it's just that I haven't setup my Git workflow for signatures since I usually don't need them, and I'll need to find a spare-time slot to read the docs and configure that. It's just that free time is a rare commodity, but I'll do set up the digital signature on Git and my GitHub profile, it might take me some time since I'll need a few hours to dedicate to it.

tajmone:
> Yes, English is indeed the reference translation, but as said, there is no separate integration process. 🙂 Sometimes, when adding new keys to the English file, we remember to add the corresponding keys to all other files, but sometimes… we don’t. So yeah, almost all other translations (except probably German) will be lagging behind the English one.

Ok, no problem: I'll integrate the missing lines via a diff tool, as I work on the file. That's fairly easily done (albeit a bit tedious).

I've managed to setup the commit signing, so I've forced-pushed the PR commit which is now signed.

I'm planning to regularly work on the Italian translation, until it integrates all the missing entries and it's fully translated.

ArneBab:
We’re usually getting the translations via localization labs (previously transifex):
https://explore.transifex.com/otf/freenet/

That’s why there are lines missing in the files here: we only import lines that actually have translations.

tajmone:
Ok, great. If it's OK with you, I'd rather proceed with the translation work via GitHub PRs, using a diff tool to integrate the missing entries (by integrating and translate them directly, a group at the time). Working directly on the source files allows me a better overview of the context, e.g. quickly finding how a given menu or term was translated, thus enabling consistence throughout the translation.

The PR commit is now signed, so if it's OK to merge it I'll then carry on the translation work approximately once a week, in chunks compatible with my free time.  

> If they don’t exist in the English translation, it’s highly likely that they have been removed, yes. If you want to be properly diligent about verifying that, I fear that looking through the history of the English translation (to locate where an entry has been introduced and/or removed) is the only way.

That's OK, the entries of these type which I've spotted seem to belong to a same group no longer in the English source, using Git blame I should easily manage to track their deletion and find out the reason via the commit messages. If they are obsolete, I'll propose their deletion. In any case, their presence should be harmless.

Bombe:
> I wasn't sure if I should add more commits to the PR or squash them into a single commit, since I've come to understand that integration into the source code is not done via GitHub. Anyhow, if you'd rather have me squash all changes into a single (signed) commit just let me know.

Nah, we don’t insist on squashing. Multiple commits are fine, within reason.

> I'll proceed with the translation by integrating & translating all the missing entries, from start to finish, and translate any entries that were still in in English in the Italian file. It's a huge file, but hopefully it will be done soon.

Thank you!

src/freenet/l10n/freenet.l10n.it.properties
```diff
@@ -95,50 +95,82 @@ BookmarkManager.list=Segnalibri
...
+Bookmarks.Defaults.ShortDescription.sharesitedocumentation=Documentazione di Sharesite, a cura di terzi
+Bookmarks.Defaults.ShortDescription.CleanSpider=Indice automatico con filtraggio rigoroso
+Bookmarks.Defaults.ShortDescription.FFS=Primi passi su Freenet
 Bookmarks.Defaults.ShortDescription.FMS=I forum su Freenet
 Bookmarks.Defaults.ShortDescription.Fniki=Una wiki vecchia a proposito di Freenet
 Bookmarks.Defaults.ShortDescription.Freemail=Email attraverso Freenet
 Bookmarks.Defaults.ShortDescription.unofficialfmsguide=La guida non ufficiale a FMS (Forums over Freenet)
+Bookmarks.Defaults.ShortDescription.statistics=Statistiche sull’uso di Freenet
 Bookmarks.Defaults.ShortDescription.PrivacyAndSoftware=Le migliori pratiche per gli utenti Freenet
+Bookmarks.Defaults.ShortDescription.PrivacyAndSoftware=Pratiche consolidate per gli utenti di Freenet
```
bertm, line 142: The `Bookmarks.Defaults.ShortDescription.PrivacyAndSoftware` key is duplicated.

bertm (changes requested):
Several lines only contain quotation changes to original non-translated (English) text.

It'd be better to remove those entries from the translation files (making them "untranslated") than to adjust them (giving the impression that the entries are translated; even though they are still in English).

bertm:
> If they don’t exist in the English translation, it’s highly likely that they have been removed, yes.

There's a tool dedicated to removing such orphaned entries from our translation files in the repository. I've opened #1122 to clean things up a little. From a quick check it appears that all seemingly orphaned entries are indeed no longer used.

src/freenet/l10n/freenet.l10n.it.properties
tajmone: Fixed!

tajmone:
# Translation Completed!

With the last commit I finished integrating and translating all the missing entries from the English source, and translating any entries that were untranslated in the Italian source. Having had a chance to sift through the entire source, I think that in the future it could be improved by enforcing a bit more consistency in similar sentences and terminology usage and translation. When I have time I'll annotate in a Markdown document some guidelines for the Italian translation maintenance—e.g. how specific terms are rendered, and so on, so that future collaborators might benefit from it.

If you prefer, I can squash all four commits into one.

ArneBab:
Thank you for fixing the translation and adding what was missing!

And respect for getting through more than 2342 lines of translations! From cursory reading this looks good (though my Italian isn’t good enough to do any more than plausibility checking).

Four commits are good -- no need to squash.

Merged. And thank you again!

PR #1120, opened 2026-02-07
Fix Empty Responses when IOExceptions Are Thrown in Toadlets

Bombe:
(This PR is built upon [PR-1121](https://github.com/hyphanet/fred/pull/1121), because it needs all the test scaffolding from that PR. Merge that one first, this only adds the last commit.)

In `ToadletContextImpl.handle()`, in the outer-most `try`-`catch` block, `IOException`s are completely ignored and do not generate a response; it was probably assumed that `IOException`s will always mean that either the input stream or the output stream of the request itself had some kind of issues, in which case generating a response would be mostly useless; however, `Toadlet`s are also allowed to throw `IOException`s from their `handleMethod*` methods. These exceptions now generate an HTTP status 500 response and include a stack trace in the response body.


Bombe:
Indeed, we are actually closing the socket (by way of closing the socket’s `InputStream` by way of the try-with-resources) before we’re actually sending out any error replies. That’s sub-optimal!

ArneBab:
merged - thank you!

PR #1122, opened 2026-02-08
Remove orphaned translations (run freenet.tools.CleanupTranslations)

bertm:
Several locales still contain translations that are no longer used and hence have been removed from the English (baseline) translation file.

Remove those orphaned translations from their locales using freenet.tools.CleanupTranslations.


ArneBab:
Looks good to me. I’ll resolve the conflicts in the Italian translation by keeping the keys.

ArneBab:
Merged -- thank you!

PR #1115, opened 2025-11-22
Add a bunch of more new MIME types

torusrxxx:
Adds a bunch of new MIME types not previously recognized, including webm, docx and others.

PR #1123, opened 2026-02-13
Fix Hashes Generated by Insert Compressor

Bombe:
The `InputStream.skip()` method is _not_ required to skip as many bytes as specified, or skip as many as it can, or actually skip any bytes at all; `return 0;` is a 100% valid implementation. It also is not required to actually _read_ the input stream; if e.g. `seek()`-type functionality is available, it may be used to not read anything at all but simply skip (hence the name!) to the desired offset in the stream.

As the `MultiHashInputStream` changed its superclass from `InputStream` to `SkipShieldingInputStream`, its implementation of `skip()` changed, too: `InputStream`’s `skip()` actually tries to `read()` as many bytes as specified, only ending prematurely if the stream EOFs. `SkipShieldingInputStream`’s `skip()`, however, only reads up to 8k of data — which is rather questionable, but okay.

Both implementations correctly return the number of bytes they skipped, and so, while it can be argued that `MultiHashInputStream`’s `skip()` implementation has changed its observable behaviour and should be fixed, it is in fact the _usage_ of the `skip()` method in the `InsertCompressor` which is beyond any doubt incorrect. This pull request fixes the usage.

### DIFF
--- a/src/freenet/client/async/InsertCompressor.java
+++ b/src/freenet/client/async/InsertCompressor.java
@@ -67,7 +67,7 @@ public class InsertCompressor implements CompressJob {
 		this.minSize = minSize;
 		this.bucketFactory = bf;
 		this.persistent = persistent;
-		this.compressorDescriptor = inserter.ctx.compressorDescriptor;
+		this.compressorDescriptor = inserter.getInsertContext().compressorDescriptor;
 		this.generateHashes = generateHashes;
 		this.pre1254 = pre1254;
 		this.config = config;
--- a/src/freenet/client/async/InsertCompressor.java
+++ b/src/freenet/client/async/InsertCompressor.java
@@ -148,7 +148,11 @@ public class InsertCompressor implements CompressJob {
 									amountOfDataToCheckCompressionRatio, minimumCompressionPercentage);
 						} catch (CompressionOutputSizeException | CompressionRatioException e) {
 							if(hasher != null) {
-								is.skip(Long.MAX_VALUE);
+								int eof;
+								do {
+									is.skip(Long.MAX_VALUE);
+									eof = is.read();
+								} while (eof != -1);
 								hashes = hasher.getResults();
 								first = false;
 							}
--- a/src/freenet/client/async/SingleFileInserter.java
+++ b/src/freenet/client/async/SingleFileInserter.java
@@ -67,7 +67,7 @@ class SingleFileInserter implements ClientPutState, Serializable {
 	
 	final BaseClientPutter parent;
 	InsertBlock block;
-	final InsertContext ctx;
+	private final InsertContext ctx;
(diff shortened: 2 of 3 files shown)


src/freenet/client/async/InsertCompressor.java
```diff
@@ -148,7 +148,11 @@ public boolean run(ClientContext context) {
 									amountOfDataToCheckCompressionRatio, minimumCompressionPercentage);
 						} catch (CompressionOutputSizeException | CompressionRatioException e) {
 							if(hasher != null) {
-								is.skip(Long.MAX_VALUE);
+								int eof;
+								do {
+									is.skip(Long.MAX_VALUE);
```
bertm, line 153: This could benefit from `FileUtil.skipFully`.

src/freenet/client/async/SingleFileInserter.java
```diff
@@ -1008,6 +1008,10 @@ public BaseClientPutter getParent() {
 		return parent;
 	}
 
+	public InsertContext getInsertContext() {
```
bertm: Could this method be package-private?

bertm (changes requested):
Excellent catch! Here's a few nitpicks for the finishing touch.

src/freenet/client/async/InsertCompressor.java
Bombe: `skipFully()` requires that you know exactly how many bytes you want to skip, which we don’t know after the compressor has read an unknown amount of bytes from the stream.

src/freenet/client/async/SingleFileInserter.java
Bombe: It absolutely could.

src/freenet/client/async/InsertCompressor.java
bertm: Yikes, you're absolutely right!

bertm (approved):

bertm:
This look OK. As far as I can tell this fixes inserts for non-compressible files above 8MiB (the default "check if this file is compressible" limit) - without this fix all large non-compressible inserts are broken.

Note that even with this fix, the key for the example file provided by ArneBab in #1055 is different compared to the pre-1504 key. This seems to be caused by the newly introduced WebM MIME types.

Additionally reverting #1115 (on top of the fix provided here) gets us back to the pre-1504 key - assuming `video/webm` is explicitly provided as MIME type (auto-detected by `pyFreenet`). This is not an error but rather an encoding difference (the manifest now contains the registered short-id for the MIME type rather than the string value).

We should get this out soon. @Bombe anything you're still working on?

Meanwhile we should probably look into introducing some regression tests regarding CHK stability that would have caught this error when I introduced it - I'll have a go at that.

PR #1130, opened 2026-03-28
update Actions

qupo1:
- gradle/actions v5 -> v6
- actions/upload-artifact v6 -> v7

Adding these versions to repo CI settings will be required.
I think removing old versions from allowlist is also good idea.


ArneBab:
Looks good to me. Merged. Thank you!

PR #1131, opened 2026-04-06
fix DNS checks: fast during startup, throttle after first connection

ArneBab:
DNS requests are throttled to reduce visibility of requests on DNS servers.

The throttling caused seednode checks to be delayed by 30s per seed, slowing down initial connection.

The fix reduces the throttling while the node is unconnected and increases the throttling once at least one node is connected.

### DIFF
--- a/test/freenet/node/DNSRequesterTest.java
+++ b/test/freenet/node/DNSRequesterTest.java
@@ -1,56 +1,264 @@
 package freenet.node;
 
+import freenet.io.comm.Peer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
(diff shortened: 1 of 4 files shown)


test/freenet/node/DNSRequesterTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import junit.framework.TestCase;
+
+@RunWith(Parameterized.class)
+public class DNSRequesterTest extends TestCase {
```
Bombe: `TestCase` is JUnit 3. Please don’t! 😄

ArneBab: arg … killed it off. That was completely unnecessary …

Thank you!

Bombe (approved):
Guess I have no choice but to approve it now, eh? 😄

ArneBab:
merged -- thank you for the review!

PR #1132, opened 2026-05-09
Remove downloading plugins from clearnet URL: try FreenetURI instead

ArneBab:


Bombe:
I’m missing the change for the plugins page where the whole section dealing with URLs is removed.

RealJosephKnapp:
This is a good idea. Not everyone has a free internet and using centralized servers presents a single point of failure.

PR #1134, opened 2026-06-22
Remove old statistics site from bookmarks.

ArneBab:
It’s no longer updated.


src/freenet/clients/http/staticfiles/defaultbookmarks.dat
```diff
@@ -100,36 +100,31 @@ BookmarkCategory2.Content.Bookmark6.hasAnActivelink=true
 BookmarkCategory2.Content.Bookmark6.Name=The Unofficial FMS Guide
 BookmarkCategory2.Content.Bookmark6.Description=L10N:unofficialfmsguide
 BookmarkCategory2.Content.Bookmark6.ShortDescription=L10N:unofficialfmsguide
-BookmarkCategory2.Content.Bookmark7.URI=USK@WMa1Z40iYdZZ51yctQ3toFl9zuuFEnNdsm3NejJU5KE,jCBcaNBeKD5~sSQeSkyKz737Bh5ibBGqdzfD8mgfdMY,AQACAAE/statistics/802/
-BookmarkCategory2.Content.Bookmark7.hasAnActivelink=true
-BookmarkCategory2.Content.Bookmark7.Name=Freenet Statistics
-BookmarkCategory2.Content.Bookmark7.Description=L10N:statistics
```
Bombe, line 106: Does this mean we can remove stuff from the translations as well, or is that still used somewhere else?

ArneBab: Yes, we can remove the translations, too … I’ll try to get to that (didn’t think of the translations of the bookmarks, though I should have …)

ArneBab: need to walk back on this statement ☺ ⇒ must always keep bookmark translations.
But that also means that this PR should be good to go, right?

Bombe: Also, we should probably use better names than `Bookmark7`. 😄

ArneBab: "Bookmark7" is a technical identifier from the parser that reads the bookmarks.dat file. The l10n key uses `statistics`.

ArneBab:
merged -- thank you for the review!

PR #1135, opened 2026-07-05
Recover lingering default bookmarks localization

ArneBab:
These localization lines are needed for nodes that still have the old
default bookmarks. Removing them broke the descriptions.


Bombe:
Maybe we could/should add “(This bookmark has since been removed from the default bookmarks.)” after some of the translated values?

ArneBab:
That sounds like a good way to deal with this. If a site goes bad and we remove it, it then at least makes it clear to people that it’s not endorsed by the devs in any way.

ArneBab:
@Bombe added the note that the removed bookmarks are removed.

ArneBab:
Merged (with the "removed" note added) -- thank you for the review!

PR #1138, opened 2026-08-16
prepare version switch with Java 17

ArneBab:
Split shares for the UPDATE_URI have been sent to core devs as contingency.

### DIFF
--- a/src/freenet/node/updater/NodeUpdateManager.java
+++ b/src/freenet/node/updater/NodeUpdateManager.java
@@ -92,11 +92,11 @@ public class NodeUpdateManager {
 	public final static String UPDATE_URI = "USK@rW3wdla3fvFIBcMyl1xyd8umjiQQaZaQ-UqaZQCDOMQ,FfihUHiGuHNgNchv5paweQV5oHD0k4~MdlS-eNXcDF0,AQACAAE/jar/"
 			+ Version.buildNumber();
 
-	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on EoL JVMs. */
+	/** The URI for post-TRANSITION_VERSION_JAVA8 builds' freenet.jar on EoL JVMs up to Java 8. */
 	public final static String LEGACY_UPDATE_URI_JAVA8 = "SSK@vCKGjQtKuticcaZ-dwOgmkYPVLj~N1dm9mb3j3Smg4Y,-wz5IYtd7PlhI2Kx4cAwpUu13fW~XBglPyOn8wABn60,AQACAAE/jar-"
 			+ TRANSITION_VERSION_JAVA8;
 
-	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on EoL JVMs. */
+	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on EoL JVM up to Java 7. */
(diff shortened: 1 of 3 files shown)


src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -76,16 +76,26 @@
...
 	 * update to this point via old UOM.
 	 */
 	public final static int TRANSITION_VERSION = 1481;
 
 	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on modern JVMs. */
-	public final static String UPDATE_URI = "USK@vCKGjQtKuticcaZ-dwOgmkYPVLj~N1dm9mb3j3Smg4Y,-wz5IYtd7PlhI2Kx4cAwpUu13fW~XBglPyOn8wABn60,AQACAAE/jar/"
+	public final static String UPDATE_URI = "USK@rW3wdla3fvFIBcMyl1xyd8umjiQQaZaQ-UqaZQCDOMQ,FfihUHiGuHNgNchv5paweQV5oHD0k4~MdlS-eNXcDF0,AQACAAE/jar/"
 			+ Version.buildNumber();
 
+	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on EoL JVMs. */
```
Bombe: Now we have _two_ “post-TRANSITION_VERSION builds” and I think we should update both of these comments to be more explicit about what they are. 🙂

test/freenet/support/JVMVersionTest.java
```diff
@@ -13,6 +13,22 @@ public void testTooOldWarning() {
 		assertTrue(JVMVersion.isEOL("1.5"));
 		assertTrue(JVMVersion.isEOL("1.7.0_65"));
 		assertTrue(JVMVersion.isEOL("1.7"));
+		assertTrue(JVMVersion.isEOL("10"));
```
Bombe, line 18: This name of this method and the `testRecentEnoughWarning` method are terrible confusing, as e.g. Java 10 is in _both_ of them. Is it too old or recent enough?

Bombe (changes requested):

src/freenet/node/updater/NodeUpdateManager.java
```diff
@@ -76,17 +76,27 @@
...
+
 	/**
 	 * The last build on the previous key with Java 7 support. Older nodes can
 	 * update to this point via old UOM.
 	 */
 	public final static int TRANSITION_VERSION = 1481;
 
 	/** The URI for post-TRANSITION_VERSION builds' freenet.jar on modern JVMs. */
-	public final static String UPDATE_URI = "USK@vCKGjQtKuticcaZ-dwOgmkYPVLj~N1dm9mb3j3Smg4Y,-wz5IYtd7PlhI2Kx4cAwpUu13fW~XBglPyOn8wABn60,AQACAAE/jar/"
+	public final static String UPDATE_URI = "USK@rW3wdla3fvFIBcMyl1xyd8umjiQQaZaQ-UqaZQCDOMQ,FfihUHiGuHNgNchv5paweQV5oHD0k4~MdlS-eNXcDF0,AQACAAE/jar/"
```
Bombe, line 92: The comments still don’t match up 100% with the situation, because `UPDATE_URI` is _not_ for post-1481 build, but in fact for post-1507 builds.

Bombe (changes requested):

src/freenet/node/updater/NodeUpdateManager.java
ArneBab: fixed - thanks!

Bombe (approved):

PR #1141, opened 2026-08-30
ignore validation of javadoc and sources

ArneBab:
these can fail builds but do not enter the resulting build.

IntelliJ tries to get them even though they are not defined -- and consequently fails to build.


Bombe (approved):
Oh yeah, very good change. ❤️

PR #1142, opened 2026-08-30
Fix randomize probe location

ArneBab:


src/freenet/node/probe/Probe.java
```diff
@@ -649,7 +678,11 @@ private void respond(final Type type, final Listener listener) {
 			listener.onLinkLengths(linkLengths);
 			break;
 		case LOCATION:
-			listener.onLocation((float)node.getLocation());
+			/*
+			 * Noise that stays within 50% of the peers short distance connections: shift the location
+			 * by a value drawn from a gaussian distribution with mean 0 and std 0.005 (long_distance/2).
```
Bombe: Technically, we currently don’t do a Gaussian here… 😄

test/freenet/node/probe/ProbeTest.java
```diff
@@ -0,0 +1,66 @@
...
+		when(node.getRandom().nextDouble()).thenReturn(1.0);
+
+		probe.start((byte) 1, 12345L, Type.LOCATION, listener);
+		verify(listener).onLocation((float) ((0.999 + (1.0 - 0.5) * 0.01 / 2.0) - 1.0));
+	}
+
+	@Test
+	public void underflowOnNodeLocationRandomizationIsHandledCorrectly() {
+		when(node.getLocation()).thenReturn(0.001);
+    // the fake nextDouble is also used for wait, so we cannot go below Probe.getRandomizedWaitStepTime()
```
Bombe: Some spaces slipped through!

Bombe (changes requested):

src/freenet/node/probe/Probe.java
ArneBab: forgot to update the comment with the code :sweat: -- fixed.

test/freenet/node/probe/ProbeTest.java
ArneBab: fixed, thank you!

Bombe (approved):

